PR #35, opened 2020-03-18
Add new class SnoContext - resolves --repo consistently

olsen232:
Lots of sno commands will need machine readable output (ie, json)...
... including reporting errors in a consistent, machine readable way.

In order that "non-existent repo" is reported in a consistent way, I have moved the non-existent repo error handling code into the context object itself. In a follow up change, I will make sure that this error is output using json when json flag is set.

At least two (very minor) bugs were fixed as a side effect of this:
- Some subcommands catch and re-raise a GitError as a click.BadParameter, but not all do (eg sno diff) - instead they test if the the repo returned is falsey, but this is not how pygit2 works.
- In cli.py, code tests for ctx.obj["repo_dir"] - but this string is always truthy, it defaults to "." - the result of this being always-true is that the flag "-C ." is always passed to the git executable, even if this is not the user's intent. (It doesn't seem to matter much though).

### DIFF
--- a/sno/sno_context.py
+++ b/sno/sno_context.py
@@ -1,0 +1,43 @@
+import click
+import pygit2
+import os
+
+
+class SnoContext(object):
+    def __init__(self):
+        self._repo_dir = None
+
+    @property
+    def repo_dir(self):
+        """The path of the repository. Defaults to the current directory."""
+        return self._repo_dir or os.curdir
+
+    @repo_dir.setter
+    def repo_dir(self, repo_dir):
+        self._repo_dir = repo_dir
+        if hasattr(self, "_repo"):
+            del self._repo
+
+    @property
+    def repo_dir_is_set(self):
+        """True if repo_dir has been explicitly set."""
+        return self._repo_dir is not None
+
+    @property
+    def repo(self):
+        """
+        Returns the sno repository at repo_dir.
+        Raises an error if there isn't a valid repo at repo_dir.
+        """
+        if not hasattr(self, "_repo"):
+            try:
+                self._repo = pygit2.Repository(self.repo_dir)
+            except pygit2.GitError:
+                self._repo = None
+
+        if not self._repo or not self._repo.is_bare:
+            if self.repo_dir_is_set:
+                raise click.BadParameter("Not an existing repository", param_hint="--repo")
+            else:
+                raise click.UsageError("Current directory is not an existing repository")
+        return self._repo
--- a/sno/checkout.py
+++ b/sno/checkout.py
@@ -18,7 +18,7 @@ from .working_copy import WorkingCopy
 @click.argument("refish", default=None, required=False)
 def checkout(ctx, branch, fmt, force, path, datasets, refish):
     """ Switch branches or restore working tree files """
-    repo_dir = ctx.obj.repo_dir
+    repo_path = ctx.obj.repo_path
     repo = ctx.obj.repo
 
     # refish could be:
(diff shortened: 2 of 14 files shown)


rcoup:
> Some subcommands catch and re-raise a GitError as a click.BadParameter, but not all do (eg sno diff) - instead they test if the the repo returned is falsey, but this is not how pygit2 works.

Yeah, some of this is leftovers from the initial GitPython prototype. Good catch 👍

rcoup:
> the flag "-C ." is always passed to the git executable, even if this is not the user's intent. (It doesn't seem to matter much though).

I think this is ok, means it won't pick up any unexpected `GIT_DIR` environment setting.

sno/sno_context.py
```diff
@@ -0,0 +1,43 @@
...
+        if hasattr(self, "_repo"):
+            del self._repo
+
+    @property
+    def repo_dir_is_set(self):
+        """True if repo_dir has been explicitly set."""
+        return self._repo_dir is not None
+
+    @property
+    def repo(self):
```
rcoup: Can we add a `repo.delattr()` to allow us to deallocate the cached repo via `del ctx.obj.repo`? Windows is aggressive about file-locking, which potentially means we can't shell out to `git` with an open repo in all circumstances

sno/sno_context.py
```diff
@@ -0,0 +1,43 @@
...
+        self._repo_dir = None
+
+    @property
+    def repo_dir(self):
+        """The path of the repository. Defaults to the current directory."""
+        return self._repo_dir or os.curdir
+
+    @repo_dir.setter
+    def repo_dir(self, repo_dir):
+        self._repo_dir = repo_dir
```
rcoup:
```suggestion
        self._repo_dir = Path(repo_dir)
```

sno/sno_context.py
```diff
@@ -0,0 +1,43 @@
...
+        if hasattr(self, "_repo"):
+            del self._repo
+
+    @property
+    def repo_dir_is_set(self):
```
rcoup: In the style of is/has, maybe:
```suggestion
    def has_repo_dir(self):
```
?

sno/sno_context.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @property
+    def repo(self):
+        """
+        Returns the sno repository at repo_dir.
+        Raises an error if there isn't a valid repo at repo_dir.
+        """
+        if not hasattr(self, "_repo"):
+            try:
+                self._repo = pygit2.Repository(self.repo_dir)
```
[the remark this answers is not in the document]
rcoup: https://github.com/libgit2/pygit2/pull/990

sno/sno_context.py
```diff
@@ -0,0 +1,43 @@
+import click
```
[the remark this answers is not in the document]
olsen232: Done

sno/sno_context.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class SnoContext(object):
+    def __init__(self):
+        self._repo_dir = None
+
+    @property
+    def repo_dir(self):
+        """The path of the repository. Defaults to the current directory."""
+        return self._repo_dir or os.curdir
```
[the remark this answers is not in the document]
olsen232: Done

sno/sno_context.py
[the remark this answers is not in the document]
olsen232: Have passed str(self.repo_path) for now

sno/sno_context.py
olsen232: Not quite sure about this name since obj.has_repo_path is False even though obj.repo_path is never None. But, done.

sno/sno_context.py
olsen232: Done

PR #36, opened 2020-03-19
Added support for sno status --json

olsen232:
Rather than add another "sno status" command with the same logic but which outputs json, I tried to reuse the same logic but at every step, return the information gathered in the local function either in JSON or in text, depending on which is required.

Still TODO (not in this PR):
- Maybe add a repoStatus field that summarizes the state of the entire repo ie empty, detached head, no working copy, working copy but no diff, diff.
- Maybe put the repoStatus in the process exit code.

### DIFF
--- a/sno/status.py
+++ b/sno/status.py
@@ -1,34 +1,132 @@
 import click
 import pygit2
 
+from .cli_util import MutexOption
 from .structure import RepositoryStructure
+from .output_util import merge_outputs, print_output
+
+
+JSON_DEFAULT_ATTRS = {
+    "commit": None,
+    "branch": None,
+    "upstream": None,
+    "workingCopy": None,
+}
 
 
 @click.command()
 @click.pass_context
-def status(ctx):
+@click.option(
+    "--text",
+    "is_output_json",
+    flag_value=False,
+    default=True,
+    help="Get the status in text format",
+    cls=MutexOption,
+    exclusive_with=["json"],
+)
+@click.option(
+    "--json",
+    "is_output_json",
+    flag_value=True,
+    help="Get the status in JSON format",
+    cls=MutexOption,
+    exclusive_with=["text"],
+)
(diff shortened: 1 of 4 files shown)


rcoup:
As a bigger question, I'm wondering whether we split out interface from logic consistently? So assembling JSON structures as the internal interface (and using that for `--json`), then converting the console UI to formatting the JSON structure as text.

For some commands (eg. log) we'll likely need to have a streaming interface, or incrementally yield items, but even that would be possible via this approach — use a generator internally for a per-item sequence, then for the external APIs we emit it incrementally using a streaming JSON lib.

And I guess we'll (like git) end up having some commands for specific purposes for building tools on as well as the general-purpose do-everything commands used more for human interaction.

olsen232:
I rewrote it so that text output is now a transform of the json output - it is cleaner. Good thought.
The text transform layer could live in another file if necessary one day. 

Added tests. 

Some tests are turned off right now due to raise pytest.skip() in conftest.py?

sno/status.py
```diff
@@ -1,102 +1,206 @@
...
+import sys
+
 import click
 import pygit2
 
+from .cli_util import MutexOption
 from .structure import RepositoryStructure
 
 
+EMPTY_REPO_JSON = frozenset({
```
craigds: the goal being to have an immutable dict-y thing?

I reckon this is too unclear; if you _really_ need an immutable dict you can install https://pypi.org/project/frozendict/

but you're only using this once, and it's uppercased so it's clear it's supposed to be constant. A regular dict seems fine

craigds: Last release of that thing was in 2016 too, might not be great to use

craigds: https://pypi.org/project/immutabledict/ is better

olsen232: Found a better compromise. Done

PR #38, opened 2020-03-24
Add json output for sno commit.

olsen232:
Update text and JSON output for sno commit (added --json flag):

Sample text output:
```
[master e4a9193] commit message
  wellington_building_outlines/
    modified:   1 feature
  Date: 2020-03-25 13:29:46 +1300
```

Sample JSON output:
```
{
  "sno.commit/v1": {
    "commit": "e4a9193153eb27f2bd38f2226951d536de11f5bc",
    "abbrevCommit": "e4a9193",
    "branch": "master",
    "message": "commit message",
    "changes": {
      "wellington_building_outlines": {
        "schemaChanges": null,
        "featureChanges": {
          "modified": 1,
          "new": 0,
          "deleted": 0
        }
      }
    },
    "commitTime": "2020-03-25T00:29:46Z"
  }
}
```


rcoup:
> 1. When --json is specified, it json.dump's the result of the commit to stdout:

👍 

> 2. It changes the print statements in the called functions to log.info statements, so that they don't mess up the json output.

👍 

> But, when JSON mode is not specified, it tees log.info statements from "sno.structure" to stdout. This is then the text output, just as it was before (the tree SHA no longer shows up twice however).

I think we drop the noise altogether and just make output match git? The tree ID isn't relevant to the end-user, it was internal debugging.

```console
$ git commit
[windows-1 8c83941] Merge branch 'master' into windows-1
 Date: Tue Mar 24 13:29:36 2020 +0000
```

sno/status.py
```diff
@@ -44,12 +44,13 @@ def get_status_json(repo):
 
 
 def get_branch_status_json(repo):
-    output = {"commit": None, "branch": None, "upstream": None}
+    output = {"commit": None, "abbrevCommit": None, "branch": None, "upstream": None}
```
rcoup: re terminology. upstream → remote?

olsen232:
PTAL - Removed dependency on arrow again

sno/commit.py
```diff
@@ -75,15 +95,32 @@ def commit(ctx, message, message_file, allow_empty):
...
     if not commit_msg:
         raise click.Abort()
 
-    new_commit = rs.commit(wcdiff, commit_msg, allow_empty=allow_empty)
+    rs.commit(wcdiff, commit_msg, allow_empty=allow_empty)
+    new_commit = repo.head.peel(pygit2.Commit)
+
+    branch = None if repo.head_is_detached else repo.branches[repo.head.shorthand].shorthand
+    jdict = {
+        "sno.commit/v1": {
```
rcoup: I think we should _try_ and mirror [git](https://git-scm.com/docs/gitglossary) and/or [libgit's](https://libgit2.org/libgit2/#HEAD) terminology here as much as possible.

(no particular comments here, I think these are all fine)

sno/commit.py
```diff
@@ -147,5 +193,29 @@ def get_commit_message(repo, wc_changes):
...
+    commit = jdict["abbrevCommit"]
+    message = jdict["message"].replace("\n", " ")
+    diff = diff_status_to_text(jdict["changes"])
+    datetime = to_long_local_iso8601(from_short_iso8601(jdict["commitTime"]))
+    return f"[{branch} {commit}] {message}\n{diff}\n  Date: {datetime}"
+
+
+def to_short_iso8601(datetime):
+    """Returns a string like: 2020-03-26T09:10:11Z"""
+    return datetime.strftime("%Y-%m-%dT%H:%M:%SZ")
```
rcoup: isn't this just `f"{datetime.isoformat()}Z"`

sno/commit.py
```diff
@@ -147,5 +193,29 @@ def get_commit_message(repo, wc_changes):
...
+
+
+def from_short_iso8601(s):
+    return datetime.fromisoformat(s.replace("Z", "+00:00"))
+
+
+def to_long_local_iso8601(datetime):
+    """Returns a string like: 2020-03-26 21:10:11 +12:00"""
```
rcoup: It's valid ISO8601 with or without the `T`/`Z`/etc, so why not have human (ctime) and iso8601, rather than two flavours of the latter? For human timestamps, If we wanted to take into account `Commit.commit_time_offset` too:

```python
def human_committer_time(commit_time, commit_time_offset):
    class CommitTZ(tzinfo):
        def utcoffset(self, dt): return timedelta(minutes=commit_time_offset)
    return datetime.utcfromtimestamp(timestamp).astimezone(CommitTZ()).strftime("%c %z")
```

(I _think_ git normally displays committer-times rather than local times?)

sno/status.py
rcoup: heh, in my review this comment was under that one ⇩

sno/commit.py
rcoup: datetime is a terrible variable name though

olsen232: I think it is datetime.isoformat().replace("+00:00", "Z") -- is that better?

I was a bit uncompromising on my vision - I started with an idea of how I wanted timestamps to look in JSON, and then forced poor datetime.datetime to comply.

rcoup: > I think it is datetime.isoformat().replace("+00:00", "Z") -- is that better?

You're writing it, any further and I'm totally bikeshedding.

> I was a bit uncompromising on my vision

👍 As long as we're consistent everywhere, that's fine.

sno/status.py
olsen232: Will change this in another PR. I think I need to change "schemaChanges" to "metaChanges" as well somewhere, for consistency.

sno/commit.py
olsen232: Now including commit_time_offset in json output, and use it for formatting the time in text output.

PR #41, opened 2020-03-30
Add JSON output for git branch. Only when listing branches.

olsen232:
Something like:
```
{
    "sno.branch/v1": {
        "current": "newbie",
        "branches": {
            "master": {
                "commit": "abcdef1234567890",
                "abbrevCommit": "abcdef",
                "branch": "master",
                "upstream": {
                    "branch": "myremote/master",
                    "ahead": 0,
                    "behind": 0,
                },
            },
            "newbie": {
                "commit": "1234567890abcdef",
                "abbrevCommit": "123456",
                "branch": "newbie",
                "upstream": None,
            },
        },
    }
}
```

### DIFF
--- a/sno/branch.py
+++ b/sno/branch.py
@@ -1,0 +1,86 @@
+import json
+import sys
+
+import click
+import pygit2
+
+from .cli_util import MutexOption
+from .exec import execvp
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.pass_context
+@click.option(
+    "--text",
+    "is_output_json",
+    flag_value=False,
+    default=True,
+    help="Get the status in text format",
+    cls=MutexOption,
+    exclusive_with=["json"],
+)
+@click.option(
+    "--json",
+    "is_output_json",
+    flag_value=True,
+    help="Get the status in JSON format",
+    cls=MutexOption,
+    exclusive_with=["text"],
+)
+@click.argument("args", nargs=-1, type=click.UNPROCESSED)
+def branch(ctx, is_output_json, args):
+    """ List, create, or delete branches """
+    repo_path = ctx.obj.repo_path
+    repo = ctx.obj.repo
+
+    sargs = set(args)
+    if is_output_json:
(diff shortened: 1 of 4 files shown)


sno/branch.py
```diff
@@ -0,0 +1,86 @@
...
+import click
+import pygit2
+
+from .cli_util import MutexOption
+from .exec import execvp
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.pass_context
+@click.option(
```
rcoup, line 13: Could we make this a single decorator we re-use?

eg:
```python
def sno_output_format_options(func):
  """ Apply Text & JSON output format options to a Click command """
  func = click.option("--text", ...)(func)
  func = click.option("--json", ...)(func)
  return func

@click.command(context_settings=dict(ignore_unknown_options=True))
@click.pass_context
@sno_output_format_options
@click.argument("args", nargs=-1, type=click.UNPROCESSED)
def branch(ctx, is_output_json, args):
  ...
```

sno/branch.py
```diff
@@ -0,0 +1,86 @@
...
+)
+@click.argument("args", nargs=-1, type=click.UNPROCESSED)
+def branch(ctx, is_output_json, args):
+    """ List, create, or delete branches """
+    repo_path = ctx.obj.repo_path
+    repo = ctx.obj.repo
+
+    sargs = set(args)
+    if is_output_json:
+        if sargs - {"--list"}:  # sno branch --list --json is allowed.
```
rcoup, line 38:
```suggestion
        if "--list" not in sargs:  # sno branch --list --json is allowed.
```

sno/branch.py
olsen232: Will do this in a separate PR

sno/branch.py
olsen232: This has a different behavior. Here is what is allowed:
git branch --json
git branch --list --json 

Here is what is not allowed:
git branch -D delete_me --json

So I'm testing if sargs is empty, or just contains list. I'll rearrange to make it a bit clearer.

PR #46, opened 2020-04-15
Add exit codes that indicate what went wrong

olsen232:
Useful for any app built on top of sno - until now, any error caused an exit code of either 1 or 2 (except in `sno diff --exit-code` where all errors caused exit code 2).

See exceptions.py for a list of exit codes so far. More can be added later as needed.

INVALID_ARGUMENT = 10

INVALID_OPERATION = 20

NOT_YET_IMPLEMENTED = 30

NOT_FOUND = 40
NO_REPOSITORY = 41
NO_DATA = 42
NO_BRANCH = 43
NO_CHANGES = 44
NO_WORKING_COPY = 45
NO_USER = 46
NO_IMPORT_SOURCE = 47
NO_TABLE = 48

SUBPROCESS_ERROR = 50

UNCATEGORIZED_ERROR = 99

### DIFF
--- a/sno/branch.py
+++ b/sno/branch.py
@@ -5,6 +5,7 @@ import click
 import pygit2
 
 from .cli_util import do_json_option
+from .exceptions import InvalidOperation
 from .exec import execvp
 
 
@@ -33,7 +34,7 @@ def branch(ctx, do_json, args):
     if sargs & {"-d", "--delete", "-D"}:
         branch = repo.head.shorthand
         if branch in sargs:
-            raise click.ClickException(
+            raise InvalidOperation(
                 f"Cannot delete the branch '{branch}' which you are currently on."
             )
 
--- a/sno/checkout.py
+++ b/sno/checkout.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import click
 import pygit2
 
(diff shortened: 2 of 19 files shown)


rcoup:
1. I think `sno diff --exit-code` should match `git diff --exit-code` behaviour.

Others are all pretty minor suggestions, I don't feel strongly about them:

2. `2` is pretty standard for bad-syntax/arguments/options IME?
3. why not make `UNCATEGORIZED_ERROR` = `1`?
4. subprocess-launch-error `128` ?
5. subprocess error: `128+(subprocess exit code)?

olsen232:
1. I think sno diff --exit-code should match git diff --exit-code behaviour.
Yup this is done. 0 is same, 1 is diff.

2. 2 is pretty standard for bad-syntax/arguments/options IME?
Made INVALID_ARGUMENT 2

3. why not make UNCATEGORIZED_ERROR = 1?
This would work except in --exit-code mode, in which case, it would need a different value.
Seems easiest just to make it the same number all the time.
Made it 11 which is a very one-ish number.

4. subprocess-launch-error 128 ?
subprocess error: `128+(subprocess exit code)?
Done

PR #48, opened 2020-04-15
Expand 'show' command to display patches

craigds:
`sno show` already exists, but takes no options and actually just displays a text log entry for the given commit. This PR expands it to work much more like `git show`.

Shows a commit (default `HEAD`) as a patch, in either text
(a la `git show`) or JSON (with `--json`).

HTML and 'quiet' output could be added in future.

# To do

* [x] add tests

# Output

The patch is a diff plus some meta-information. The diff
is just the same as the `sno diff` output for the relevant
commit.

In text mode, the meta-information is the same as what
git show outputs:

```
commit ad9797ea83000864c2ab98c421d2d5256a12db8f
Author: Craig de Stigter <craig@destigter.nz>
Date:   Wed Apr 15 13:19:16 2020 +1200

    commit

<diff output follows>
```

In JSON mode, the metadata is added to the top level JSON object
so that the patch is also a diff. It looks like:

```json
{
  "sno.diff/v1": {...},
  "sno.patch/v1": {
    "authorName": "Craig de Stigter",
    "authorEmail": "craig@destigter.nz",
    "authorTime": "2020-04-15T01:19:16Z",
    "authorTimeOffset": "+12:00",
    "message": "commit"
  }
}
```

### DIFF
--- a/sno/diff.py
+++ b/sno/diff.py
@@ -351,79 +351,25 @@ class Diff(object):
         return self._datasets.values()
 
 
-@click.command()
-@click.pass_context
-@click.option(
-    "--text",
-    "output_format",
-    flag_value="text",
-    default=True,
-    help="Get the diff in text format",
-    cls=MutexOption,
-    exclusive_with=["html", "json", "geojson", "quiet"],
-)
-@click.option(
-    "--json",
-    "output_format",
-    flag_value="json",
-    help="Get the diff in JSON format",
-    hidden=True,
-    cls=MutexOption,
-    exclusive_with=["html", "text", "geojson", "quiet"],
-)
-@click.option(
-    "--geojson",
-    "output_format",
-    flag_value="geojson",
-    help="Get the diff in GeoJSON format",
-    cls=MutexOption,
-    exclusive_with=["html", "text", "json", "quiet"],
-)
-@click.option(
-    "--html",
-    "output_format",
-    flag_value="html",
-    help="View the diff in a browser",
-    hidden=True,
-    cls=MutexOption,
-    exclusive_with=["json", "text", "geojson", "quiet"],
-)
-@click.option(
-    "--quiet",
-    "output_format",
-    flag_value="quiet",
-    help="Disable all output of the program. Implies --exit-code.",
-    cls=MutexOption,
-    exclusive_with=["json", "text", "geojson", "html"],
-)
-@click.option(
-    "--exit-code",
-    is_flag=True,
-    help="Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.",
-)
-@click.option(
-    "--output",
-    "output_path",
-    help="Output to a specific file/directory instead of stdout.",
-    type=click.Path(writable=True, allow_dash=True),
-)
-@click.argument("args", nargs=-1)
-def diff(ctx, output_format, output_path, exit_code, args):
+def diff_with_writer(ctx, diff_writer, *, output_path='-', exit_code, args):
     """
-    Show changes between commits, commit and working tree, etc
+    Calculates the appropriate diff from the arguments,
+    and writes it using the given writer contextmanager.
 
-    sno diff [options] [--] [<dataset>[:pk]...]
-    sno diff [options] <commit> [--] [<dataset>[:pk]...]
-    sno diff [options] <commit>..<commit> [--] [<dataset>[:pk]...]
+      ctx: the click context
+      diff_writer: One of the `diff_output_*` contextmanager factories.
+                   When used as a contextmanager, the diff_writer should yield
+                   another callable which accepts (dataset, diff) arguments
+                   and writes the output by the time it exits.
+      output_path: The output path, or a file-like object, or the string '-' to use stdout.
+      exit_code:   If True, the process will exit with code 1 if the diff is non-empty.
+      args:        The arguments given on the command line, including the refs to diff.
     """
     from .working_copy import WorkingCopy
     from .structure import RepositoryStructure
 
-    if output_format == "quiet":
-        exit_code = True
-
     try:
-        if output_path and output_path != "-":
+        if isinstance(output_path, str) and output_path != "-":
             output_path = Path(output_path).expanduser()
 
         repo = ctx.obj.repo
@@ -485,7 +431,6 @@ def diff(ctx, output_format, output_path, exit_code, args):
         if paths:
             all_datasets = set(filter(lambda dsp: dsp in paths, all_datasets))
 
-        diff_writer = globals()[f"diff_output_{output_format}"]
         writer_params = {
             "repo": repo,
             "base": base_rs,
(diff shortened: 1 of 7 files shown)


sno/diff.py
```diff
@@ -565,18 +575,12 @@ def _out(dataset, diff):
...
-            )
-        else:
-            pecho["file"] = output_path.open("w")
-
-        pecho["color"] = False
+    fp = resolve_output_path(output_path)
+    pecho = {'file': fp, 'color': fp.isatty()}
+    if isinstance(output_path, Path) and output_path.is_dir():
+        raise click.BadParameter(
+            "Directory is not valid for --output + --text", param_hint="--output"
```
rcoup, line 582:
```suggestion
            "Directory is not valid for --output with --text", param_hint="--output"
```

sno/diff.py
```diff
@@ -734,6 +738,33 @@ def _out(dataset, diff):
     yield _out
 
 
+def resolve_output_path(output_path):
+    if isinstance(output_path, io.IOBase):
+        return output_path
+    elif (not output_path) or output_path == "-":
+        return sys.stdout
+    else:
+        return output_path.open("w")
```
rcoup, line 747: Should we use a mode here?

sno/show.py
```diff
@@ -0,0 +1,96 @@
...
+        ctx,
+        patch_writer,
+        output_format,
+        exit_code=False,
+        args=[f"{refish}^..{refish}"],
+    )
+
+
+@contextlib.contextmanager
+def patch_output_text(*, commit_target, output_path, **kwargs):
```
rcoup, line 51: needs docstring, particularly with expected argument types

sno/show.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        click.secho(**pecho)
+        for line in commit_target.message.splitlines():
+            click.secho(f'    {line}', **pecho)
+        click.secho(**pecho)
+        yield diff_writer
+
+
+@contextlib.contextmanager
+def patch_output_json(*, commit_target, output_path, **kwargs):
```
rcoup, line 73: needs docstring, particularly with expected argument types

sno/diff.py
```diff
@@ -344,71 +344,9 @@ def datasets(self):
...
-@click.argument("args", nargs=-1)
-def diff(ctx, output_format, output_path, exit_code, args):
-    """
-    Show changes between commits, commit and working tree, etc
-
-    sno diff [options] [--] [<dataset>[:pk]...]
-    sno diff [options] <commit> [--] [<dataset>[:pk]...]
-    sno diff [options] <commit>..<commit> [--] [<dataset>[:pk]...]
-    """
+def diff_with_writer(
```
rcoup, line 347: needs docstring, particularly with expected argument types/values

sno/diff.py
```diff
@@ -344,71 +344,9 @@ def datasets(self):
...
+    ctx, diff_writer, output_format, *, output_path='-', exit_code, args
```
rcoup, line 348: why does it need a format-writer _and_ an output-format value?

sno/diff.py
rcoup: output_path can be a directory as well (eg: for geojson files)

craigds:
> Possible to split the refactoring from the new additions into separate commits?

Sure I can split it up a bit.

> Feels like we'll need the console pager to work reliably cross-platform though if we're including massive diffs.

I'm not really sure what this refers to; is there some context somewhere? The existing `sno diff` output  isn't much smaller than the output of `sno show`, so it sounds like if the console pager isn't working well we should fix that regardless. That can go in a separate change though?

sno/diff.py
craigds: >  Should we use a mode here?

I assume you mean "a" for append-mode. I think probably not? For most formats (json, geojson, html) it doesn't make sense to append. I don't know why you'd want to for text either. if you need to you can just use `-` and then use `>> outfile` to do it in your shell

craigds: >  output_path can be a directory as well (eg: for geojson files)

this function can't handle that, but `diff_output_geojson` deals with it specially

rcoup: > I assume you mean "a" for append-mode.

I meant text-mode with an encoding, since we're writing text?

rcoup:
> I'm not really sure what this refers to; is there some context somewhere?

I'm not sure the console pager works at all unless we're exec'ing git.

> , so it sounds like if the console pager isn't working well we should fix that regardless. That can go in a separate change though?

Can you ticket?

rcoup:
Rethinking:

> This PR expands it to work much more like git show.
> The existing sno diff output isn't much smaller than the output of sno show

I think we need to revisit what the goal is here — _why_ are we making this change? "matching git" isn't automatically desirable, we _want_ to avoid a pile of historic git mis-steps _and_ make it usable for _our audience_. Datasets aren't code.

_Personally_ i don't see the point of `sno show` returning a diff, seems like as an (effective) alias of `log -1` it's more useful than as an alias of `log -1 -p`. For me that shows the context of "what's the last change made? where did I get up to?", and I feel 100s of lines of IDs doesn't seem particularly helpful in that regard? Though I'm but one user, and I have my own ways of working that are different too. Thoughts @hamishcampbell?

sno/diff.py
craigds: ah right. 

I don't think this is an issue nowadays? Everything defaults to UTF-8 in recent python versions, even if you have other locale settings:

```

LANG="en_NZ.latin-1" LC_COLLATE="en_NZ.latin-1" LC_CTYPE="en_NZ.latin-1" LC_MESSAGES="en_NZ.latin-1" LC_MONETARY="en_NZ.latin-1" LC_NUMERIC="en_NZ.latin-1" LC_TIME="en_NZ.latin-1" LC_ALL= ipython3

...

>>> import sys, locale, os
>>> locale.getpreferredencoding(True), sys.getdefaultencoding()
('UTF-8', 'utf-8')

```

craigds:
i'm a bit confused about it... if you don't use show to show what's in your commit, how would you view what's in your commit?

I guess `sno diff {sha1}^..{sha1}` would work, but it's annoying to type, and that won't show the log message, so to get all the context for what's in the commit you need to run two commands? 

Why not just make show do what git does, and show the whole commit?
or are you saying text diffs aren't useful at all for data?

craigds:
talked with @rcoup out-of-band, he said:

> Sure, ok.
> 
> I guess I pretty rarely use `show {sha}`, I’m more likely to do `log -p`. `show` (no args) I kinda use to see what the last change is and whether it’s mine 😀 

To which I commented

> I use them both in tandem - I have a git log open in one screen, and when i find a commit i think i might care about, i use git show {sha} in a separate window


IMHO since there's no objective or consensus reason to differ from general git behaviour here, we should stick closely to what git does.

As @rcoup pointed out, it does seem important to implement an output pager so that the commit message and author info is visible on screen. That's tracked in #50 and can be implemented as a separate effort

sno/diff.py
```diff
@@ -835,19 +925,13 @@ def diff_output_html(*, output_path, repo, base, target, dataset_count, **kwargs
...
 
     fo.write(
         template.substitute({"title": title, "geojson_data": json_data.getvalue()})
     )
     if fo != sys.stdout:
         fo.close()
-
-    if not output_path:
-        webbrowser.open_new(f"file://{html_path}")
+        webbrowser.open_new(f"file://{output_path}")
```
olsen232, line 937: Seems like this will fail for relative paths

sno/show.py
```diff
@@ -0,0 +1,148 @@
...
+@click.option(
+    "--json",
+    "--patch",
+    "-p",
+    "output_format",
+    flag_value="json",
+    help="Show commit in JSON patch format",
+    cls=MutexOption,
+    exclusive_with=["text"],
+)
```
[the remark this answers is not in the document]
craigds: Discussed out-of-band: going to leave this as-is because:

* we're likely to add HTML output to this command shortly (we already have HTML diff output, so it's just a matter of adding a header with commit information)
* the aliases `-p` and `--patch` are helpful (since JSON output is also the way you generate an applyable patch), so we don't want to remove those

PR #49, opened 2020-04-16
Refactor diff.py so RepositoryStructure needn't contain a commit.

olsen232:
A pure refactoring change - things should be the same as before.
However, also written so that - if needed - we can change
RepositoryStructure to not wrap a commit, but only a tree.
More details follow.

I experimented with a change so that RepositoryStructures could
point at any particular tree, not just at any particular commit.
This could be useful for doing diffs against the empty tree, which
could be used to `sno show` the first commit in a repository.
(sno show always delegates to sno diff, and the first commit cannot
be diffed right now since there are no other commits to diff against.)

(Or it could possibly be useful to do arbitrary diffs between
any two repository structures, without reference to their
shared history. I don't know how useful this might be, but git
diff can diff arbitrary trees).

However, leaving the usefulness of that change aside for a moment:

The refactoring required turned out to be a useful
little refactoring on its own, so I've broken it off into its own
change - which you see here. Allowing diffs of trees as well as
commits can be discussed separately.

### DIFF
--- a/sno/diff.py
+++ b/sno/diff.py
@@ -20,7 +20,6 @@ from .exceptions import (
     NotYetImplemented,
     NotFound,
     NO_WORKING_COPY,
-    NO_COMMIT,
     UNCATEGORIZED_ERROR,
 )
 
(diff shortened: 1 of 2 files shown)


sno/diff.py
```diff
@@ -428,97 +427,90 @@ def diff(ctx, output_format, output_path, exit_code, args):
...
+            elif merge_base_id not in (base_rs.id, target_rs.id):
+                # this needs a 3-way diff and we don't support them yet
+                raise NotYetImplemented(f"Sorry, 3-way diffs aren't supported yet.")
 
-        base_rs = RepositoryStructure(repo, commit_base)
         all_datasets = {ds.path for ds in base_rs}
-
-        target_rs = RepositoryStructure(repo, commit_target or commit_head)
-        all_datasets |= {ds.path for ds in target_rs}
+        if target_rs:
```
craigds, line 484: AFAICT this is always truthy, it's set to a RespositoryStructure instance in both branches above, and `bool(rs)` is always True

olsen232: Done

PR #53, opened 2020-04-19
Allow diffs against the empty tree

### LINKED ISSUE #44: sno show/diff doesn't work for initial import commits
There doesn't seem to be a way to view the content of an import commit.

Behold a fledgling sno repository:

```
$ sno log
commit 736a79dcb844a55e88813635b3535dd227635a87 (HEAD -> master)
Author: Craig de Stigter <craig@destigter.nz>
Date:   Tue Apr 7 13:55:50 2020 +1200

    Import from manawatu-whanganui-palmerston-north-lidar-index-tiles-2018.gpkg to /manawatu_whanganui_palmerston_north_lidar_index_tiles_2018/
```

None of the following commands produce much to show the commit's contents:

```
cdestigter@behindyou  $ sno diff 736a79dcb844a55e88813635b3535dd227635a87
```

```
cdestigter@behindyou  $ sno diff HEAD^..HEAD
Usage: sno diff [OPTIONS] [ARGS]...

Error: Invalid value for commit: Invalid commit spec
```

```
cdestigter@behindyou  $ sno show
commit 736a79dcb844a55e88813635b3535dd227635a87 (HEAD -> master)
Author: Craig de Stigter <craig@destigter.nz>
Date:   Tue Apr 7 13:55:50 2020 +1200

    Import from manawatu-whanganui-palmerston-north-lidar-index-tiles-2018.gpkg to /manawatu_whanganui_palmerston_north_lidar_index_tiles_2018/
```

olsen232:
Right now `sno diff` only works between commits. This means there is no way to diff the first commit against the empty repository.

Fixes #44.

This change makes it so a RepositoryStructure can be a commit, or a tree - since the commit is mostly not used except as a way to get a particular tree.
This means getting a diff from the first commit to the empty repository can be done as follows:

`sno diff 4b825dc642cb6eb9a060e54bf8d69288fbee4904...HEAD`

where `4b825dc642cb6eb9a060e54bf8d69288fbee4904` is the hash of the empty repository.

This is not very easy to remember, but we could still build upon this change to make any of the following:
- Make `sno show` show this diff if you sno show the first commit. This works in git.
- Make an alias for the empty tree, eg `sno diff EMPTY...HEAD`
- Make `HEAD^` etc an alias for the empty tree if they would otherwise refer to the commit before the first commit.

Small change, but structural change: will wait for rcoup@ to review to make sure this makes sense to him.

### DIFF
--- a/sno/structure.py
+++ b/sno/structure.py
@@ -28,21 +28,35 @@ class RepositoryStructure:
         try:
             obj = repo.revparse_single(key)
         except KeyError:
-            raise NotFound(f"Lookup failed for {key}", exit_code=NO_COMMIT)
+            raise NotFound(f"{key} is not a commit or tree", exit_code=NO_COMMIT)
 
         try:
-            return RepositoryStructure(repo, commit=obj)
+            return RepositoryStructure(repo, commit=obj.peel(pygit2.Commit))
         except pygit2.InvalidSpecError:
-            raise NotFound(f"{key} is not a commit", exit_code=NO_COMMIT)
+            pass
 
-    def __init__(self, repo, commit=None):
+        try:
+            return RepositoryStructure(repo, tree=obj.peel(pygit2.Tree))
+        except pygit2.InvalidSpecError:
+            pass
+
+        raise NotFound(
+            f"{key} is a {obj.type_str}, not a commit or tree", exit_code=NO_COMMIT
+        )
+
+    def __init__(self, repo, commit=None, tree=None):
         self.L = logging.getLogger(__class__.__qualname__)
         self.repo = repo
 
+        # If _commit is not None, self.tree -> self._commit.tree, so _tree is not set.
         if commit is not None:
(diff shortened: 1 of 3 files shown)


craigds:
LGTM. I don't expect anyone to be using `4b825dc642cb6eb9a060e54bf8d69288fbee4904` explicitly, but the main thing is that this enables `sno show <first-commit>`, which currently doesn't work because it requires a diff from the empty tree.

craigds:
> Make sno show show this diff if you sno show the first commit. This works in git.

If you merge master into this branch that should work now (since #48 was merged)

olsen232:
Re: parametrizing. I broke up the test and gave it more parametrization, so it's not such a mess. But, not to the extent you suggested - it seemed wasteful to set up a fresh repository six times, just to assert that the diff of a commit with itself is the empty diff six times.

PR #54, opened 2020-04-20
JSON syntax highlighting for all current JSON-output commands

craigds:
Produces syntax-highlighted JSON output for all JSON-output commands.

Adds a dependency on [Pygments](https://pypi.org/project/Pygments/).

Note: This also turns on indented JSON in all circumstances (previously the `diff`/`show` commands produced compact JSON with `--output FILENAME`.) I suspect indented/pretty is what's desired in most situations.

The syntax highlighting is disabled when output is not a TTY; e.g. when redirected to another command, or sent to a file with `--output FILENAME`. However, the indentation is still enabled.

### Examples of output

`sno branch --json`:

Screen Shot 2020-04-20 at 3 07 36 PM

`sno diff --json HEAD`:

Screen Shot 2020-04-20 at 3 08 35 PM

### DIFF
--- a/sno/output_util.py
+++ b/sno/output_util.py
@@ -1,0 +1,43 @@
+import io
+import json
+import sys
+
+
+def dump_json_output(output, output_path):
+    """
+    Dumps the output to JSON in the output file.
+    """
+
+    pretty = (not output_path) or output_path == "-"
+    fp = resolve_output_path(output_path)
+
+    json_params = {}
+    if pretty:
+        json_params.update({"indent": 2, "sort_keys": True})
+    if pretty and sys.stdout.isatty():
+        # Add syntax highlighting
+        from pygments import highlight
+        from pygments.lexers import JsonLexer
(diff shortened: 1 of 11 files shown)


craigds:
Looks like this adds ~1MB to the size of the installer packages (e.g. deb is 76.4 MB , increased from 75.4 MB). I assume that's okay

sno/output_util.py
```diff
@@ -0,0 +1,43 @@
...
+import json
+import sys
+
+
+def dump_json_output(output, output_path, pretty=True):
+    """
+    Dumps the output to JSON in the output file.
+    """
+
+    pretty = (not output_path) or output_path == "-" or output_path == sys.stdout
```
olsen232, line 11: You could just set pretty = (fp == sys.stdout) if you resolve the output path first.

craigds: removed this line altogether; `pretty` is a kwarg so this was just overriding it

PR #59, opened 2020-04-23
Very basic support for resolving merge conflicts.

olsen232:
Allows users to resolve a merge conflict, provided -
- they do it in one go, while `sno merge` is running
- they do it in an interactive terminal (ie, not from a script)
- the only merge conflicts are features, not metadata or schemas
- they resolve each feature one by one
- they are happy for the resolution for each feature to be one of the three pre-existing versions at the time of merge - ancestor, ours, or theirs (ie, there is no way for the resolution to be a new feature that merges fields from both ours and theirs)

Much more to do, but this can be built upon.

Screen Shot 2020-04-23 at 4 18 20 PM

### DIFF
--- a/sno/conflicts.py
+++ b/sno/conflicts.py
@@ -1,0 +1,244 @@
+import sys
+from collections import namedtuple
+
+import click
+
+from .diff_output import repr_row
+from .exceptions import InvalidOperation, NotYetImplemented
+from .structure import RepositoryStructure
+
+
+CommitWithReference = namedtuple("CommitWithReference", ("commit", "reference"))
+
+
+def first_true(iterable):
+    """Returns the value from the iterable that is truthy."""
+    return next(filter(None, iterable))
+
+
+def is_interactive_terminal():
+    if sys.stdin.isatty() and sys.stdout.isatty():
+        return True
+    elif sys.stdin.isatty() and not sys.stdout.isatty():
+        raise InvalidOperation(
+            "Redirecting stdout but not stdin breaks the interactive prompts"
+        )
+    return False
+
+
+def interactive_pause(prompt):
+    """Like click.pause() but waits for the Enter key specifically."""
+    click.prompt(prompt, prompt_suffix="", default="", show_default=False)
+
+
+def is_empty_stream(stream):
+    if stream.seekable():
+        pos = stream.tell()
+        if stream.read(1) == "":
+            return True
+        stream.seek(pos)
+    return False
+
+
+def _get_long_label(commit, reference):
+    """Given a commit and a reference (can be None), returns a label"""
+    if reference is not None:
+        return f'"{reference.shorthand}" ({commit.id.hex})'
+    else:
+        return f"({commit.id.hex})"
+
+
+def _safe_get_dataset_for_index_entry(repo_structure, index_entry):
+    """Gets the dataset that a pygit2.IndexEntry refers to, or None"""
+    try:
+        return repo_structure.get_for_index_entry(index_entry)
+    except KeyError:
+        return None
+
+
+def _get_pk_for_index_entry(dataset, index_entry):
+    """Uses a dataset to decode the primary key that a pygit2.IndexEntry refers to"""
+    return dataset.index_entry_to_pk(index_entry)
+
+
+def _safe_get_feature(dataset, pk):
+    """Gets the dataset's feature with a particular primary key, or None"""
+    try:
+        _, feature = dataset.get_feature(pk)
+        return feature
+    except KeyError:
+        return None
+
+
+# We have three versions of lots of objects - the 3 versions are ancestor, ours, theirs.
+# This namedtuple helps us keep track of them all.
+AncestorOursTheirs = namedtuple("AncestorOursTheirs", ("ancestor", "ours", "theirs"))
+
+AncestorOursTheirs.names = AncestorOursTheirs._fields
+AncestorOursTheirs.chars = tuple(n[0] for n in AncestorOursTheirs.names)
+
+
+def aot(*args):
+    """Creates an AncestorOursTheirs - from 3 positional args, an array, a tuple, or a generator."""
+    if len(args) == 1:
+        return AncestorOursTheirs(*tuple(args[0]))
+    else:
+        return AncestorOursTheirs(*args)
+
+
+def resolve_merge_conflicts(repo, merge_index, ancestor, ours, theirs, dry_run=False):
+    """
+    Supports resolution of basic merge conflicts, fails in more complex unsupported cases.
+
+    repo - a pygit2.Repository
+    merge_index - a pygit2.Index containing the attempted merge and merge conflicts.
+    ancestor, ours, theirs - each is a either a pygit2.Commit, or a CommitWithReference.
+    """
+
+    # We have three versions of lots of objects - ancestor, ours, theirs.
+    args3 = aot(ancestor, ours, theirs)
+    commits3 = aot(getattr(arg, "commit", arg) for arg in args3)
+    refs3 = aot(getattr(arg, "reference", None) for arg in args3)
+    labels3 = aot(_get_long_label(*x) for x in zip(commits3, refs3))
+    repo_structures3 = aot(RepositoryStructure(repo, commit=c) for c in commits3)
+
+    conflict_pks = {}
+    for index_entries in merge_index.conflicts:
+        datasets = aot(
+            _safe_get_dataset_for_index_entry(*x)
+            for x in zip(repo_structures3, index_entries)
+        )
+        dataset = first_true(datasets)
+        dataset_path = dataset.path
+        if None in datasets:
+            for i in range(3):
+                presence = "present" if datasets[i] is not None else "absent"
+                click.echo(f"{labels3[i]}: {dataset_path} is {presence}")
+            raise NotYetImplemented(
+                "Sorry, resolving conflicts where datasets are added or removed isn't supported yet"
+            )
+
+        pks3 = aot(_get_pk_for_index_entry(*x) for x in zip(datasets, index_entries))
+        if "META" in pks3:
+            click.echo(f"Merge conflict found in metadata for {dataset_path}")
+            raise NotYetImplemented(
+                "Sorry, resolving conflicts in metadata isn't supported yet"
+            )
+        pk = first_true(pks3)
+        if pks3.count(pk) != 3:
+            click.echo(
+                f"Merge conflict found where primary keys have changed in {dataset_path}"
+            )
+            for i in range(3):
+                click.echo(f"{labels3[i]}: {dataset_path}:{pks3[i]}")
+            raise NotYetImplemented(
+                "Sorry, resolving conflicts where primary keys have changed isn't supported yet"
+            )
+        conflict_pks.setdefault(dataset_path, [])
+        conflict_pks[dataset_path].append(pk)
+
+    num_conflicts = sum(len(pk_list) for pk_list in conflict_pks.values())
+    click.echo(f"\nFound {num_conflicts} conflicting features:")
+    for dataset_path, pks in conflict_pks.items():
+        click.echo(f"{len(pks)} in {dataset_path}")
+    click.echo()
+
+    # Check for dirty working copy before continuing - we don't want to fail after interactive part.
+    ours_rs = repo_structures3.ours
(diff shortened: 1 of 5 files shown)


sno/merge.py
```diff
@@ -34,52 +38,52 @@ def merge(ctx, ff, ff_only, commit):
...
-        print("Already merged!")
+        click.echo("Already merged!")
         return
 
     # We're fastforwardable if we're our own common ancestor.
     can_ff = merge_base_id == c_ours.id
 
     if ff_only and not can_ff:
-        print("Can't resolve as a fast-forward merge and --ff-only specified")
+        click.echo("Can't resolve as a fast-forward merge and --ff-only specified")
```
craigds, line 61: This should be a ClickException of some kind

sno/merge.py
```diff
@@ -187,34 +183,121 @@ def handle_merge_conflicts(
...
+    # At this point, the failure should be dealt with so we can start resolving conflicts interactively.
+    # We don't want to fail during conflict resolution, since then we would lose all the user's work.
+    # TODO: Support other way(s) of resolving conflicts.
+    if sys.stdout.isatty():
+        _interactive_pause(
+            "Press enter to begin resolving merge conflicts, or Ctrl+C to abort at any time..."
+        )
+    else:
+        click.echo(
+            "Printing conflicts but not resolving - merge conflicts must be resolved in an interactive terminal"
```
craigds, line 222: We need to test this... somehow.

Maybe we should add a `-X {ours|theirs|ancestor}` option like `git merge` has, then we can use it to specify the behaviour in testing. That sidesteps the need to test terminal input.

Is `ancestor` ever a valid choice? It makes sense to display it I think, because it can be useful to understand how the two branches have diverged from their starting position, but I can't imagine a reason for ever choosing it to resolve a conflict.

craigds:
This is a great first step!

 I think we need more iteration on it, but probably we can start by getting this merged and improve it in another PR.

I think we need some basic tests for conflict resolution. I don't know how to test terminal input, so I'd suggest adding a way to make conflict resolution non-interactive, perhaps by adding `-X ours` or `-X theirs`

Some suggestions for later (not blockers for this PR):
* highlight what _fields_ differ when displaying the conflicts. possibly display unchanged fields in white instead of green/red
* Provide some way of merging different fields from _both_ `ours` and `theirs`. e.g. if I change `fieldX` and you change `fieldY`, a valid merged feature might include both our changes.

sno/merge.py
rcoup: > Maybe we should add a -X {ours|theirs|ancestor} option like git merge has, then we can use it to specify the behaviour in testing. That sidesteps the need to test terminal input.

👍 

> Is ancestor ever a valid choice? It makes sense to display it I think, because it can be useful to understand how the two branches have diverged from their starting position, but I can't imagine a reason for ever choosing it to resolve a conflict.

"Make both these things go away and I'll fix it myself"?

rcoup:
> I don't know how to test terminal input

IIUC `runner.invoke(input="\n-delimited-responses-to-prompts")`](https://click.palletsprojects.com/en/7.x/testing/#input-streams)

> so I'd suggest adding a way to make conflict resolution non-interactive, perhaps by adding -X ...

do this too 👍

sno/merge.py
```diff
@@ -148,23 +141,26 @@ def first_true(iterable):
...
-    print("Merge conflicts!")
+
+    ANCESTOR, OURS, THEIRS = 0, 1, 2
     # All of the following are 3-tuples of the form (ancestor, ours, theirs)
     commit_args3 = (ancestor, ours, theirs)
-    commits3 = tuple(_get_commit(arg) for arg in commit_args3)
-    labels3 = tuple(_get_commit_label(arg) for arg in commit_args3)
+    commits3 = tuple(arg[0] for arg in commit_args3)
+    refs3 = tuple(arg[1] for arg in commit_args3)
+    labels3 = tuple(_get_long_label(*arg) for arg in commit_args3)
```
rcoup, line 163: maybe namedtuple/classify these so it's cleaner to read & understand? ala:

```python
Head = namedtuple('Head', ['commit', 'ref', 'label')  # i guess also needs repo_structure/dataset/pk
Heads = namedtuple('Heads', ['ancestor', 'ours', 'theirs'])

heads = Heads([
  Head(
    commit=h[0],
    label=_get_long_label(*h),
    ref=h[1]
  ) for h in (ancestor, ours, theirs)
])

...

for h in heads:
  print(h.label)

...
heads.ancestor.commit
```

or maybe simpler:
```python
Heads = namedtuple('Heads', ['ancestor', 'ours', 'theirs'])

heads = Heads(ancestor, ours, theirs)
commits = Heads(*(h[0] for h in heads))
refs = Heads(*(h[1] for h in heads))
labels = Heads(*(_get_long_label(*h) for h in heads))
...
labels.ancestor
refs.theirs
```

sno/merge.py
```diff
@@ -187,34 +183,121 @@ def handle_merge_conflicts(
...
+    # Conflicts are resolved, time to commit
+    assert not merge_index.conflicts
+    merge_tree_id = merge_index.write_tree(repo)
+    click.echo(f"Merge tree: {merge_tree_id}")
+
+    user = repo.default_signature
+    merge_message = "Merge '{}'".format(
+        refs3[THEIRS].shorthand if refs3[THEIRS] else refs3[THEIRS].id.hex
+    )
+    commit_id = repo.create_commit(
```
rcoup, line 252: worth adding `--commit/--no-commit` at this stage? I guess not, since we don't have a persistent index?

sno/merge.py
```diff
@@ -187,34 +183,121 @@ def handle_merge_conflicts(
...
+        user,
+        merge_message,
+        merge_tree_id,
+        [commits3[OURS].id, commits3[THEIRS].id],
+    )
+    click.echo(f"Merge commit: {commit_id}")
+    return commit_id
+
+
+_version_names3 = ("ancestor", "ours", "theirs")
```
rcoup, line 264: could all be `Heads` (🚲🏠-away) namedtuples too 😄

sno/merge.py
```diff
@@ -187,34 +183,121 @@ def handle_merge_conflicts(
...
+    labels3 - labels for each version.
+    """
+    click.secho(f"\n=========== {feature_label} ==========", bold=True)
+    for i in range(3):
+        click.secho(f"{_big_prefixes3[i]} {_version_names3[i]:>9}: {labels3[i]}")
+        if features3[i] is not None:
+            click.secho(repr_row(features3[i], prefix=_prefixes3[i]), fg=_cols3[i])
+
+
+_version_chars3 = ("a", "o", "t")
```
rcoup, line 284: here too...

olsen232:
Yeah we need to think more about a stateful approach for resolving conflicts like git - one where your work so far isn't lost just because you get to something you can't figure out how to deal with right now, and ideally one which lets you view the conflicting geometries in QGis etc - and which somehow lets the sno app show the conflicts too, not sure what that involves. This code maybe just proves that we have the understanding necessary to print and resolve conflicts , but there's no reason not to check it in until we get a better way of doing it. Much of it (basically everything except the prompt itself) will still end up being used, albeit in a modified form

rcoup:
> view the conflicting geometries in QGis etc 

wonder whether we take a similar approach as `diff` (json/geojson) combined with the well-known named files git uses (ala `COMMIT_MESSAGE`, etc)

sno/merge.py
```diff
@@ -23,9 +26,10 @@
         "or the merge can be resolved as a fast-forward."
     ),
 )
+@click.option("--force", "-f", is_flag=True)
```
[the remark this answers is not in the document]
olsen232: Same as in `sno checkout`. Added some help to both.

sno/merge.py
```diff
@@ -91,35 +95,24 @@ def merge(ctx, ff, ff_only, commit):
...
-    if type(commit_or_tuple) is tuple:
-        commit = commit_or_tuple[0]
-        label = commit_or_tuple[1]
-        if label != commit.id.hex:
-            label = f"{label} ({commit.id.hex})"
-        return label
+def _get_long_label(commit, ref):
+    """Given a commit and a reference (can be None), returns a label"""
+    if ref is not None:
+        return f"\"{ref.shorthand}\" ({commit.id.hex})"
```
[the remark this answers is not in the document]
olsen232: Done

sno/merge.py
olsen232: Made a couple of namedtuples -
AncestorOursTheirs - contains 3 versions of basically anything
CommitWithReference - contains (pygit2.Commit, pygit2.Reference)

A bit nicer in places, overall structure is much the same.

sno/merge.py
olsen232: Done

sno/merge.py
olsen232: Done

sno/merge.py
olsen232: Punted this for now - need to think lots more about the stateful way of doing things. Will write up a doc.

sno/conflicts.py
```diff
@@ -0,0 +1,244 @@
...
+    """Gets the dataset that a pygit2.IndexEntry refers to, or None"""
+    try:
+        return repo_structure.get_for_index_entry(index_entry)
+    except KeyError:
+        return None
+
+
+def _get_pk_for_index_entry(dataset, index_entry):
+    """Uses a dataset to decode the primary key that a pygit2.IndexEntry refers to"""
+    return dataset.index_entry_to_pk(index_entry)
```
rcoup, line 61: seems like a needless level of indirection?

sno/conflicts.py
```diff
@@ -0,0 +1,244 @@
...
+
+# We have three versions of lots of objects - the 3 versions are ancestor, ours, theirs.
+# This namedtuple helps us keep track of them all.
+AncestorOursTheirs = namedtuple("AncestorOursTheirs", ("ancestor", "ours", "theirs"))
+
+AncestorOursTheirs.names = AncestorOursTheirs._fields
+AncestorOursTheirs.chars = tuple(n[0] for n in AncestorOursTheirs.names)
+
+
+def aot(*args):
```
rcoup, line 81: fwiw, you can also use namedtuple as a parent class more easily under Py3 by inheriting from `typing.NamedTuple` and then add/override methods/whatever like a normal class.

(what you have is fine though)

sno/conflicts.py
```diff
@@ -0,0 +1,244 @@
...
+        datasets = aot(
+            _safe_get_dataset_for_index_entry(*x)
+            for x in zip(repo_structures3, index_entries)
+        )
+        dataset = first_true(datasets)
+        dataset_path = dataset.path
+        if None in datasets:
+            for i in range(3):
+                presence = "present" if datasets[i] is not None else "absent"
+                click.echo(f"{labels3[i]}: {dataset_path} is {presence}")
```
rcoup, line 116:
```suggestion
            for ds, label in zip(datasets, labels):
                presence = "present" if ds else "absent"
                click.echo(f"{label]}: {dataset_path} is {presence}")
```

sno/merge.py
```diff
@@ -34,187 +39,74 @@ def merge(ctx, ff, ff_only, commit):
...
+                ours=CommitWithReference(c_ours, r_ours),
+                theirs=CommitWithReference(c_theirs, r_theirs),
+                dry_run=dry_run,
             )
         else:
-            print("No conflicts!")
+            click.echo("No conflicts!")
             merge_tree_id = merge_index.write_tree(repo)
-            print(f"Merge tree: {merge_tree_id}")
+            click.echo(f"Merge tree: {merge_tree_id}")
```
rcoup, line 89: Tree IDs aren't meaningful to anyone, change it to log?

tests/test_merge.py
```diff
@@ -215,18 +206,41 @@ def test_merge_conflicts(
...
+        assert r.exit_code == 0, r
+        assert repo.head.target.hex != alternate_commit_id
+        assert repo.head.target.hex != master_commit_id
+
+        r = cli_runner.invoke(["diff", master_commit_id])
+        assert r.exit_code == 0, r
+
+        # Diff merged in from alternate automatically
+        assert feature_name(sample_pks[0]) in r.stdout
+        # Diff merged in from alternate explicitly with "t" for theirs
```
rcoup, line 240: can we have a test for ours & ancestor resolution;
and a test for types of changes (meta, others?) that don't work atm?

PR #64, opened 2020-04-30
Faster tests via a data_archive_readonly fixture

craigds:
The `data_archive` fixture extracts an archive for each test,
even if a previous test used the same archive and didn't modify the
extracted output.

This change adds a new `data_archive_readonly` fixture, that re-uses
the same extracted directory for multiple tests. It uses the pytest cache.

I get consistent results of 215s for a full test run, compared to 230s before, for a win of 15 seconds (6.5%). I was hoping for a bigger win here, but it's better than nothing.

I was hoping to do a similar thing for `data_working_copy`, but all the tests that use that fixture end up writing to either the repo directory or the working copy, so they can't use this unfortunately.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -104,12 +104,16 @@ def cleanup_dir(d):
     try:
         shutil.rmtree(d)
     except PermissionError as e:
-        L.debug("W: Issue cleaning up temporary folder: %s", e)
+        L.debug("Issue cleaning up temporary folder (%s): %s", d, e)
+
+
+def get_archive_path(name):
+    return Path(__file__).parent / "data" / f"{name}.tgz"
 
 
 def extract_archive(name, extract_dir):
     archive_name = f"{name}.tgz"
-    archive_path = Path(__file__).parent / "data" / archive_name
+    archive_path = get_archive_path(name)
     with tarfile.open(archive_path) as archive:
         archive.extractall(extract_dir)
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -101,61 +100,87 @@ def chdir():
     return chdir_
 
 
+def cleanup_dir(d):
+    try:
+        shutil.rmtree(d)
+    except PermissionError as e:
+        L.debug("W: Issue cleaning up temporary folder: %s", e)
```
rcoup, line 107:
```suggestion
        L.debug("W: Issue cleaning up temporary folder (%s): %s", d, e)
```

tests/conftest.py
```diff
@@ -101,61 +100,87 @@ def chdir():
...
+
+    Context-manager produces the directory path and sets the current working directory.
+    """
+
+    @contextlib.contextmanager
+    def ctx(name):
+        root = Path(request.config.cache.makedir('data_archive_readonly'))
+        path = root / name
+        if path.exists():
+            L.info("Found cache at %s", path)
```
rcoup, line 176: How does caching work — if we update the tgz does it automatically flush somehow? Or do we do it based off a hash/something of the archive?

craigds: At present you have to clear the cache via `pytest --cache-clear ...`

We'd have to read the whole thing to hash it. Presumably that's still (much) faster than extracting it, but it's not ideal...

But maybe it's worth it to avoid confusion if/when the archives change

craigds: Done in 15dbaa8b927a171e13c7b44ed29d7e6d869a8de3

PR #71, opened 2020-05-04
Use hexWKB in JSON diff/show output

### LINKED ISSUE #62: Don't use GeoJSON for geometry representations in diffs
(Extracting this discussion from https://github.com/koordinates/sno/pull/61#discussion_r414505669 )

Currently, JSON diffs and patches from sno represent geometries as GeoJSON.

However, GeoJSON doesn't support curved geometries, which means we won't support the full simple features spec.

Whatever we use:
 * should support empty geometries, circular geometries, and geometry collections
 * doesn't need to include CRS information (that's a layer-level concern, doesn't need to be included on each feature)

Obvious alternatives include:

* WKT
* Hex-encoded WKB
* some other encoding of WKB (e.g. base64). (Non-standard, but 50% more compact than hex-encoded)

Notes:

* WKT has the advantage of being somewhat human-readable.
* WKB has the advantage of being lossless, and _also_ not requiring parsing in some situations (i.e. we can read straight from PostGIS/GPKG and encode in hex without going through OGR)

I suggest we use hex-encoded WKB since I don't think anyone cares much about readability of geometry blobs, and I don't think the 50% size reduction of base64-encoding is worth being different to everyone else.

craigds:
This replaces GeoJSON geometries with (little-endian) hex-encoded WKB
in `show --json` and `diff --json` commands. `--geojson` output is
unaffected.

This:
 * reduces patch/diff size by quite a lot, especially for large
geometries
 * makes patch generation/consumption more efficient; no pass via OGR is
   required in most circumstances.
 * improves support for exotic geometries (GeoJSON doesn't support
   curved geometry types)

fixes #62

# Size of output

```
# points, with GeoJSON
$ sno show --json | wc -c
5405


# points, with hexWKB:
$ sno show --json | wc -c
4205
```

If you combine this with `--json-style=extracompact` from #70:

```
# points, with GeoJSON
$ sno show --json --json-style=extracompact | wc -c
2576


# points, with hexWKB:
$ sno show --json --json-style=extracompact | wc -c
2316
```


tests/test_diff.py
```diff
@@ -594,335 +561,85 @@ def test_diff_polygons(output_format, data_working_copy, geopackage, cli_runner)
                 len(odata["sno.diff/v1"]["nz_waca_adjustments"]["featureChanges"]) == 4
             )
             assert odata == {
-                "sno.diff/v1": {
-                    "nz_waca_adjustments": {
-                        "metaChanges": {},
-                        "featureChanges": [
+                'sno.diff/v1': {
```
rcoup, line 564: Need some indicator of what sort of geometry encoding is being used ([see](https://github.com/koordinates/sno/issues/62#issuecomment-620507966))

craigds: I don't think we do. (Let's discuss/resolve this on that thread^ on #62 though)

sno/gpkg.py
```diff
@@ -202,10 +200,66 @@ def geom_to_ogr(gpkg_geom, parse_srs=False):
...
+    elif envelope_typ == 4:
+        wkb_offset += 64
+    elif envelope_typ == 0:
+        pass
+    else:
+        raise ValueError("Invalid envelope contents indicator")
+
+    wkb = gpkg_geom[wkb_offset:]
+
+    if not is_le:
```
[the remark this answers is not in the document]
craigds: yikes, that's crazy. So you could have a big-endian GPKG embedding some little-endian WKB. I'll add a test for that :/

sno/gpkg.py
```diff
@@ -185,14 +186,11 @@ def geom_cols(db, table):
     return tuple(r[0] for r in q)
 
 
-def geom_to_ogr(gpkg_geom, parse_srs=False):
+def _validate_geom(gpkg_geom):
```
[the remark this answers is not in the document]
craigds: once again I've just realised this PR duplicates some effort from the postgis branch :( any ETA on finishing that off?

I've added tests in 0a14e9c058c083d31daf6baeb4d56dc1f64a7508 comprehensively covering envelope and endianness varieties. it doesn't cover SRID since I haven't implemented any handling of that - it produces WKB (not EWKB) and thus strips the SRID.

I haven't covered any exotic curve geometries etc because they don't really seem like edge cases for this, since we don't bother probing inside the WKB itself.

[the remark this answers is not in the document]
rcoup: > any ETA on finishing that off?

After internal data format changes. It got real hard without them.

sno/gpkg.py
```diff
@@ -238,31 +292,44 @@ def geom_to_ogr(gpkg_geom, parse_srs=False):
...
 
-def ogr_to_geom(ogr_geom):
+def ogr_to_gpkg_geom(
+    ogr_geom, _little_endian=True, _little_endian_wkb=True, _add_envelope=False
+):
     """
     Given an OGR geometry object, construct a GPKG geometry value.
     http://www.geopackage.org/spec/#gpb_format
 
     Arbitrarily, this only produces little-endian geometries.
```
rcoup, line 322: Presume the `_*` parameters are for testing? Though the docstring is wrong now.

tests/test_gpkg.py
```diff
@@ -61,3 +66,77 @@ def test_roundtrip_geometry_conversion(wkt):
             srs and orig_srs
         ), f'{srs and srs.ExportToProj4()} != {orig_srs and orig_srs.ExportToProj4()}'
         assert srs.ExportToProj4() == orig_srs.ExportToProj4()
+
+
+@pytest.mark.parametrize(
+    'big_endian,little_endian',
+    [
+        pytest.param(
+            '000000000100000000000000000000000000000000',
```
rcoup, line 75: Can you add WKT in comments?

tests/test_gpkg.py
```diff
@@ -61,3 +66,77 @@ def test_roundtrip_geometry_conversion(wkt):
...
+    gpkg_geom = hex_wkb_to_gpkg_geom(big_endian)
+
+    assert gpkg_geom_to_hex_wkb(gpkg_geom) == little_endian
+
+
+@pytest.mark.parametrize(
+    'input',
+    [
+        pytest.param(
+            b'GP\x00\x01\x00\x00\x00\x00\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00',
```
rcoup, line 104: Can you add WKT (+other relevant stuff) in comments?

sno/diff_output.py
```diff
@@ -283,6 +284,23 @@ def _out(dataset, diff):
...
 
 def _json_row(row, change, pk_field):
+    f = {
+        "geometry": None,
+        "properties": {},
+        "id": f"{change}::{row[pk_field]}",
+    }
+
+    for k in row.keys():
+        v = row[k]
```
rcoup, line 294:
```suggestion
    for k, v in row.items():
```

sno/diff_output.py
```diff
@@ -283,6 +284,23 @@ def _out(dataset, diff):
...
+        if isinstance(v, bytes):
+            f["geometry"] = gpkg.gpkg_geom_to_hex_wkb(v)
```
rcoup, line 296: as we've ditched geojson, should we make it a `{field: value}` dict and drop properties/geometry/etc?

sno/diff_output.py
```diff
@@ -293,8 +311,8 @@ def _json_row(row, change, pk_field):
     for k in row.keys():
         v = row[k]
         if isinstance(v, bytes):
-            g = gpkg.geom_to_ogr(v)
-            f["geometry"] = json.loads(g.ExportToJson())
+            g = gpkg.gpkg_geom_to_ogr(v)
+            f['geometry'] = json.loads(g.ExportToJson())
```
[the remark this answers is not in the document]
craigds: ie some kind of `ogr.ExportToJsonDict()` method?

One problem is that different bindings would have to implement it differently. I'm completely ignorant of how Java handles deserialized JSON but I'm not sure I want to know either...

sno/gpkg.py
craigds: Will fix. Yes, the underscored kwargs are only for testing. I'll make them keyword only at least I guess, and update the docstring

sno/diff_output.py
craigds: I was going to get rid of `id`, but it's kind of useful since when you're looking at a patch you don't know what that PK field is.

`geometry` is sort-of the same, although there's two problems:

- at present you _still_ don't know what the geometry field is
- if we moved it from `geometry` into `properties.<fieldname>` you gain the field name but lose the fact that it's a geometry field

For applying patches, the dataset already has type information so I guess the patch doesn't _have_ to. So I'd tend to just put all properties together, including geometry. So remove _all_ of `geometry`, `id` and `properties` and just make the feature a dict of `fieldname: value`

This raises the question of whether patches should include schema information for conflict detection, so if you change the PK field or the geometry field the patch can be rejected. Thoughts?

That all might be worth a separate effort though; this PR could be just about old enough

craigds: On `id`: I think if we keep `id` then it should cease to be a weird `"U-::2"` thing and should just be the value of the PK field. (This applies to geojson output too). That's a separate PR though.

rcoup: > On id: I think if we keep id then it should cease to be a weird "U-::2" thing and should just be the value of the PK field. (This applies to geojson output too).

`id`'s need to be unique in GeoJSON, so can't be the PK field (because there's a new & old side)?

rcoup: > This raises the question of whether patches should include schema information for conflict detection, so if you change the PK field or the geometry field the patch can be rejected. Thoughts?

I think including the schema hash from the new data format might make sense? That will indicate compatibility, so if the hash doesn't match and there's no schema-change information then you can't really apply it.

rcoup: (but yes, no point doing that until it's implemented)

sno/diff_output.py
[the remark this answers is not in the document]
rcoup: In principle, yes (`ExportToJsonStruct()`?). It's possible to add language-specific methods in SWIG, so could ignore Java in the meantime.

sno/diff_output.py
craigds: > id's need to be unique in GeoJSON, so can't be the PK field (because there's a new & old side)?
okay. So remove the `id` altogether in other formats, but leave it in the existing form for GeoJSON 👍

rcoup:
Can you summarize/include/link examples of whatever formats we have finalised on?

sno/diff-view.html
```diff
@@ -125,7 +125,7 @@
             }).addTo(map)
 
             var layerGroup = L.featureGroup()
-            for (let [dataset, diff] of Object.entries(DATA['sno.diff/v1'])) {
+            for (let [dataset, diff] of Object.entries(DATA['sno.diff/v1+hexwkb'])) {
```
rcoup, line 128: how's it going to view wkb in the browser? Surely needs the GeoJSON output?

tests/test_diff.py
```diff
@@ -1121,7 +1130,7 @@ def _extract(diff_json):
         # changes from HEAD (R1 -> WC)
```
rcoup, line 1130: comments to remove?

sno/diff-view.html
craigds: I did this by accident, so good catch. I'm surprised this is using the _json_ diff and not the _geojson_ diff. That throws a spanner in the works a bit, because I can't just switch tot he geojson one, it's quite different. Will attempt to though.

rcoup: It's using the JSON because the geojson ends up in multiple files

rcoup: Alternatively, it could serialise all the geojson into an array in the HTML or something. No problem doing that, but it needs to work :)

craigds: The `--geojson` diff doesn't actually differentiate multiple datasets, so it's not enough for this.

So I guess to make this work, I need to add a format variant for `--json` diffs with GeoJSON geometries?

tests/test_diff.py
craigds: seems like a fairly useful part of the test AFAICT. the test passes when it's uncommented. I'll uncomment it.

sno/diff-view.html
rcoup: Don’t understand... the geojson gets written to multiple files in a directory?

craigds: aha, so it does (if you have >1 dataset). I just saw it with one dataset and got lost on how the different ones were going to be differentiated, but didn't realise it used multiple _files_

rcoup: otherwise there's no meaningful way to view them in QGIS/etc

craigds: fixed in cb28c3b - it now writes geojson files to a temp dir and then reads them back in again. _Could_ have done it some other way, e.g. by passing an empty dict to `resolve_object_path` and making it create a StringIO for each dataset, but it felt like a Weird Hack so I didn't

sno/diff_output.py
craigds: to clarify, I'm not planning to do that in this PR; we can clean it up in another PR

PR #73, opened 2020-05-06
New class ConflictIndex - like pygit2.Index with serialisable conflicts

olsen232:
The pygit2.Index class can be written to a tree, but only once all conflicts are resolved.
This ConflictIndex can be written to a tree even with unresolved conflicts, which means the resolution of merge conflicts could be done gradually as the user looks into each conflict one by one - the machine can even be restarted mid merge without losing progress.

ConflictIndex is also more modifiable than pygit2.Index - the pygit2 Index is backed by C code and so not every operation is possible, and those which are missing cannot be worked around from within python. For instance, conflicts in pygit2.Index.conflicts can be deleted but not modified. ConflictIndex contains its data in python dicts which can be modified as needed.

ConflictIndex is not used yet but will be used to move away from having to resolve all merge conflicts interactively in a single command.


sno/conflicts.py
```diff
@@ -12,6 +14,113 @@
...
+
+    def __getitem__(self, path):
+        return self.entries[path]
+
+    def __setitem__(self, path, index_entry):
+        assert path == index_entry.path
+        self.entries[path] = index_entry
+
+    def add_conflict(self, key, conflict3):
+        assert type(key) == str
```
craigds, line 68: feels like if these asserts are important they should be turned into

```python
if not isinstance(...):
    raise TypeError("expected ...")
```

sno/structs.py
```diff
@@ -72,9 +72,55 @@ def resolve_refish(repo, refish):
...
+    """
+
+    NAMES = _ANCESTOR_OURS_THEIRS_ORDER
+    CHARS = tuple(n[0] for n in NAMES)
+
+    @staticmethod
+    def of(*args, **kwargs):
+        """
+        Coerces any of the following into an AncestorOursTheirs:
+        3 positional args, 3 keyword args, a tuple, a dict, or a generator.
```
craigds, line 95: Seems unnecessarily loose, confuses just exactly what this thing accepts (it accepts everything!). These could all just fold back to "use either *args or **kwargs" I think.

generators can be starred into args by the caller: `f(*gen())`
dicts can be double-starred by the caller: `f(**mydict)`

If caller can do that then I don't really know what this method adds over the existing namedtuple `__init__`

sno/conflicts.py
```diff
@@ -12,6 +14,113 @@
 L = logging.getLogger("sno.conflicts")
 
 
+class ConflictIndex:
```
craigds, line 17: Is it worth implementing `__eq__` so you can compare two ConflictIndexes? It'd be sane to do that in the roundtrip test you've got. Presumably it'd just compare `entries` and `conflicts`

sno/conflicts.py
```diff
@@ -12,6 +14,113 @@
...
+            if not index_entry:
+                continue
+            result_path = f".conflicts/{key}/{version}/{index_entry.path}"
+            result.append(
+                pygit2.IndexEntry(result_path, index_entry.id, index_entry.mode)
+            )
+        return result
+
+    _PATTERN = re.compile(
+        r"^.conflicts/(?P<key>.*)/(?P<version>ancestor|ours|theirs)/(?P<path>.*)$"
```
craigds, line 86:
```suggestion
        r"^\.conflicts/(?P<key>.+?)/(?P<version>ancestor|ours|theirs)/(?P<path>.+)$"
```

The `?` in the `key` match is important to make it non-greedy, otherwise if `path` contains `"ancestor"` I think `key` would gobble up too much

sno/conflicts.py
olsen232: Done

sno/conflicts.py
olsen232: Done

sno/conflicts.py
olsen232: In the end, I just stopped using IndexEntrys internally - in favour of a namedtuple with exactly the same content. Then __eq__ and __repr__ got a lot easier - otherwise they were quite a mess.

sno/structs.py
olsen232: Removed.

But, I did instead make a little function in conflicts.py which lets you use generators easily, so I can do this:
```
datasets3 = aot(
    _safe_get_dataset_for_index_entry(rs, ie)
    for rs, ie in zip(repo_structures3, index_entries3)
)
```
instead of this:
```
datasets3 = AncestorOursTheirs(
    *(
        _safe_get_dataset_for_index_entry(rs, ie)
        for rs, ie in zip(repo_structures3, index_entries3)
    )
)
```

sno/conflicts.py
```diff
@@ -12,6 +16,157 @@
...
+            for key, conflict3 in enumerate(index.conflicts):
+                self.add_conflict(str(key), conflict3)
+
+    def __eq__(self, other):
+        if not isinstance(other, ConflictIndex):
+            return False
+        return self.entries == other.entries and self.conflicts == other.conflicts
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
craigds, line 62: This method isn't required since python 3.x, `__ne__` delegates to `__eq__` by default

sno/structs.py
```diff
@@ -72,9 +72,39 @@ def resolve_refish(repo, refish):
...
+    commits, repository-structures, datasets, features, primary keys.
+    The 3 versions  are the common ancestor, and 2 versions to be merged, "ours" and "theirs".
+    Like pygit2, we keep the 3 versions always in the same order - ancestor, ours, theirs.
+    """
+
+    NAMES = _ANCESTOR_OURS_THEIRS_ORDER
+    CHARS = tuple(n[0] for n in NAMES)
+
+    @staticmethod
+    def partial(ancestor=None, ours=None, theirs=None, *args):
```
craigds, line 92: not sure why this method accepts `*args` when it just asserts there aren't any

olsen232: This method is supposed to be called with only kwargs - args have the wrong semantics - but without this assert, partial(), partial(1), partial(1, 2) would all be valid, but I have decided that is not how this method is used and so those calls are errors. (Maybe a similar idea to your comment earlier about how I had written an overly accepting method)

sno/conflicts.py
olsen232: Done.

sno/structs.py
craigds: I don't think this stops it being called with args, because args fill up the earlier arguments first even if they're named:

```python
>>>
... def partial(ancestor=None, ours=None, theirs=None, *args):
...     print(ancestor, ours, theirs, args)
...
...
>>> partial()
None None None ()
>>> partial(1)
1 None None ()

```

However, you _can_ specify keyword-only arguments, like:

```python
>>> def partial(*, ancestor=None, ours=None, theirs=None):
...     print(ancestor, ours, theirs)

>>> partial(1)
  File "<ipython-input-14-24db84d21a38>", line 1, in <module>
    partial(1)
TypeError: partial() takes 0 positional arguments but 1 was given

# doesn't help you with zero-args case though:
>>> def partial(*, ancestor=None, ours=None, theirs=None):
...     print(ancestor, ours, theirs)
>>> partial()
None None None
```

PR #77, opened 2020-05-08
App merge conflict flow step 1 - JSON output

olsen232:
Doing a sno-merge should either succeed if there are no conflicts, or return a list / summary of conflicts while putting the repository into a "merging" state.
Then the user can run follow-up commands to see and resolve conflicts one by one, finally getting the repo back into a normal state once they are all resolved and the merge is committed.

This PR moves us in that direction, but there is much more to do.

Changes in this PR (some functionality added, some removed):
- JSON/text output for `sno merge` is now supported, and gives the result of the merge operation, including a summary of conflicts. 
(Getting the complete list of conflicts in text or JSON will need a follow up change - this could even be a separate command, since querying the full list of remaining conflicts is a separate operation to beginning the merge).
- Seeing or resolving individual conflicts is no longer supported - that will come back in a follow up change, with a separate command eg `sno resolve <conflict_id> <conflict_resolution>`
- Completing a merge in a single go by resolving every conflict interactively is no longer supported - this might come back eventually, perhaps built on top of `sno resolve`.

### DIFF
--- a/sno/conflicts.py
+++ b/sno/conflicts.py
@@ -2,15 +2,12 @@ from collections import namedtuple
 import json
 import logging
 import re
-import sys
 
 import click
 import pygit2
 
-from .diff_output import repr_row
-from .exceptions import InvalidOperation, NotYetImplemented, MERGE_CONFLICT
-from .structs import AncestorOursTheirs, CommitWithReference
-from .structure import RepositoryStructure
+from .exceptions import NotYetImplemented
+from .structs import AncestorOursTheirs
 
 
 L = logging.getLogger("sno.conflicts")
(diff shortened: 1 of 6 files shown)


sno/merge.py
```diff
@@ -35,11 +43,31 @@
...
+    merge_jdict = do_merge(repo, ff, ff_only, dry_run, commit)
+    no_op = merge_jdict.get("no-op", False) or merge_jdict.get("dryRun", False)
+
+    if not no_op:
+        # Update working copy.
+        repo_structure = RepositoryStructure(repo)
+        wc = repo_structure.working_copy
+        L.debug(f"Updating {wc.path} ...")
+        merge_commit = repo[merge_jdict["mergeCommit"]]
+        wc.reset(merge_commit, repo_structure)
```
craigds, line 60: Can you merge without a working copy? Should be possible, right? If so this needs to either handle it or raise NotYetImplemented or something

sno/merge.py
```diff
@@ -48,18 +76,29 @@ def merge(ctx, ff, ff_only, dry_run, commit):
...
+        "message": merge_message,
+        "conflicts": None,
+    }
+
     # We're up-to-date if we're trying to merge our own common ancestor.
     if ancestor_id == theirs.id:
-        click.echo("Already merged!")
-        return
+        merge_jdict["mergeCommit"] = ours.id.hex
+        merge_jdict["no-op"] = True
```
craigds, line 96: I feel weird about this hyphen when other things are camelCase

sno/conflicts.py
```diff
@@ -162,211 +159,74 @@ def _ensure_conflict(self, conflict):
...
-    elif input_mode == InputMode.NO_INPUT:
-        click.echo(
-            "Printing conflicts but not resolving - run from an interactive terminal to resolve"
-        )
+        tables3 = aot(p.split("/.sno-table/", 2)[0] if p else None for p in paths3)
+        actual_tables = [t for t in tables3 if t]
+        all_same_table = all(a == actual_tables[0] for a in actual_tables)
+        if not all_same_table:
+            # This is a really bad conflict - it seems to involve multiple tables.
+            conflicts.setdefault("<other>", 0)
```
craigds, line 180: Any idea what these are? Could you describe in more detail

sno/conflicts.py
```diff
@@ -162,211 +159,74 @@ def _ensure_conflict(self, conflict):
...
-        click.echo("Printing conflicts but not resolving due to --dry-run")
-    elif input_mode == InputMode.INTERACTIVE:
-        interactive_pause(
-            "Press enter to begin resolving merge conflicts, or Ctrl+C to abort at any time..."
-        )
-    elif input_mode == InputMode.NO_INPUT:
-        click.echo(
-            "Printing conflicts but not resolving - run from an interactive terminal to resolve"
-        )
+        tables3 = aot(p.split("/.sno-table/", 2)[0] if p else None for p in paths3)
```
craigds, line 175: could you comment what these paths look like?
should `p.split("/.sno-table/", 2)` be `p.split("/.sno-table/", 1)` ?

tests/test_conflicts.py
```diff
@@ -44,94 +51,40 @@ def test_merge_conflicts(data, data_working_copy, geopackage, cli_runner, update
...
-            in r.stderr
-        )
+        if output_format == "text":
+            assert r.stdout.split("\n") == [
+                'Merging branch "alternate" into master',
+                "Conflicts found:",
+                "",
+                f"{data.LAYER}:",
+                "  Feature conflicts:",
+                "    edit/edit: 3",
```
craigds, line 68: Feels like this should list what the conflicts actually are? rather than just counting them by type

sno/merge.py
olsen232: no-op -> noOp

sno/merge.py
olsen232: For now, just fixed this so it won't crash if you have no working copy.

Potentially needs more work later: a user who is halfway through a merge with lots of conflicts to resolve should probably not be messing with their working copy, since this is a good way to a) get very confused and b) lose updates. (Just like it isn't a good idea to be editing files in your editor at the same time as you update them on disk using a different program).

So, we may want to hide or lock the working copy for the duration of a merge. But this is not done in this PR

sno/conflicts.py
olsen232: Will add a comment. But basically it means features moved from one table to another, or tables were renamed. And it probably means you will not be resolving this conflict anytime soon.

sno/conflicts.py
olsen232: Done

tests/test_conflicts.py
olsen232: Maybe - inclined to do this in a follow up PR though.
Certainly we need to have the ability to display all of the conflicts, otherwise the user can't resolve them. And, we likely need a separate command to show conflicts, because the user if the user returns to half-finished merge, they will want to see them again. So, if we do want to display all resulting conflicts in full in the `sno merge` output, it makes sense to do it by delegating to that other (as yet unimplemented ) command.

Counting them by type is still useful because they will probably still need to be sorted by type. Eventually (when we support this) the user will be required to resolve schema conflicts for a particular table before resolving feature conflicts. Until the resolved schema is known, it is not clear how the features look, or if they are even still conflicts which need resolving.

sno/merge.py
craigds: ok 👍 

It's a good way to get confused yes. I think _removing_ the WC would be even more confusing. Locking it would be reasonable, might work for GPKG (not sure) but isn't going to work for future WC types (postgres)

WWGD? I think git would probably do the commit but then refuse to update the working copy afterwards. Then you'd need to type some special git reset command to allow it to reset to the new commit.

tests/test_conflicts.py
craigds: > in a follow up PR
ok

> we likely need a separate command to show conflicts, because the user if the user returns to half-finished merge, they will want to see them again

IMHO, `sno status` should do that, if you're halfway through a merge

sno/merge.py
craigds: ah - it's different for git since the WC is meaningful during a merge, since it's used for resolving conflicts. Even though you have to add the resovled conflicts to the staging area, the WC is otherwise at the same state as the resolved merge. That isn't the case in sno

rcoup: 1. removing the WC isn't an option.
2. but we're not pushing conflicts through the WC though, right? So can't we just ignore it and basically refuse to look at it until the merge is completed or aborted? (similarly I've seen git give some "you're in the middle of merging, I can't do that!" errors)

PR #80, opened 2020-05-12
Implement "move_repo_to_merging_state". Resolving conflicts not implemented

olsen232:
Running "sno merge <other_branch> will merge other branch, unless there are conflicts.
In which case, it will put the repository into a merging state.
```
~/mysnorepo (BARE:mybranch) $ sno merge other
~/mysnorepo (BARE:mybranch|MERGING) $
```
Unfortunately, since conflicts cannot be resolved, the only option is to abandon the merging state.
```
~/mysnorepo (BARE:mybranch|MERGING) $ sno merge --abort
~/mysnorepo (BARE:mybranch) $
```
However, the merging state is real, and contains (most of?) the information we will need to make step-by-step resolution of conflicts possible. Resolving conflicts is to be done in a follow up PR.

Also still to be done is a way to display the merging state including conflicts / summary of conflicts. Running `sno status` and `sno show` right now doesn't mention that you are in merging state. This is to be done in a follow up PR.
Other commands may also have to be deactivated, since they make no sense in the context of a merge.

Some tidying:
- Added a new file sno/repo_files.py to make it easier to read and write files inside the bare repository. Modified commit.py to use this functionality.
- Changed the JSON output of merge.py to focus on commits, not commits + branches. Commits are what is actually merged, and the branch head may have moved on by the the merge is complete.

### DIFF
--- a/sno/repo_files.py
+++ b/sno/repo_files.py
@@ -1,0 +1,91 @@
+import os
+
+from pathlib import Path
+import shlex
+import shutil
+import subprocess
+
+
+from . import is_windows
+from .exceptions import SubprocessError
+
+
+HEAD = "HEAD"
+COMMIT_EDITMSG = "COMMIT_EDITMSG"
+ORIG_HEAD = "ORIG_HEAD"
+MERGE_HEAD = "MERGE_HEAD"
+MERGE_MSG = "MERGE_MSG"
+MERGE_INDEX = "MERGE_INDEX"
+MERGE_LABELS = "MERGE_LABELS"
+
+
+def repo_file_path(repo, filename):
+    return Path(repo.path) / filename
+
+
+def repo_file_exists(repo, filename):
+    return repo_file_path(repo, filename).exists()
+
+
+def write_repo_file(repo, filename, contents):
+    if not isinstance(contents, str):
+        raise TypeError("File contents must be a string", type(contents))
+    if not contents.endswith("\n"):
+        contents += "\n"
+
+    path = repo_file_path(repo, filename)
+    path.write_text(contents, encoding="utf-8")
+
+
+def fallback_editor():
+    if is_windows:
+        return "notepad.exe"
+    else:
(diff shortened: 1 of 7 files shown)


sno/repo_files.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def repo_file_exists(repo, filename):
+    return repo_file_path(repo, filename).exists()
+
+
+def write_repo_file(repo, filename, contents):
+    if not isinstance(contents, str):
+        raise TypeError("File contents must be a string", type(contents))
+    if contents[-1] != "\n":
```
craigds, line 32: Could cause IndexError for empty strings. Use `.endswith()`

sno/repo_files.py
```diff
@@ -0,0 +1,87 @@
...
+        contents += "\n"
+
+    path = repo_file_path(repo, filename)
+    path.write_text(contents, encoding="utf-8")
+
+
+if is_windows:
+    FALLBACK_EDITOR = "notepad.exe"
+else:
+    FALLBACK_EDITOR = "nano"
```
craigds, line 42:
```python
    import shutil
    FALLBACK_EDITOR = shutil.which("nano") or "vi"
```

I'd use nano if possible but not all linux distros have it. `vi` is part of POSIX so we should probably use that as a last-ditch fallback if `nano` doesn't exist?

Hopefully users can figure out how to exit it...

rcoup: `nano` is included with default Debian, Centos, Fedora, and Ubuntu.

sno/repo_files.py
olsen232: Done

sno/repo_files.py
olsen232: Done - rcoup@ says everyone has nano, but since it is easy to workaround if they don't, fallback-of-fallback to vi is done.

PR #82, opened 2020-05-13
In non-CI, skip benchmark tests by default

craigds:
Use `--run-benchmarks` to run them.

This avoids multiple screens worth of output at the end of test runs (even in non-verbose mode), which can be annoying for debugging other tests.

All tests still run in CI.

This also adds a bold+yellow message at the start of test runs when it skips tests:
Screen Shot 2020-05-14 at 9 19 27 AM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,8 @@ ci-test:
 		-p no:sugar \
 		--cov-report term \
 		--cov-report html:test-results/coverage/ \
-		--junit-xml=test-results/junit.xml
+		--junit-xml=test-results/junit.xml \
+		--benchmark-enable
 
 # Cleanup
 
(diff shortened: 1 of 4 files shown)


rcoup:
Fine in principle.

1. these are already skipped with `-m not slow` which is done via a mark, could we keep the approach consistent somehow (doesn't really matter how/which)?
2. If there are non-benchmarking does-it-work aspects of the benchmark tests then I think they need to be pulled out into separate tests.

craigds:
I was poking around the config object looking for a way to check the marker expressions passed on the CLI, and realised pytest-benchmark actually does have its own way to disable benchmarks (`--benchmark-disable`.) So,  I've replaced this PR with just using that by default. It can be overridden with `--benchmark-enable`.

That addresses your concerns I think - tests now always run, but just don't actually do any benchmarking.

craigds:
CI failed but only because of a strange windows build failure, which also occurs on master (https://github.com/koordinates/sno/actions/runs/105153414 ) but passed on master yesterday. So that's not a blocker for merging.

PR #83, opened 2020-05-14
Imports from generic OGR datasets

craigds:
Implements a generic importer using OGR, so that we can implement importers for any OGR-supported format easily. This applies to both `sno import` and `sno init --import` commands.

For now I have whitelisted the accepted formats to TAB/SHP/GPKG. We may wish to test & benchmark other formats before adding them, or reconsider how they are specified on the command line. Note that we have a very restricted GDAL, so the set of possible formats is very limited at present anyway.

The new `--list-formats` option shows importable formats:

```
$ sno import --list-formats
GPKG
SHP
TAB
```

# Implementation details

The GPKG importer doesn't actually _use_ OGR for its feature iterators; I tried doing so but noticed a sizeable drop in performance when running the `test_import` benchmarks (50-100% slowdown depending on dataset, and this persisted when loading a large 1.0GB geopackage also.) Overriding the feature iterators to use the APSW cursors restored original benchmark scores.

# Usage changes

In preparation for handling non-file datasets (such as postgres), the old specifier of `GPKG:/path/to.gpkg:tablename` has been altered a bit:

## prefix (`GPKG:`) is now optional.

If not specified, we let OGR figure out what format the given path is, and then we instantiate our own importer subclass based on OGR's decision. If it _is_ specified, we only allow OGR to use the specified driver to load the dataset.

## `:{tablename}` replaced with `--table`

I thought it would be best to drop the `:tablename` part of the source spec for future extensibility, since `:` is a valid character in filenames and URLs (particularly schemes, but it's valid in the 'path' part too)

### The new behaviour is:

- If only one table is present in the source, that table is automatically chosen. That's always the case for SHP/TAB sources.
- Otherwise, if you gave `--table={name}` on the command line, that table will be chosen.
- Otherwise, it will prompt to select a table interactively
- Unless sno's not being run interactively, in which case it will error with a message.

### Future compatibility concerns

We've identified that probably in the near future, we'll want to import:

1. multiple tables (potentially >10) at once from a single source
2. all the tables from a source
3. multiple sources (at once? not sure)

We could soon allow multiple `--table` options to easily address (1). (2) could be addressed via some future option (`--all-tables` ?). (3) might not be possible with the current argument layout, but I guess the user could always run `sno import` multiple times.

@rcoup asked elsewhere why I didn't choose to do tables as just a list of positional arguments. My reasoning is:

Both init and import commands currently take an _optional_ positional argument (called `DIRECTORY`, although it's for different purposes in both).

```
Usage: sno init [OPTIONS] [DIRECTORY]
Usage: sno import [OPTIONS] SOURCE [DIRECTORY]
```

This means that simply specifying a list of multiple tables as separate trailing args would be ambiguous; maybe you meant a table or maybe you meant a directory. Hence the `--table` is required.

#### Alternatives 

Of course, there is at least one alternative to using `--table`. Both usages of DIRECTORY are probably _somewhat_ rare; we could make those named options instead:

```
# init a repo in DIRECTORY, and import two tables from ~/my.gpkg
sno init --repo-path=DIRECTORY --import ~/my.gpkg table1 table2

# import two tables from ~/my.gpkg and rename one of them
sno import ~/my.gpkg table1:renamed_table1 table2
```

Re the latter usage, my experience is that colons are rare enough in table names that this is pretty unlikely to trip up users. `@` might be another sane separator choice, especially because 'at' already sounds like a rename. Neither `:` nor `@` requires shell quoting.


rcoup:
Can we split out the CLI semantics to a separate issue?

sno/init.py
```diff
@@ -43,200 +57,368 @@ def import_gpkg(ctx, geopackage, table, list_tables):
...
-        prefix, value = value.split(":", 1)
+class OgrImporter:
+    """
+    Imports from an OGR source, currently from a whitelist of formats.
+    """
 
-        # need to deal with "GPKG:D:\foo\bar.gpkg:table"
-        search_from = 2 if is_windows else 0
+    @classmethod
+    def adapt_source_for_ogr(cls, source):
```
rcoup, line 78: suggest we add `OGR:{whatever}` which is passed directly without whitelisting?

sno/init.py
```diff
@@ -43,200 +57,368 @@ def import_gpkg(ctx, geopackage, table, list_tables):
...
 
-        prefix = prefix.upper()
-        if prefix not in self.prefixes:
-            self.fail(
-                f'invalid prefix: "{prefix}" (choose from {", ".join(self.prefixes)})',
-                exit_code=INVALID_ARGUMENT,
+        writable = True
+        try:
+            # Try to open writable
+            ds = gdal.OpenEx(
```
rcoup, line 116: Why do we ever want a writeable import source? Seems... bad.

sno/init.py
```diff
@@ -43,200 +57,368 @@ def import_gpkg(ctx, geopackage, table, list_tables):
...
-                )
+        self.source = source
+        self.ogr_source = ogr_source
+        self.writable = writable
 
-    def get_table_list(self):
-        db = gpkg.db(self.source)
-        dbcur = db.cursor()
+    @property
+    @functools.lru_cache(maxsize=1)
```
rcoup, line 176: is this safe to do wrt GDAL's GC handling?

sno/init.py
```diff
@@ -43,200 +57,368 @@ def import_gpkg(ctx, geopackage, table, list_tables):
...
+            self.ds.RollbackTransaction()
 
     @property
     @functools.lru_cache(maxsize=1)
     def row_count(self):
-        dbcur = self.db.cursor()
-        dbcur.execute(f"SELECT COUNT(*) FROM {gpkg.ident(self.table)};")
-        return dbcur.fetchone()[0]
+        # note: choose FAST method if possible ( i recall them being different with MITAB especially)
+        return self.ogrlayer.GetFeatureCount()
```
rcoup, line 262: API is documented as `GetFeatureCount(Layer self, int force=1) -> GIntBig`, so presumably we need `force=0`? Or there's the `OLCFastFeatureCount` capability

sno/init.py
```diff
@@ -43,200 +57,368 @@ def import_gpkg(ctx, geopackage, table, list_tables):
...
 
     @property
     @functools.lru_cache(maxsize=1)
     def primary_key(self):
-        return gpkg.pk(self.db, self.table)
+        return self.ogrlayer.GetFIDColumn()
```
rcoup, line 267: FID always has to be a bigint in OGR AFAIK — it's not the same as a primary key.

sno/init.py
```diff
@@ -43,200 +57,368 @@ def import_gpkg(ctx, geopackage, table, list_tables):
...
+                l.ResetReading()
+                return
+            # Turn an OGRFeature into a name:value dict
+            yield f
+
+    def iter_features(self):
+        for ogr_feature in self._iter_ogr_features():
+            yield self._ogr_feature_to_dict(ogr_feature)
+
+    def iter_features_sorted(self, pk_callback, limit=None):
```
rcoup, line 344: what's this actually used for? (I vaguely remember it being related to deterministic imports when we were generating IDs per feature, but we don't do that anymore)

sno/init.py
```diff
@@ -271,52 +453,83 @@ def iter_features_sorted(self, pk_callback, limit=None):
...
+@click.argument("source")
 @click.argument(
     "directory",
     type=click.Path(file_okay=False, exists=False, resolve_path=True),
     required=False,
 )
+@click.option(
+    "--table",
+    "-t",
+    help="Which table to import. If not specified, this will be selected interactively",
```
rcoup, line 514: renaming of tables needs to come back

sno/ogr_util.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def adapt_value_noop(value):
+    return value
+
+
+def adapt_ogr_datetime(value):
+    # OGR uses this strange format: '2012/07/09 09:01:52+00'
+    # We convert back to a normal ISO8601 format.
+    return value.replace('/', '-').replace(' ', 'T').replace('+00', 'Z')
```
rcoup, line 11: don't we want to adapt it to a Python datetime?

sno/utils.py
```diff
@@ -0,0 +1,38 @@
+import functools
+from itertools import islice
+
+
+def ungenerator(cast_function):
```
rcoup, line 5: isn't this just `dict(mygenerator())` / `list(mygenerator())`?

tests/test_structure.py
```diff
@@ -89,6 +91,53 @@ def _import_check(repo_path, table, source_gpkg, geopackage):
     return dataset
 
 
+def _ogr2ogr(*, src_path, dst_path, dst_driver_name, layer_name):
+    """
+    Converts a datasource from one format to another.
+    Only converts the given named layer.
+
+    We don't seem to have the OGR binaries in our build of GDAL, so here's a
```
rcoup, line 99: Use `gdal.VectorTranslate()`

sno/init.py
craigds: Performance. It means we can use temp tables, in the case of postgres/GPKG (well, GPKG is overridden anyway, but the `iter_features_sorted` implementation uses temp tables)

sno/init.py
craigds: hm. that might have implications. What would you do here instead then? The GPKG driver exposes the PK of the ogr layer as the FID column.

Other drivers we're dealing with right now (SHP/TAB) don't have a concept of PK at all, so this always returns "FID"

sno/init.py
craigds: good question. I have no idea. The only way to use it from the CLI is via:

```
SNO_IMPORT_OPTIONS='{"iter_func": 2}'
```
Which is confusing. But we test that behaviour, so I couldn't omit it without removing the test. Do you reckon it can be removed?

sno/init.py
craigds: _could_ do... Is the goal to 'unofficially' support things other than what's been tested?

sno/init.py
craigds: This just means the layer won't be collected until the OGRImporter is collected (which is at the end of the `import_table` method, just before the process ends). Seems probably-okay to me?

small aside: GDAL's GC handling has always confused me. 
* Why do all the python usage examples do `del ds` at the end of the method, just before it goes out of scope (and thus would be implicitly deleted) _anyway_. (I remove those del statements and have never had any trouble)
* I've noticed before that GDAL's objects get deleted _immediately_ upon going out of scope. They don't hang around waiting for a future GC run. Not quite sure how they implemented this, or what its implications are

sno/utils.py
craigds: yes, but it's a decorator that applies that to a generator-returning function

tests/test_structure.py
craigds: 👍 i swear i looked for that everywhere

sno/init.py
craigds: Only used it for `iter_features_sorted`, which I've now removed (#87 ) so I've removed this now.

sno/init.py
craigds: it is still possible, and hasn't changed. 

`sno import --table=orig_table renamed_table/`

craigds: It only needs to change if we need to remove the `--table` option and use unnamed arguments for table names. In which case it would have to be something like

```
sno import path/to.gpkg orig_table:renamed_table [orig_table2:renamed_table2]
```

sno/ogr_util.py
craigds: I don't think so - the output of this ends up being passed directly to `msgpack.packb` (in [import_iter_feature_blobs](https://github.com/koordinates/sno/blob/ogr-import/sno/structure.py#L953-L954)), which can't handle datetimes

craigds:
> Can we split out the CLI semantics to a separate issue?

This PR began as an effort to support postgres imports, but it evolved to not quite be able to do that yet (our GDAL doesn't yet have a postgres driver)

The change from `GPKG:/path:tablename` to `GPKG:/path --table=tablename` was driven by the need to support postgres (via URL, probably) in future.

Given that this PR no longer deals with postgres URLs, I _could_ split all those changes into a separate PR, but am reluctant to do so because

a. it's a pile of work to do so
b. I'll just have to add it all back in again in the next PR

sno/init.py
rcoup: "they" didn't implement it... IIUC Python's GC deallocates objects immediately when the reference count drops to 0 and the generational GC is only for cleaning up cycles.

sno/init.py
rcoup: [Requirement 150](http://www.geopackage.org/spec/#r150):

> A feature table or view SHALL have a column that uniquely identifies the row. For a feature table, the column SHOULD be an integer primary key. If there is no primary key column, the first column SHALL be of type INTEGER and SHALL contain unique values for each row.

Try a GPKG with a non-integer PK? I suspect GDAL might use the Sqlite ROWID as the FID.

sno/ogr_util.py
rcoup: Should probably have it as a unix timestamp in that case? Strings are yuck.

Add to the data-structure-v3 doc?

rcoup:
> As a way forward, I could reinterpret your question as "Can we get this PR merge as-is and then deal with the --table semantics separately?"
>
> In which case, yes we should do that 😄

That's what I meant

craigds:
cheers. So the last remaining issue here is the cause of the test failures - I need to preserve field _length_ in the `sqlite_table_info` metadata from OGRImporter. If I can get that passing I'll merge this.

sno/utils.py
craigds: I got britta'd.

Are you saying this isn't useful? Or could be implemented simpler?

It's just a nice sugar to avoid having to build up a dict/list in the function and then return it. The decorator can aggregate rather than the function code having to.

sno/init.py
rcoup: Basically yes

sno/utils.py
rcoup: I don’t really get it, but you’re very capable so it must be useful 😀👍

sno/init.py
craigds: ok. that makes sense. I'm pretty sure this caching is fine 👍

sno/ogr_util.py
craigds: Personally I don't have a problem with strings, doesn't seem worth changing.

If we change it to some other format, we need to convert it when checking out a working copy - GPKG mandates the storage of datetimes as ISO8601 strings:  http://www.geopackage.org/spec/#table_column_data_types

sno/init.py
craigds: done in e2fe09acbfc727938bc12a087a6bb9eebfb9a5de

sno/init.py
craigds: You're right. with a string PK:

```python
>>> ogrlayer.GetFIDColumn()
''

>>> field.IsNullable()
1
```

I can't see any other way to get a PK field via OGR; this might have to be overridden on a per-driver basis

craigds: Fixed in 4950534ee936ecb65362b4a2d005e4ed6e0bdd52

PR #90, opened 2020-05-26
Import from postgres databases

craigds:
This adds support for importing from Postgres databases using a connection string:

```
$ sno init --import 'postgresql://docker:docker@localhost:5432/gis' --table TABLE
```

## To do:

* [x] Enable running the postgres import test in CI
* [x] vendor-add-pq needs to be merged before this
* [x] postgres test needs to run/pass in linux in CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,11 @@ install: | $(sno-app-any)
 test: $(py-install-test)
 	pytest -v --cov-report term --cov-report html:coverage
 
+ifeq ($(PLATFORM),Linux)
+# (github actions only supports docker containers on linux)
+ci-test: export SNO_POSTGRES_URL ?= postgresql://docker:docker@postgis:5432/gis
+endif
+
 .PHONY: ci-test
 ci-test:
 	pytest \
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ test: $(py-install-test)
 
 ifeq ($(PLATFORM),Linux)
 # (github actions only supports docker containers on linux)
-ci-test: export SNO_POSTGRES_URL ?= postgresql://docker:docker@postgis:5432/gis
+ci-test: export SNO_POSTGRES_URL ?= postgresql://docker:docker@localhost:5432/gis
 endif
 
 .PHONY: ci-test
--- a/sno/init.py
+++ b/sno/init.py
@@ -111,6 +111,14 @@ class OgrImporter:
         return ogr_source, allowed_formats
 
     @classmethod
+    def _ogr_open(cls, ogr_source, **open_kwargs):
+        return gdal.OpenEx(
+            ogr_source,
+            gdal.OF_VECTOR | gdal.OF_VERBOSE_ERROR | gdal.OF_READONLY,
+            **open_kwargs,
+        )
+
+    @classmethod
     def open(cls, source, table=None):
         ogr_source, allowed_formats = cls.adapt_source_for_ogr(source)
         if allowed_formats is None:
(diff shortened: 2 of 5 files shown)


sno/init.py
```diff
@@ -503,6 +520,82 @@ def build_meta_info(self):
...
+                SELECT a.attname
+                FROM   pg_index i
+                JOIN   pg_attribute a ON a.attrelid = i.indrelid
+                                     AND a.attnum = ANY(i.indkey)
+                WHERE  i.indrelid = %s::regclass
+                AND    i.indisprimary;
+                """,
+                [self.table],
+            )
+            return cur.fetchone()[0]
```
rcoup, line 596: if there's a multi-column PK? At a minimum should error out, not silently use one of them. That'd be bad.

sno/init.py
```diff
@@ -503,6 +520,82 @@ def build_meta_info(self):
...
+        Takes a URL ('postgresql://..')
+        and turns it into a key/value connection string, prefixed by 'PG:' for OGR.
+
+        libpq actually handles URIs fine, but OGR doesn't :(
+        So to import via OGR we have to convert them.
+        https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
+
+        We don't currently handle all of the things these URLs can contain.
+        In particular, we don't (currently) handle:
+         * domain sockets
```
rcoup, line 536: Is that because OGR doesn't handle them? They're the default on POSIX for local DBs, would be handy to support if it's straightforward.

Makefile
```diff
@@ -9,6 +9,12 @@ else
 	PLATFORM := $(shell uname -s)
 endif
 
+ifeq ($(PLATFORM),Linux)
+	# used by ci-test
+	# (github actions only supports docker containers on linux)
+	SNO_POSTGRES_URL='postgresql://docker:docker@postgis:5432/gis'
```
rcoup, line 15: can you add this down at `ci-test` then? And can make it a target-specific variable if it's only used there.

eg:
```
ifeq ($(PLATFORM),Linux)
ci-test: export SNO_POSTGRES_URL ?= postgresql://docker:docker@postgis:5432/gis
endif

ci-test: 
    ...
```

craigds: wow, Makefile is a _weird_ language. Done in latest

sno/init.py
craigds: OGR handles them. 

It's mostly because I didn't want to take the time to understand the syntactical nuances, but I'll take a look at it tomorrow. I don't know if OGR supports them or not, I assume it does

rcoup: https://stackoverflow.com/a/29356068/2662 has syntax details (emphasis mine)

> The host component is interpreted as described for the parameter host. In particular, **a Unix-domain socket connection is chosen if the host part is either empty or starts with a slash**, otherwise a TCP/IP connection is initiated. Note, however, that the slash is a reserved character in the hierarchical part of the URI. So, to specify a non-standard Unix-domain socket directory, either omit the host specification in the URI and specify the host as a parameter, or percent-encode the path in the host component of the URI:
>
>     postgresql:///dbname?host=/var/lib/postgresql
>     postgresql://%2Fvar%2Flib%2Fpostgresql/dbname

craigds: added support for those

sno/init.py
```diff
@@ -503,6 +520,84 @@ def build_meta_info(self):
...
+
+        url = urlsplit(url)
+        scheme = url.scheme.lower()
+        if scheme not in ('postgres', 'postgresql'):
+            raise ValueError("Bad scheme")
+
+        params = dict(parse_qsl(url.query))
+
+        if url.username:
+            params.setdefault('user', url.username)
```
olsen232, line 547: Looks like "user" could be specified in two different places, just like the other fields, and one has precedence. Add a comment or add something about it to the docstring.

tests/test_structure.py
```diff
@@ -561,6 +566,152 @@ def test_shp_import_meta(
         ]
 
 
+def quote_ident(part):
+    """
+    SQL92 conformant identifier quoting, for use with OGR-dialect SQL
+    (and most other dialects)
+    """
+    part = part.replace('"', '""')
+    return '"%s"' % part
```
olsen232, line 575: python3 means f-strings, which we use most everywhere

rcoup:
@craigds I changed the base on this to master now `vendor-add-pq` was merged

PR #94, opened 2020-05-28
Support `git merge --continue`

olsen232:
`git merge --continue`
- only works when in a state "merging"
- only works when all conflicts are resolved
- writes all the merged files plus the resolved conflicts as a new commit, moves branch head to the new commit

### DIFF
--- a/sno/cli_util.py
+++ b/sno/cli_util.py
@@ -53,3 +53,26 @@ def do_json_option(func):
         default=False,
         help="Whether to format the out output as JSON instead the default text output.",
     )(func)
+
+
+def call_and_exit_flag(*args, callback, **kwargs):
+    """
+    Add an is_flag option that, when set, eagerly calls the given callback with only the context as a parameter.
+    The callback may want to exit the program once it has completed, using ctx.exit(0)
+    Usage:
+    @call_and_exit_flag("--version", callback=print_version, help="Print the version number")
+    """
+
+    def actual_callback(ctx, param, value):
+        if value and not ctx.resilient_parsing:
+            callback(ctx)
+            ctx.exit()
+
+    return click.option(
+        *args,
+        is_flag=True,
+        callback=actual_callback,
+        expose_value=False,
+        is_eager=True,
+        **kwargs,
+    )
--- a/sno/merge.py
+++ b/sno/merge.py
@@ -3,7 +3,7 @@ import sys
 
 import click
 
-from .cli_util import do_json_option
+from .cli_util import do_json_option, call_and_exit_flag
 from .conflicts import (
     list_conflicts,
     conflicts_json_as_text,
@@ -12,14 +12,11 @@ from .exceptions import InvalidOperation
 from .merge_util import AncestorOursTheirs, MergeIndex, MergeContext
 from .output_util import dump_json_output
 from .repo_files import (
-    ORIG_HEAD,
     MERGE_HEAD,
     MERGE_MSG,
-    MERGE_INDEX,
-    MERGE_LABELS,
-    repo_file_path,
     write_repo_file,
-    remove_repo_file,
+    read_repo_file,
+    remove_all_merge_repo_files,
     is_ongoing_merge,
     repo_file_exists,
 )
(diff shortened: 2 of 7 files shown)


sno/merge.py
```diff
@@ -207,24 +135,76 @@ def move_repo_to_merging_state(repo, merge_index, merge_context, merge_message):
...
+        raise InvalidOperation('Repository is not in "merging" state.')
+    merge_index = MergeIndex.read_from_repo(repo)
+    if merge_index.unresolved_conflicts:
+        raise InvalidOperation(
+            "Merge cannot be completed until conflicts are resolved."
+        )
+
+    merge_context = MergeContext.read_from_repo(repo)
+    commit_ids = merge_context.versions.map(lambda v: v.repo_structure.id)
+    merge_message = read_repo_file(repo, MERGE_MSG)
```
rcoup, line 172: User doesn't get prompted to edit this like a normal commit?

sno/merge.py
```diff
@@ -267,3 +247,69 @@ def output_merge_json_as_text(jdict):
...
+)
+@eager_callback_flag(
+    '--continue',
+    callback=complete_merging_state,
+    help="Completes and commits a merge once all conflicts are resolved and leaves the merging state",
+)
+@click.argument("commit", required=True, metavar="COMMIT")
+@do_json_option
+@click.pass_context
+def merge(ctx, ff, ff_only, dry_run, do_json, commit):
```
rcoup, line 288: Should `merge` be a [Click group](https://click.palletsprojects.com/en/7.x/commands/) (like `sno` is)? 

(So `merge continue`/`merge abort`/`merge whatever` can have separate options eg: `--message` without needing complex argument-parsing machinations?)

Maybe that's later changes as it gets more complicated.

craigds: IMO no:
* we don't need any continue/abort options (at the moment?)
* it'd make it hard/confusing/ambiguous to merge a branch called `continue` or `abort`, which is bound to come up sometimes
* without a good reason to change it, we should stick with the git usage: `git merge --continue`  / `git merge --abort`

sno/cli_util.py
```diff
@@ -53,3 +53,25 @@ def do_json_option(func):
         default=False,
         help="Whether to format the out output as JSON instead the default text output.",
     )(func)
+
+
+def eager_callback_flag(*args, callback, **kwargs):
```
craigds, line 58: There are a couple of other existing eager flags (in sno/cli.py and sno/init.py) , should make them use this also

craigds: I'd probably take this one step further and make it call `ctx.exit(0)` so the callbacks don't have to. 

And call it something like `call_and_exit`:

```python
@call_and_exit("--version", callback=print_version, help="Print the version number") 
```

Since all our usages of it do use it like that already...

sno/merge_util.py
```diff
@@ -196,13 +196,15 @@ def remove_conflict(self, key):
...
             if not entry:
                 continue
             result_path = f".conflicts/{key}/{version}/{entry.path}"
-            result.append(cls.Entry(result_path, entry.id, entry.mode))
-        return result
+            yield cls.Entry(result_path, entry.id, entry.mode)
+
+    def _serialise_conflicts(self):
+        for key, conflict3 in self.conflicts.items():
+            yield from self._serialise_conflict(key, conflict3)
```
craigds, line 207: 💯 you have seen the light

sno/merge_util.py
```diff
@@ -221,10 +223,9 @@ def _deserialise_conflict_part(cls, index_entry):
         result = AncestorOursTheirs.partial(**{version: result_entry})
         return key, result
 
-    def conflicts_as_entries(self):
-        for key, conflict3 in self.conflicts.items():
-            for index_entry in self._serialise_conflict(key, conflict3):
-                yield index_entry
+    def _conflicts_paths(self):
+        """All the paths in all the entries in all the conflicts, as a set."""
+        return set(e.path for c in self.conflicts.values() for e in c if e)
```
craigds, line 228: 👎 double comprehensions are really confusing, can you split this into a proper for loop?

```python
from .utils import ungenerator

@ungenerator(set)
def _conflicts_paths(self):
    """All the paths in all the entries in all the conflicts, as a set."""
    for c in self.conflicts.values():
        for e in c:
            if e:
                yield e.path
```

sno/merge_util.py
```diff
@@ -275,10 +277,9 @@ def _deserialise_resolve_part(cls, index_entry):
         result = MergedOursTheirs.partial(**{version: result_entry})
         return key, result
 
-    def resolves_as_entries(self):
-        for key, resolve3 in self.resolves.items():
-            for index_entry in self._serialise_resolve(key, resolve3):
-                yield index_entry
+    def _resolves_entries(self):
+        """All the entries in all the resolves."""
+        return (e for r in self.resolves.values() for e in r if e)
```
craigds, line 282: same here re double comprehensions

sno/merge.py
olsen232: ... I guess they should? I copied the existing behaviour from sno merge (no conflicts). I'll fix both in a follow up PR

sno/merge.py
olsen232: Leaving as-is for now

sno/cli_util.py
olsen232: Done

olsen232: Done

sno/merge_util.py
olsen232: Done

sno/merge_util.py
olsen232: Done

PR #96, opened 2020-05-29
Add an output pager.

### LINKED ISSUE #50: Add output pager to verbose commands
This came up as part of #48. Some commands generate a lot of output, and mostly we don't pipe them to a pager at the moment.

Some of our commands exec git directly, so those may use a pager already, but over time more of the commands will probably be directly implemented in sno rather than execcing git.

We should add an output pager (presumably usually `less`) and ways of customising it.

craigds:
Fixes #50

What a mission. This took _way_ longer than I expected. 

But I'm fairly happy with the results:

Screen Shot 2020-05-29 at 8 51 17 PM

# Notes:

~I changed `resolve_output_path` to a contextmanager, and made it yield a file-like object that pushes lines of text into a queue, which are read by a worker thread which yields them via `click.echo_via_pager`.~

I implemented pallets/click#1572 to support doing this without having to use threads... that should ideally be merged and released before this PR is merged.

### DIFF
--- a/sno/output_util.py
+++ b/sno/output_util.py
@@ -1,6 +1,14 @@
 import io
 import json
+import os
+import shutil
 import sys
+import threading
+from contextlib import closing, contextmanager
+from queue import Queue, Empty
+
+import click
+from click._compat import should_strip_ansi
 
 JSON_PARAMS = {
     "compact": {},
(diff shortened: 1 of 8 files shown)


sno/output_util.py
```diff
@@ -14,35 +21,48 @@ def dump_json_output(output, output_path, json_style="pretty"):
...
+            dumped = json.dumps(output, **JSON_PARAMS[json_style])
+            highlighted = highlight(dumped.encode(), JsonLexer(), TerminalFormatter())
+            fp.write(highlighted)
+        else:
+            json.dump(output, fp, **JSON_PARAMS[json_style])
 
 
-def resolve_output_path(output_path):
+@contextmanager
+def resolve_output_path(output_path, allow_pager=True):
```
rcoup, line 39: We should probably support [`core.pager`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_pager) from git config, maybe `=sno` and `=false`? I [turn it off in CI E2E tests](https://github.com/koordinates/sno/blob/f25eb993a68b3800fa28f7e3ddd3b4e24b7b84a4/tests/scripts/e2e-1.ps1#L56) since the CI envs have TTY things so output is pretty.

sno/output_util.py
```diff
@@ -69,3 +89,71 @@ def is_empty_stream(stream):
...
+def _push_environment(k, v):
+    orig = os.environ.get(k)
+    _setenv(k, v)
+    try:
+        yield
+    finally:
+        _setenv(k, orig)
+
+
+DEFAULT_PAGER = shutil.which('less')
```
rcoup, line 111: Ah, click doesn't have a pager, it passes it to `less`/etc?

craigds:
In retrospect, this is quite roundabout. If I implement something in click itself I can just yield the filelike object it's already using for stdin to less. i.e. bypass the generator interface altogether. Then I dont need threads :)

I'll work on a click PR

craigds:
WIP here: https://github.com/craigds/click/tree/filelike-pager

sno/output_util.py
craigds: correct. but it does handle the Weird Issues around doing that

rcoup:
> WIP here

Looks reasonable. I'm not sure if Click currently supports it, but does your approach break combining a pager with terminal output because the stream is closed when the context manager exits?

eg: prompt for X, view some text in the pager, quit the pager, prompt for Z, view some more output, ...

craigds:
maybe one day we'll return to this, but it's been almost three years, I'm sure we'll do it differently anyway

PR #97, opened 2020-06-02
Lock commands when repo is in merging state.

olsen232:
As of this PR, almost every command is deactivated when repo is in "merging" state, with a somewhat helpful error message explaining why. Three commands still work:

sno merge
sno conflicts
sno resolve

Still TODO in future PRs:
- Make sno status show useful info when in merging state, and re-enable that command.
- Decide which other commands are harmless enough that a user can do them mid-merge without messing up the merging state. Maybe read-only commands like show, or diff (although by default, diff shows you a diff of the working copy, but the working copy is not actually useful during a merge)
- Make a plan about if the working copy should also be locked or hidden somehow during a merge


sno/merge.py
```diff
@@ -156,18 +157,23 @@ def complete_merging_state(ctx):
...
-    repo = ctx.obj.repo
-    if not is_ongoing_merge(repo):
-        raise InvalidOperation('Repository is not in "merging" state.')
+    repo = ctx.obj.get_repo(
+        allowed_states=[RepoState.MERGING],
+        bad_state_message='`sno merge --continue` only works when sno repo is in "merging" state.',
+    )
     merge_index = MergeIndex.read_from_repo(repo)
     if merge_index.unresolved_conflicts:
         raise InvalidOperation(
```
craigds, line 166: feels like these errors for invalid state transitions should probably have tests. at least for the exit codes, if not the actual messages

sno/merge.py
```diff
@@ -124,30 +125,30 @@ def move_repo_to_merging_state(repo, merge_index, merge_context, merge_message):
...
     is_ongoing_merge = repo_file_exists(repo, MERGE_HEAD)
     # If we are in a merge, we now need to delete all the MERGE_* files.
     # If we are not in a merge, we should clean them up anyway.
     remove_all_merge_repo_files(repo)
+    assert RepoState.get_state(repo) != RepoState.MERGING
 
     if not is_ongoing_merge:
-        raise InvalidOperation('Repository is not in "merging" state.')
+        raise InvalidOperation(
+            '`sno merge --abort` only works when sno repo is in "merging" state.'
```
craigds, line 150: I think I prefer the old message, because if you type `sno merge --abort` you probably already know it doesn't work unless you're in merging state, but you think you _are_ in merging state.  So if it tells that your assumption is incorrect then that's a more helpful message

sno/context.py
```diff
@@ -25,8 +38,18 @@ def repo(self):
         """
         Returns the sno repository at self.repo_path
         Raises an error if there isn't a sno repository there.
+        Ensures that the repository is in state NORMAL, since this is generally required.
+
         Accessing Context.repo.path ensures you have the path to an existing sno repository.
         """
+        return self.get_repo()
+
+    def get_repo(self, allowed_states=[RepoState.NORMAL], bad_state_message=None):
```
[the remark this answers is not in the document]
olsen232: Done

sno/merge.py
olsen232: Done

sno/merge.py
olsen232: Added current state to message.

PR #101, opened 2020-06-05
Add support for arbitrary conflict resolutions

olsen232:
```
$ sno conflicts -s --flat
table_A:fid=5
table_A:fid=6
table_A:fid=7

$ sno resolve table_A:fid=5 --with-file=5.geojson
Resolved 1 conflict. 2 conflicts to go.
```

The user can provide an arbitrary number of features in a geojson file, and store them as the resolution to a conflict. Generally, the appropriate number of features will be either 1 - in the case of an edit/edit conflict that needs manual resolution - or 2, in the case of an add/add conflict, the resolution should probably contain both features in some form. (Zero is also a useful number of features, but it is already possible to resolve a conflict by supplying zero features without providing a file).

### DIFF
--- a/sno/gpkg.py
+++ b/sno/gpkg.py
@@ -1,5 +1,6 @@
 import binascii
 import collections
+import json
 import math
 import struct
 
@@ -358,6 +359,15 @@ def ogr_to_gpkg_geom(
         )
 
     return header + envelope + wkb
+
+
+def geojson_to_gpkg_geom(geojson, **kwargs):
+    """Given a GEOJSON geometry, construct a GPKG geometry value."""
+    if not isinstance(geojson, str):
+        json_ogr = json.dumps(geojson)
+
+    ogr_geom = ogr.CreateGeometryFromJson(json_ogr)
+    return ogr_to_gpkg_geom(ogr_geom, **kwargs)
 
 
 def geom_envelope(gpkg_geom):
--- a/sno/resolve.py
+++ b/sno/resolve.py
@@ -1,53 +1,102 @@
+import copy
+import json
+from pathlib import Path
+
 import click
+import pygit2
 
 from .cli_util import MutexOption
+from .gpkg import geojson_to_gpkg_geom
 from .merge_util import MergeIndex, MergeContext, RichConflict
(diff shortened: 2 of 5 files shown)


sno/resolve.py
```diff
@@ -1,53 +1,97 @@
...
+    return result
+
+
+def ungeojson_file(file_path, dataset):
+    """
+    Given a file containing multiple geojson features belonging to dataset,
+    returns the features as dicts containing gpkg geometries.
+    """
+    content = Path(file_path).read_text(encoding="utf-8")
+    features = json.loads(content)["features"]
```
craigds, line 29: Use `json.load(Path(file_path))` to lower memory usage here

sno/resolve.py
```diff
@@ -1,53 +1,97 @@
...
+
+    features = ungeojson_file(file_path, dataset)
+
+    entries = []
+    for feature in features:
+        pk = feature[dataset.primary_key]
+        object_path = "/".join([dataset.path, dataset.get_feature_path(pk)])
+        bin_feature = dataset.encode_feature(feature)
+        blob_id = repo.create_blob(bin_feature)
+        entries += [pygit2.IndexEntry(object_path, blob_id, pygit2.GIT_FILEMODE_BLOB)]
```
craigds, line 47:
```suggestion
        entries.append(pygit2.IndexEntry(object_path, blob_id, pygit2.GIT_FILEMODE_BLOB))
```

sno/resolve.py
craigds: although, there's only going to be two features... so... never mind? 😄

olsen232: Done

sno/resolve.py
olsen232: Done

sno/gpkg.py
```diff
@@ -360,6 +361,18 @@ def ogr_to_gpkg_geom(
     return header + envelope + wkb
 
 
+def json_ogr_to_gpkg_geom(json_ogr, **kwargs):
```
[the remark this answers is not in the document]
olsen232: Done

PR #106, opened 2020-06-08
Better mutually exclusive options

### LINKED ISSUE #99: cli_util.MutexOption doesn't work if all options share the same variable
cli_util.MutexOption seems not to work, as it was designed for something like the following situation - three options exist, "username", "password" and "token" - and the first two are exclusive with the third. All three options have their own variables, and may even be of different types (perhaps token is a file but the others are strings).

However, we are using MutexOption with a single variable `output_format` which accepts any of `--text` `--json` `--geojson` `--html`. We can't actually detect conflicts, since no matter how many of those are specified, the result post-parsing is always just a variable `output_format` and a value eg `text`. For MutexOption to detect conflicts, there needs to be more that one variable.

From the click issue where we got the code:
https://github.com/pallets/click/issues/257
The use-case there is specifically that the different mutually exclusive options have different types or numbers of parameters, so a `click.Choice` option is not appropriate, and each option ends up in a different variable of the appropriate type. In our case, a `click.Choice` option would work fine, but would result in a different syntax- `--output_format=text` instead of `--text`.

The other options we have are finding a fix for cli_util.MutexOption for our use-case, or just deleting MutexOption since it isn't doing anything. click already takes care of choosing an appropriate output_format in case of conflict - it uses whatever was specified last - so we don't necessarily need to detect conflicts.

craigds:
Fixes #99.

The existing MutexOption class only works where the options are given
different names. For most of the cases we're using it, that isn't the
case - the options all have the name `output_format`. By the time the
`MutexOption.handle_parse_result` is called, the options have been
collapsed into one (`{"output_format": "json"}`) which is the last
option to have been specified on the command line.

This change adds a way of handling the options together via a grouping
decorator. The options are parsed individually, but then checked for
mutual exclusion and only one value passed to the command function.

### DIFF
--- a/sno/cli_util.py
+++ b/sno/cli_util.py
@@ -1,4 +1,89 @@
+from functools import wraps
+import uuid
+
 import click
+
+
+def one_of(*option_decorators, name):
+    """
+    Pass a list of options which are mutually exclusive.
+    Only one of them will be allowed.
+    Each option should be a `click.option` call.
+
+    Pass the name as a separate kwarg.
+    The value will be passed to the function as that kwarg.
+
+        @one_of(
+            click.option(
+                "--text", flag_value="text", default=True
+            ),
+            click.option(
+                "--json", flag_value="json"
+            ),
+            name="format"
+        )
+        def mycommand(ctx, format="text"):
+            ...
+    """
+
+    def dec(cmd_func):
+        @wraps(cmd_func)
+        def wrapper(*args, **kwargs):
+            return cmd_func(*args, **kwargs)
+
+        # apply all the arguments to a dummy function so we can figure out
+        # what options we have.
+        def dummy():
+            pass
+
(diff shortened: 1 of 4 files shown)


rcoup:
`-o FORMAT`/`--output-format=FORMAT` consistently could be a good approach (a bit like kubectl). I'm easy with it either way.

rcoup:
Your implementation is

craigds:
I don't disagree lol. click makes this _really hard_ to do sanely. 

If you're all happy with `--output-format=json`, let's just switch to that instead?

olsen232:
I kinda like short options eg "--json" etc but I can see why the "proper" way to do this is supposed to be with a single option

craigds:
Closing this in favour of using `--output-format` / `-o`; don't want to be hacking around click forever just for slightly shorter options

PR #115, opened 2020-06-12
Add a --post-mortem option to debug exceptions

craigds:
While developing sno, it's helpful to have an easy way to break into pdb when uncaught exceptions occur.

This adds a hidden `--post-mortem` flag to make it do that. It uses `ipdb` if installed (it's installed in dev virtualenvs but not in releases)

### DIFF
--- a/sno/cli.py
+++ b/sno/cli.py
@@ -81,7 +81,24 @@ def print_version(ctx):
     ctx.exit()
 
 
-@click.group()
+class PdbGroup(click.Group):
+    def invoke(self, ctx):
+        try:
+            import ipdb as pdb
+        except ImportError:
+            # ipdb is only installed in dev venvs, not releases
+            import pdb
+        if ctx.params.get('post_mortem'):
+            try:
+                return super().invoke(ctx)
+            except Exception:
+                pdb.post_mortem()
+                raise
+        else:
+            return super().invoke(ctx)
+
+
+@click.group(cls=PdbGroup)
 @click.option(
     "-C",
     "--repo",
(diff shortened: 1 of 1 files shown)


sno/cli.py
```diff
@@ -94,8 +111,15 @@ def print_version(ctx):
     "--version", callback=print_version, help="Show version information and exit.",
 )
 @click.option("-v", "--verbose", count=True, help="Repeat for more verbosity")
+# NOTE: this option isn't used in `cli`, but it is used in `PdbGroup` above.
+@click.option(
+    "--post-mortem",
```
rcoup, line 116: wonder if we should have a consistent prefix for experimental/internal options? eg: `--x-post-mortem`

sno/cli.py
```diff
@@ -81,7 +81,24 @@ def print_version(ctx):
     ctx.exit()
 
 
-@click.group()
+class PdbGroup(click.Group):
+    def invoke(self, ctx):
```
rcoup, line 85: I thought using [`sys.excepthook()`](https://docs.python.org/3/library/sys.html#sys.excepthook) is preferred over a try/catch?

https://stackoverflow.com/a/1237407/2662

craigds: TIL. I guess using that allows us to catch exceptions that happen after our command exits, i.e. in whatever click does just as the process is about to exit.

Seems odd but okay.

craigds: I actually like the try/except version better, because it also handles exceptions that will be caught by click (e.g. usage errors). So it comes in useful when you're trying to debug "Error: No commit found at abc132" as well.

sno/cli.py
craigds: IMO, not required. It's not like we're going to have a lot of them, and `hidden=True` already implicitly means we can break them when we want to.

PR #118, opened 2020-06-15
Multi table import

### LINKED ISSUE #92: import: Import multiple tables at once
`sno init --import` and `sno import` handle only one table at a time. To import a whole database you have to run `sno import` multiple times. We can improve this.

# Background

#83 changed the import command semantics a little, adding the `--table={table}` flag.

While we _could_ just make that flag usable multiple times (`--table=table1 --table=table2`) it feels clunky, makes it hard to apply arguments with tools like `xargs`, etc.

A better pattern would be `sno import {source} table1 [table2...]`.

However we also have some existing positional arguments, which would make this approach ambiguous:

* in `sno import`, an existing positional argument (confusingly named `DIRECTORY` in the usage text) which is the imported name of the table. This allows the user to rename tables as they import.
* in `sno init`, a separate thing _also_ called `DIRECTORY`, which is the directory where the repo should be initialised. This is the same as the `git init` argument of the same name.

# Proposal

1. change `--table=TABLE` and turn it into a positional argument with multiple values allowed.
2. To allow for renaming tables during import, use something like `tablename:newtablename` or `tablename@newtablename`. This removes the need for the `DIRECTORY` argument to `sno import`, so we can remove that.
3. Import all tables from a source with an `--all-tables` flag. If `--all-tables` is given, don't support renaming.
4. Turn the `DIRECTORY` arg of `sno init` into a `--path DIRECTORY` option.

craigds:
This changes the `init` and `import` commands so they can import multiple tables at once.

Further, it changes the usage so that:

* there's an `--all-tables` option
* tables are now positional arguments for the import command ~~both commands~~
* any table can be renamed during import using `source_table:dest_table`
* the existing optional (poorly named) `DIRECTORY` argument to `import` has been removed. It renamed tables, so is obsoleted by the above.
* tables cannot be specified in the `init --import` command - it always imports all tables.
* ~~the existing optional `DIRECTORY` argument to the `init` has been moved to a `--path PATH` option. This was required to make table names unambiguous in that command. The command raises a helpful error if you attempt to `init` with a table name without passing `--import`, because that's now clearly a mis-usage.~~

# todo:

tests:

- [x] add a test for multiple table import

Fixes #92 .

### DIFF
--- a/sno/init.py
+++ b/sno/init.py
@@ -514,26 +514,6 @@ def list_import_formats(ctx):
         click.echo(n)
 
 
-def get_table_names_map(tables):
-    names_map = {
-        # dst_table: src_table
-    }
-    for table in tables:
-        (src_table, *rest) = table.split(':', 1)
-        dst_table = rest[0] if rest else src_table
-        if not dst_table:
-            raise click.BadParameter("Invalid table name", param_hint="tables")
-        if is_windows:
-            dst_table = dst_table.replace("\\", "/")  # git paths use / as a delimiter
-
-        if dst_table in names_map:
-            raise InvalidOperation(
-                f'table "{dst_table}" was specified more than once', param_hint="tables"
-            )
-        names_map[dst_table] = src_table
-    return names_map
-
-
 @click.command("import")
 @click.pass_context
 @click.argument("source")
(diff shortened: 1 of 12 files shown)


craigds:
In the latest commit, I changed tack after some offline feedback:


> remove just the option to specify _tables_ from `sno init --import`. ie, you can import in your init command, but you can only import a whole database and you can't rename tables. if you need more fine-grained control you have to use `sno init` and `sno import` seperately.

@rcoup agreed: 

> Agree that any advanced behaviours should use init + import separately.

The latest commit implements that behaviour - it's not possible to specify tables for the `sno init --import` command, and thus it was okay to reinstate the `DIRECTORY` argument for `sno init` as it no longer conflicted with positional table arguments.

sno/init.py
```diff
@@ -577,51 +575,53 @@ def import_table(
...
+        if is_windows:
+            dst_table = dst_table.replace("\\", "/")  # git paths use / as a delimiter
 
-    importer = structure.DatasetStructure.importer(directory, version=version)
-    params = json.loads(os.environ.get("SNO_IMPORT_OPTIONS", None) or "{}")
-    if params:
-        click.echo(f"Import parameters: {params}")
-    importer.fast_import_table(repo, source_loader, message=message, **params)
+        if dst_table in loaders:
+            raise InvalidOperation(
```
olsen232, line 602: Although there's no rule for deciding this, maybe this should be a click.UsageError

sno/structure.py
```diff
@@ -18,6 +18,126 @@
...
+                    click.echo(
+                        f"Importing {num_rows:,d} of {source.row_count:,d} features from {source} to {path}/ ..."
+                    )
+                else:
+                    num_rows = source.row_count
+                    click.echo(
+                        f"Importing {num_rows:,d} features from {source} to {path}/ ..."
+                    )
+
+                t0 = time.monotonic()
```
olsen232, line 85: Repeated code - there's a function you can extract out here. If there are minor differences between how you import meta blobs and how you import feature blobs, you can add it as an option to the function (eg, how the operation is logged)

sno/structure.py
```diff
@@ -18,6 +18,126 @@
...
+                src_iterator = source.iter_features()
```
olsen232, line 86: weird order of statements: src_iterator initialised here but not used until after meta import step is finished.

PR #120, opened 2020-06-16
Added basic Schema support to Dataset2

olsen232:
(schemer)

Allows for writing a feature to the dataset using one schema, and then changing the schema, and reading the schema back.

Supports two ways of representing features - {column-name:value} dicts or row tuples, since both are already used in existing sno code.

More optimised version of this code could be written for reading thousands of features quickly, but this is not my first priority.

### DIFF
--- a/sno/dataset2.py
+++ b/sno/dataset2.py
@@ -1,6 +1,8 @@
 import base64
+from collections import namedtuple
 import functools
 import hashlib
+import json
 import os
 
 import msgpack
@@ -10,12 +12,24 @@ from .structure import IntegrityError
 
 def _pack(data):
     """data (any type) -> bytes"""
-    return msgpack.packb(data)
+    return msgpack.packb(data, use_bin_type=True)
 
 
 def _unpack(bytestring):
     """bytes -> data (any type)"""
     return msgpack.unpackb(bytestring, raw=False)
+
+
+# _json and _unjson are functionally identical to _pack and _unpack,
+# but their storage format is less compact and more human-readable.
+def _json(data):
+    """data (any type) -> bytes"""
+    return json.dumps(data).encode("utf8")
+
+
+def _unjson(bytestring):
+    """bytes -> data (any type)"""
+    return json.loads(bytestring, encoding="utf8")
 
 
 def _b64encode_str(bytestring):
@@ -34,19 +48,9 @@ def _hexhash(data):
     return hashlib.sha256(data).hexdigest()[:40]
 
 
-def _pk_index(c):
-    # TODO - maybe write a class to describe a schema column, so this could be
-    # `c.pk_index` instead of _pk_index(c)
-    return c["primaryKeyIndex"]
-
-
-def _is_pk_column(c):
-    return _pk_index(c) is not None
-
-
 class Legend:
     """
-    A legend is like a table-header that describes how a row is read.
+    A legend is like a table-header that describes how a row is read. Legends are immutable.
     A row is an array of values - eg ["Null Island", POINT(0, 0), "Imaginary points of interest"]
     To read this, we need to look at the legend, which is an array of UUIDs, like:
     ["1144c97-73a6-78d38ab68ed", "2f1551c-a094-c1a3678e317", "2feac23-08af-73e61dbced9"],
@@ -70,8 +74,16 @@ class Legend:
             pk_columns - a list of column IDs for primary key columns.
             non_pk_columns - a list of column IDs for non primary key columns.
         """
-        self.pk_columns = tuple(pk_columns)
-        self.non_pk_columns = tuple(non_pk_columns)
+        self._pk_columns = tuple(pk_columns)
+        self._non_pk_columns = tuple(non_pk_columns)
+
+    @property
+    def pk_columns(self):
+        return self._pk_columns
+
+    @property
+    def non_pk_columns(self):
+        return self._non_pk_columns
 
     @classmethod
     def loads(cls, data):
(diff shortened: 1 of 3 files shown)


sno/dataset2.py
```diff
@@ -122,6 +125,141 @@ def hexhash(self):
...
+
+
+class ColumnSchema:
+    """
+    The schema for a single column. A column has a unique ID that is constant for the columns lifetime
+    - even if the column is moved or renamed - and it has a name, a datatype, info about if this
+    column is one of the primary keys, and potentially extra info about the specific datatype.
+    """
+
+    def __init__(self, id, name, data_type, pk_index=None, extra_type_info=None):
```
rcoup, line 144:
```suggestion
    def __init__(self, id, name, data_type, pk_index=None, **extra_type_info):
```
might be simpler? Simplifies `from_json_dict()` too.

sno/dataset2.py
```diff
@@ -122,6 +125,141 @@ def hexhash(self):
...
+
+    @classmethod
+    def from_json_dict(cls, json_dict):
+        return cls(
+            json_dict.pop("id"),
+            json_dict.pop("name"),
+            json_dict.pop("dataType"),
+            json_dict.pop("primaryKeyIndex"),
+            json_dict,
+        )
```
rcoup, line 159:
```suggestion
        return cls(**json_dict)
```

sno/dataset2.py
```diff
@@ -122,6 +125,141 @@ def hexhash(self):
...
+    - even if the column is moved or renamed - and it has a name, a datatype, info about if this
+    column is one of the primary keys, and potentially extra info about the specific datatype.
+    """
+
+    def __init__(self, id, name, data_type, pk_index=None, extra_type_info=None):
+        self.id = id
+        self.name = name
+        self.data_type = data_type
+        self.pk_index = pk_index
+        self.extra_type_info = extra_type_info if extra_type_info is not None else {}
```
rcoup, line 149:
```suggestion
        self.extra_type_info = extra_type_info
```

sno/dataset2.py
```diff
@@ -122,6 +125,141 @@ def hexhash(self):
...
+        non_pk_column_ids = []
+        for i, column in enumerate(sorted(self.columns, key=_pk_index_ordering)):
+            if column.pk_index is not None:
+                if i != column.pk_index:
+                    raise ValueError(
+                        f"Expected contiguous primaryKeyIndex {i} but only found {column.pk_index}"
+                    )
+                pk_column_ids.append(column.id)
+            else:
+                non_pk_column_ids.append(column.id)
```
rcoup, line 247: Doesn't this mean non-PK columns end up in a random order?

My understanding is we want the legend/blob to be sorted by column index, but the PK column values aren't actually present in the feature blob (because they're in the tree path).

sno/dataset2.py
```diff
@@ -122,6 +125,141 @@ def hexhash(self):
...
+            self.id == other.id
+            and self.name == other.name
+            and self.data_type == other.data_type
+            and self.pk_index == other.pk_index
+            and self.extra_type_info == other.extra_type_info
+        )
+
+
+class Schema:
+    """A schema is just an ordered list of ColumnSchemas."""
```
rcoup, line 183: Schema and Legend classes are immutable, right? Do we want to enforce that? Or alternatively, document it?

tests/test_dataset2.py
```diff
@@ -1,6 +1,12 @@
 import os
+import pytest
+import uuid
 
-from sno.dataset2 import Dataset2, Legend
+from sno.dataset2 import Dataset2, Legend, ColumnSchema, Schema
+
+
+def _uuid():
+    return str(uuid.uuid4())
```
rcoup, line 9: Would probably be simpler to have the same UUIDs for different test runs...

```suggestion
@pytest.fixture
def uuid(request):
    """ Deterministic "random" UUID generator seeded from the test ID """
    seed = int(hashlib.sha1(request.node.nodeid).hexdigest(), 16)
    _uuid_gen = random.Random(seed)
    def _uuid():
        return str(uuid.UUID(int=_uuid_gen.getrandbits(128)))
    return _uuid
```

(though should probably move to conftest)

sno/dataset2.py
rcoup: hmm, reading back that doesn't make a huge lot of sense. 😉

sno/dataset2.py
olsen232: Enforced, more-or-less.

sno/dataset2.py
rcoup: Something needs to put the columns in the right order for working copies/exports/whatever — seems like it should be the Dataset class responsibility?

olsen232: Theory: because python sort is stable, this sort leaves non-PK columns stay in their row order. Practice: There is a test that correctly roundtrips a feature in spite of the schema changing column order and in spite of the feature dictionaries provided having the fields initialised in random order.

sno/dataset2.py
olsen232: Done

sno/dataset2.py
olsen232: Done, except the JSON names don't line up with the python names - partly because of our mild convention for short python symbols, but also due to the strong python convention of underscore_case symbols

sno/dataset2.py
olsen232: Done

sno/dataset2.py
olsen232: Dataset2.read_feature_tuple returns the row in row order
Dataset2.read_feature_dict returns a dict that has been initialised in row order, so that iterating over stays in row order

Schema.feature_tuple_from_raw and Schema.feature_dict_from_raw are where the data is put into row order - they iterate through schema.columns and add data for each column in order. (They might need to be optimised one day, every feature read creates a temporary dict)

test_schema_change_roundtrip shows that reading a feature gives the feature in row-order, even if the row order is changed due to a new schema.

tests/test_dataset2.py
olsen232: Done

PR #126, opened 2020-06-19
Add a pull request template

craigds:
## Description

Since we now have a changelog in the codebase, it makes sense to have the PR template prompt us to update it.

I tried to make a PR template which is helpful yet not onerous; we don't want to prevent contributions by demanding millions of things before submission. It should serve as a reminder of good practice.

## Related links:

* #125 

## Checklist:

- [x] ~Have you included test(s)?~ N/A
- [x] ~Have you updated the [changelog](https://github.com/koordinates/sno/blob/master/CHANGELOG.md)?~ N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,13 @@
+## Description
+
+<!-- Thanks for your contribution! Describe your changes in a few sentences. Try to include discussion of tradeoffs or alternatives you considered when writing this code. -->
+
+## Related links:
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+## Description
+
+_Thanks for your contribution! Describe your changes in a few sentences. Try to include discussion of tradeoffs or alternatives you considered when writing this code._
```
rcoup, line 3: Maybe use `<!--` ... `-->` instead, so it doesn't show up in the preview view by default?

PR #131, opened 2020-06-22
Add dataset:primary_key arg filters to sno commit, sno conflicts

olsen232:
```
$ sno diff
... 100 changes
$ sno commit -m partial_commit datasetA:1
[master d4b583e] partial_commit
  datasetA/
    modified:  1 feature
  Date: Tue Jun 23 09:44:24 2020 +1200
```
But perhaps more usefully, changes of one entire dataset can be committed without committing changes in other datasets:
```
sno commit -m datasetA_changes datasetA
```

Resolves https://github.com/koordinates/sno/issues/69 (last blocker for 0.4 release)

Also gets rid of the working copy callback - working copy changes must now be called in a regular kind of way, instead of as a callback. This is because -
- the callback wasn't very necessary - it was called hundreds of times during the commit operation with arguments of "I", "U", "D", but these were no-ops, and then it was called once or twice after the commit operation with arguments like "INDEX", "TREE" - these didn't need to be callbacks since they could just be called by the caller once the commit operation completed
- there was only one implementation of the callback, so we may as well make explicit that the callback is actually just the working copy object
- the callback had special semantics - ie, the commit code had to know to call the callback with arguments like "I", "U", "D", "INDEX", "TREE". It wasn't just an arbitrary function that could be called without knowing what it was. So again, we may as well make these different operations methods on the working copy, and call them explicitly at the appropriate time. (If we do ever need to reintroduce the callback, it should probably have an interface)

### DIFF
--- a/sno/conflicts.py
+++ b/sno/conflicts.py
@@ -4,6 +4,7 @@ import sys
 import click
 
 from .exceptions import SUCCESS, SUCCESS_WITH_FLAG
+from .filter_util import build_feature_filter, UNFILTERED
 from .merge_util import MergeIndex, MergeContext, rich_conflicts
 from .output_util import dump_json_output
 from .repo_files import RepoState
@@ -18,7 +19,12 @@ _CONFLICT_PLACEHOLDER = object()
 
 
 def list_conflicts(
-    merge_index, merge_context, output_format="text", summarise=0, flat=False
+    merge_index,
+    merge_context,
+    output_format="text",
+    conflict_filter=UNFILTERED,
+    summarise=0,
+    flat=False,
 ):
     """
         Lists all the conflicts in merge_index, categorised into nested dicts.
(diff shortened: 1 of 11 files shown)


sno/conflicts.py
```diff
@@ -242,7 +251,8 @@ def conflicts_json_as_geojson(json_obj):
     hidden=True,
     help="Output all conflicts in a flat list, instead of in a hierarchy.",
 )
-def conflicts(ctx, output_format, exit_code, json_style, summarise, flat):
+@click.argument("args", nargs=-1)
```
craigds, line 254: Check `sno conflicts --help` for this, probably improve it a bit.

 * call it `filters` rather than args
 * maybe set `metavar` to something, if it makes the help text more readable: https://click.palletsprojects.com/en/7.x/documentation/#meta-variables
 * Add `help` with a filter example

sno/merge_util.py
```diff
@@ -562,6 +563,16 @@ def output(self, output_format):
...
             return geojson_row(self.feature, self.pk_field)
 
+    def matches_filter(self, conflict_filter):
+        if conflict_filter == UNFILTERED:
+            return True
+        dataset_filter = conflict_filter.get(self.dataset_path, None)
+        if not dataset_filter:
+            return False
+        return dataset_filter == UNFILTERED or (
+            not self.is_meta and str(self.pk) in dataset_filter
```
craigds, line 573:
```suggestion
            str(self.pk) in dataset_filter and not self.is_meta
```
(same thing, just clearer precedence)

sno/working_copy.py
```diff
@@ -749,33 +750,42 @@ def diff_to_tree(self, repo_structure, feature_filter=UNFILTERED):
...
-            elif action == "INDEX":
-                dbcur.execute(
-                    f"DELETE FROM {self.TRACKING_TABLE} WHERE table_name=?;",
-                    (dataset.name,),
-                )
+            for dataset, pks in reset_filter.items():
+                if pks == UNFILTERED:
+                    dbcur.execute(
+                        f"DELETE FROM {self.TRACKING_TABLE} WHERE table_name=?;",
+                        (gpkg.ident(dataset),),
```
craigds, line 766: this seems wrong... gpkg.ident double-quotes identifiers (`"mytablename"`), but here the table name is a regular parameter, so it will get single-quoted automatically (`'mytablename'`) by the apsw parameter substitution. If it's double quoted it won't ever match anything in this query

tests/test_commit.py
```diff
@@ -62,6 +62,21 @@ def edit_table(dbcur):
...
+def _count_tracking_table_changes(db, working_copy, layer):
+    with db:
+        cur = db.cursor()
+        cur.execute(
+            f"SELECT COUNT(*) FROM {working_copy.TRACKING_TABLE} WHERE table_name=?;",
+            [layer],
+        )
+        change_count = cur.fetchone()[0]
+        del cur
+    return change_count
```
craigds, line 74:
```suggestion
        return cur.fetchone()[0]
```



don't need the `del` because it gets garbage-collected on return anyway

sno/conflicts.py
olsen232: Done, also modified commit.py and diff.py which are the other commands that currently support filters.

sno/merge_util.py
olsen232: Not done - `self.pk` fails if `self.is_meta`

To make it a bit more readable, I added self.is_feature that is just the negation of self.is_meta

sno/working_copy.py
olsen232: Fixed

tests/test_commit.py
olsen232: Done

olsen232:
Filters that don't match anything just... don't match anything. There's no warning, unless you try to commit an empty commit.
```
$ sno diff HEAD # shows all diffs
$ sno diff HEAD dataset_A # shows all diffs in dataset_A
$ sno diff HEAD nonexistent # shows all diffs in nonexistent, even though there's none

$ sno commit -m "message" nonexistent
Error: No changes to commit (exit_code=NO_CHANGES)
```

craigds:
> Filters that don't match anything just... don't match anything. There's no warning, unless you try to commit an empty commit.
ok. if it's a problem we can add a warning later maybe 👍

PR #132, opened 2020-06-23
When `hub` isn't authed, fail explicitly

craigds:
## Description
At present if `hub` is installed but `~/.config/hub` doesn't contain
valid auth, `make vendor` hangs while appearing to download github
artifacts. Turns out that `hub` is waiting for input at an invisible
prompt.

This dies with a readable error instead

## Related links:
https://hub.github.com/hub.1.html#github-oauth-authentication

## Checklist:

- [ ] Have you included test(s)? **no - but have tested this manually with ~/.config/hub emptied out.**
- [ ] Have you updated the [changelog](https://github.com/koordinates/sno/blob/master/CHANGELOG.md)? **N/A**

### DIFF
--- a/vendor/github-download.sh
+++ b/vendor/github-download.sh
@@ -56,16 +56,23 @@ fi
 
 cd "$(dirname "$0")"
 mkdir -p dist/
+DOWNLOADED=false
 if command -v hub >/dev/null; then
-    # if 'hub' is installed, use its authenticated access to the API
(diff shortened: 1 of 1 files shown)


craigds:
could do. Is that better?

If hub is installed it's much easier to configure it manually than to download the packages manually

rcoup:
Sure, but if they have no idea what hub is then they have an option to do it manually.

PR #139, opened 2020-06-26
import: Options to set dataset metadata

### LINKED ISSUE #102: Add alternative ways of importing title, description, XML metadata
importing from GPKG pulls title, description, and XML metadata from `gpkg_contents` and `gpkg_metadata`, but no such thing exists in most other formats.

As we add other import sources, it'd be helpful to have the option to import that metadata in other ways.

I propose:

* `--title="NZ Memorials List"`
* `--description-from=./description.txt`
* `--metadata-from=./metadata.xml`

The description one could also have an inline `--description="This is the description"` variant, but I'm not sure how useful that would be. It wouldn't be very friendly for the XML metadata one.

craigds:
## Description

Adds a `--table-info` option which adds extra dataset metadata via a JSON object. Fixes #102

The JSON object is a bit smelly, but I tried hard to think of a better way and couldn't. Because the `import` command now imports multiple tables at once, we need a way to set title, description and xml metadata for multiple layers at once.

Some alternatives I tried:

1. `sno import SRC table1 --title=title2 table2 --title=title2`

This would probably be the nicest approach in terms of having a flat schema to the command line arguments, though it's a little at odds with how these commands _generally_ work I think.

I couldn't find any _simple_ way to do this in click. It's possible we could abuse [multi-command chaining](https://click.palletsprojects.com/en/7.x/commands/#multi-command-chaining) to do it if we tried really hard, but it'd be a hack.

2. `sno import SRC dataset1 dataset2 --dataset1-title title1 --dataset2-title title2`

This would work but it once again seems hard to do in click. We'd basically have to give up on the parser and accept _everything_ and handle validation ourselves.

3. this approach.

```
  --table-info JSON               Specifies overrides for imported tables, in
                                  a nested JSON object.  Each key is a dataset
                                  name, and each value is an object.  Valid
                                  overrides are 'title', 'description' and
                                  'xmlMetadata'. e.g.:  --table-
                                  info='{"land_parcels": {"title": "Land
                                  Parcels 1:50k"}' To import from a file,
                                  prefix with `@`, e.g. `--table-
                                  info=@filename.json`
```

Screen Shot 2020-06-26 at 4 38 58 PM

The JSON validates to a schema, so if you do something unexpected, you'll get a verbose but slightly helpful error message:
Screen Shot 2020-06-26 at 4 37 48 PM

## Checklist:

### DIFF
--- a/sno/cli_util.py
+++ b/sno/cli_util.py
@@ -1,4 +1,6 @@
+import json
 import click
+import jsonschema
 
 
 class MutexOption(click.Option):
@@ -71,6 +73,29 @@ class StringFromFile(click.types.StringParamType):
             fp = filetype.convert(filename, param, ctx)
             return fp.read()
 
+        return value
+
+
+class JsonFromFile(StringFromFile):
+    name = 'json'
+
+    def __init__(self, schema=None, **file_kwargs):
+        super().__init__(**file_kwargs)
+        self.schema = schema
+
+    def convert(self, value, param, ctx):
+        value = super().convert(value, param, ctx)
+        try:
+            value = json.loads(value)
+        except json.JSONDecodeError as e:
+            self.fail(
+                f"Invalid JSON: {e}", param, ctx,
+            )
+        if self.schema:
+            try:
+                jsonschema.validate(instance=value, schema=self.schema)
+            except jsonschema.ValidationError as e:
+                self.fail(str(e), param, ctx)
         return value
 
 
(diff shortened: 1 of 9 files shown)


sno/cli_util.py
```diff
@@ -74,6 +76,29 @@ def convert(self, value, param, ctx):
...
+            value = json.loads(value)
+        except json.JSONDecodeError as e:
+            self.fail(
+                f"Invalid JSON: {e}", param, ctx,
+            )
+        if self.schema:
+            try:
+                jsonschema.validate(instance=value, schema=self.schema)
+            except jsonschema.ValidationError as e:
+                raise click.BadParameter(str(e), ctx=ctx, param=param,) from e
```
rcoup, line 98: why does this error case use an exception and the other one use `self.fail()`?

craigds: not sure, i'll make this use self.fail() (the exception is ultimately BadParameter anyway)

PR #141, opened 2020-06-29
Basic gpkg working copy for dataset v2

olsen232:
Make existing working copy code work with dataset v2 by making dataset v2 present the more of the same interface as V1: calling get_meta_item("gpkg_contents") will generate a gkpg_contents meta item, even though none is stored directly.

Adds v2 dataset equivalents to points, polygons, and tables archives for testing.

(Note: the datasets v1 / datasets v2 interfaces, and the  org importer / gpkg importer interfaces - are slowly converging, since they actually all do the same thing: expose a list of features, and the metadata about them. At some point, we could put more work into formalising this interface, which could simplify conversions between these different formats and make the code a bit more structured.)

Still TODO in future changes: 
- diffs (commit<>commit or commit<>workingcopy) don't work due to missing code in dataset v2 and because dataset v2 can have multiple primary keys, existing diff code requires exactly one primary key
- then, commits
- changing the working copy's schema will probably still break things
- more work on SRS IDs - conversion code assumes "EPSG" in all cases.
- internal geometry representation is still the same as V1, should be updated for V2


sno/dataset2.py
```diff
@@ -541,12 +562,21 @@ def import_iter_meta_blobs(self, repo, source):
...
 
+        # TODO - tidy up SRS ID code.
+        for srs in source.get_meta_item("gpkg_spatial_ref_sys"):
+            meta_blobs.append(
+                (f"{self.SRS_PATH}EPSG:{srs['srs_id']}.wkt", srs["definition"])
+            )
+
         for meta_path, meta_content in meta_blobs:
-            yield self.repo_path(meta_path), _bytes(meta_content)
+            if meta_content:
```
craigds, line 578: is this correct? Means meta blobs with empty-string values won't be imported, which sounds incorrect. e.g. `description` is commonly an empty string.

sno/dataset2_gpkg.py
```diff
@@ -1,17 +1,90 @@
...
+def v2_to_sqlite_table_info(dataset2):
+    return [_columnschema_to_gpkg(i, col) for i, col in enumerate(dataset2.schema)]
+
+
+def _get_geometry_columns(schema):
+    return [c for c in schema.columns if c.data_type == "geometry"]
+
+
+def srs_str_to_int(srs_str):
+    # This should be more complicated.
```
craigds, line 75: what other forms could `srs_str` take here?

Feels like it should raise an error rather than 0 if it doesnt' understand the value given.

Also see `OGRImporter._get_meta_srid` which does some conversion via OGR's SRS objects to return an integer

sno/dataset2.py
olsen232: Good catch. Changed to "is not None" - that way, empty string / empty array is preserved as different to missing.

sno/dataset2_gpkg.py
olsen232: Well, right now the only SRS strings that this code deals with when creating gpkg from v2 datasets are those that were generated by the inverse function, srs_int_to_str, when creating the v2 dataset in the first place.
So right now this function is guaranteed to work. You're right, it should raise a ValueError if it can't handle it.

However, the int `srs_id` we get from geopackage during import / upgrade might not always be intended to be an EPSG identifier. The spec says that this could be `urn:ogc:def:crs:EPSG::2193` or basically any string eg `kx_custom_srs_512`, as long as a definition of the same name is included. But, it's not clear at the moment if or how we would end up with free form strings, since all our imports go via GPKG which requires an authority (string) + id (integer) format. Anyway, needs more work.

rcoup: -1/0 pretty commonly map onto "unknown". Which we shouldn't encourage, but kinda need to work.

rcoup: [`osgeo.osr.SpatialReference`](https://gdal.org/python/osgeo.osr.SpatialReference-class.html) is the way we validate/use spatial reference definitions. (`ImportFrom*` functions)

PR #144, opened 2020-07-01
Fix v2 SRS import / upgrade / export

olsen232:
As of previous commit, SRS definitions were assumed to be EPSG. Now, we actually parse the WKT in the SRS definition and extract the authority name and id from it.

Uh oh: fixing this ended up being a refactoring change also. I have begun the work of moving GPKG specific things into gpkg_adapter (renamed from dataset2_gpkg), starting with "gpkg_spatial_ref_sys". Now that GPKG is not our model, importers / datasets should not have an accessor for "gpkg_spatial_ref_sys" as their main interface. (Right now they still have this accessor, but providing it is delegated to the adapter class.) Instead, the main interface going forward is a list of (srs_name, srs_definition_in_wkt) tuples, which is basically what is stored in V2 - which makes sense as the V2 design is supposed to be simple and all-purpose.

This PR is the start of an inversion - instead of importers first importing to GPKG, and then this is adapted to V2, we should import to V2 - that is, our new model - and have adapters to adapt that to gpkg. See gpkg_adapter. More changes along these lines will come as V2 supercedes V1 entirely and V1-centric structure is removed from codebase, although this cleanup might be well into V2 development. For instance, eventually code to convert OGR types to GPKG types should go away, and instead we'll convert OGR types to V2 types and then, if necessary, from V2 to GPKG.

Import sources and datasets continue to converge - still hope to define this interface properly at some point and make them all implement it, move secondary functionality (eg GPKG meta item generation) into an adapter that works on this interface.


sno/gpkg_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+        ]
+    )
+
+
+def _sort_by_cid(sqlite_col_info):
+    return sqlite_col_info["cid"]
+
+
+def _gpkg_to_column_schema(
+    sqlite_col_info, gpkg_geometry_columns, gpkg_spatial_ref_sys, id_salt
```
craigds, line 161: THis one's complicated enough that I think it deserves a docstring

sno/gpkg_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+            gpkg_geometry_columns, gpkg_spatial_ref_sys,
+        )
+    else:
+        data_type, extra_type_info = gpkg_type_to_v2_type(sqlite_col_info["type"])
+
+    col_id = ColumnSchema.deterministic_id(name, data_type, id_salt)
+    return ColumnSchema(col_id, name, data_type, pk_index, **extra_type_info)
+
+
+def _columnschema_to_gpkg(cid, column_schema):
```
craigds, line 176: should use `columnschema` or `column_schema` consistently (also see _gpkg_to_column_schema )

sno/gpkg_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+    return _gpkg_spatial_ref_sys(spatial_ref, spatial_ref.ExportToWkt())
+
+
+def _gpkg_spatial_ref_sys(spatial_ref, wkt):
+    spatial_ref.AutoIdentifyEPSG()
+    return [
+        {
+            "srs_name": spatial_ref.GetName(),
+            "definition": wkt,
+            "organization": spatial_ref.GetAuthorityName(None),
```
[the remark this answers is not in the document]
olsen232: Passing None means "search for an authority node on the root element" which as far as I know is the desired behaviour.

sno/gpkg_adapter.py
olsen232: Done

sno/gpkg_adapter.py
olsen232: Done

sno/gpkg_adapter.py
```diff
@@ -0,0 +1,257 @@
...
+    srs_pathname = geom_columns[0].extra_type_info["geometrySRS"]
+    definition = dataset.get_srs_definition(srs_pathname)
+    return wkt_to_gpkg_spatial_ref_sys(definition)
+
+
+DEFAULT_GPKG_SPATIAL_REF_SYS = [
+    {
+        "srs_name": "Unknown CRS",
+        "definition": "",
+        "organization": "EPSG",
```
rcoup, line 85: `EPSG:0` isn't a thing. We should follow the approach of http://www.geopackage.org/spec/#r11 here, and use `NONE` if we really need a string.

sno/gpkg_adapter.py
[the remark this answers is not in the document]
rcoup: What authority does GDAL return for unknown? Lots of WKT doesn't include authority code, their use is relatively recent.

Note that GeoPackage uses WKT1, we should be storing WKT2 definitions. GDAL can export/convert

sno/gpkg_adapter.py
```diff
@@ -0,0 +1,257 @@
...
+def wkt_to_srs_str(wkt):
+    """Given a WKT srs definition, generate a sensible name for it."""
+    return osgeo_to_srs_str(SpatialReference(wkt))
+
+
+def osgeo_to_srs_str(spatial_ref):
+    """Given a osgeo SpatialReference, generate a sensible name for it."""
+    auth_name = spatial_ref.GetAuthorityName(None)
+    auth_code = spatial_ref.GetAuthorityCode(None)
+    return f"{auth_name}:{auth_code}"
```
rcoup, line 130: similar q around what is returned for unknown/custom CRS

sno/gpkg_adapter.py
```diff
@@ -0,0 +1,257 @@
...
+            "definition": wkt,
+            "organization": spatial_ref.GetAuthorityName(None),
+            "srs_id": spatial_ref.GetAuthorityCode(None),
+            "organization_coordsys_id": spatial_ref.GetAuthorityCode(None),
+            "description": None,
+        }
+    ]
+
+
+DEFAULT_SRS_STR = "EPSG:0"
```
rcoup, line 118: EPSG:0 isn't a thing

sno/gpkg_adapter.py
olsen232: This preserves the previous import behaviour where we always assumed "EPSG", but srs_id would return 0 if we couldn't find a SpatialReference. But, happy to change it to "NONE"

rcoup: GPKG specifically has set values (NONE:0 and NONE:-1) that must be in every GPKG, but I think internally we should probably just not have a crs (& wkt) set? And the working copies can decide how to represent that (eg. PostGIS uses 0)

PR #149, opened 2020-07-03
`import`: Fix import on empty branches

craigds:
## Description

If you're on an empty branch, but there are commits elsewhere
in the repository, imports currently fail with confusing errors.
This fixes the bug.

## Related links:

none

## Checklist:

### DIFF
--- a/sno/structure.py
+++ b/sno/structure.py
@@ -31,12 +31,20 @@ def write_blobs_to_stream(stream, blobs):
 def fast_import_tables(
     repo, sources, max_pack_size="2G", limit=None, message=None, *, version
 ):
+    head_tree = None
+    if not repo.is_empty:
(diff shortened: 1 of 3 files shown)


sno/structure.py
```diff
@@ -76,7 +84,7 @@ def fast_import_tables(
             )
             p.stdin.write(header.encode("utf8"))
 
-            if not repo.is_empty:
+            if head_tree is not None:
                 # start with the existing tree/contents
                 p.stdin.write(b"from refs/heads/master^0\n")
```
olsen232, line 89: Maybe put a TODO next to this refs/heads/master thing because, we probably shouldn't always be importing to master. Don't need to fix it in this PR but worth flagging.

PR #150, opened 2020-07-03
Begin rewrite of working copies for v2

olsen232:
This does the very first and simplest thing needed for V2 working copies - now that the model can have multiple primary keys, the tracking table can't be a single table of `| table | primarykey |`. Instead, multiple tracking tables are needed - one per actual table. Each tracking table's columns should be the same as the *primary key* columns of the table it tracks.

This is done in a WorkingCopy_GPKG_2 class. We also need to know when to use the v2 working copy - basically, whenever a repo has v2 datasets. We don't have an easy way of checking that, but I've added a ".sno-version" file to the repository as per https://github.com/koordinates/sno/issues/142

Also took the opportunity to rename the meta and tracking tables to start with "gpkg" - see https://github.com/koordinates/sno/issues/63
Meta table: `gpkg.sno.meta`
Tracking tables: `gpkg.sno.<table_to_be_tracked>.track`

Instead of the old names, `.sno-meta`  and`.sno-track`. As before, our tables are obviously distinct from the actual gpkg tables, even though ours now also start with gpkg.

Tests:
Existing tests pass. Nothing makes use of the new tracking tables yet, diff needs to be rewritten for v2.


rcoup:
:+1: `gpkg_{whatever}`.

> Instead, multiple tracking tables are needed - one per actual table. Each tracking table's columns should be the same as the primary key columns of the table it tracks.

Would it be possible to use a single table? Otherwise means keeping track of dataset renames/adds/deletes/etc at a schema level as well as a data level.

> And what happens if/when the PK changes? 

aahhh, this isn't a thing in SQLite, but will be for other working copies I guess. (Well, the way to change the working copy is create a new table, populate it, drop the old table and rename).

olsen232:
I punted this for now since nobody is actually using multiple-primary-key tables and since it doesn't help with the higher priority goal of supporting schema changes. So for now, we are still using a single tracking table, exactly as before

PR #152, opened 2020-07-06
Odds & ends: make working-copy diff work for datasets v2

olsen232:
## Description

Datasets v1 and v2 interfaces are pretty similar (eventually they should either have the same public interface, or v1 should be deleted, whichever comes first). They were almost similar enough for the dataset<->gpkg-working-copy diff code to work. I have made a few more changes now so that it does, and enabled those tests.

### Still TODO
(in the area of diffs specifically):
- meta item diffs (needs to be rewritten)
- writing commits (which is based on diffs)

### Changes to v2 datasets
- a few new methods, some of which hopefully won't stay in the interface forever:
    `primary_key(self)` `encode_feature_blob(self, feature)`
- handle primary key values that should be ints, but arrived as strings
- accept `ogr_geoms` flag even though it's not supported

### Changes to v1 datasets
- `get_feature` now just returns a feature, not a `(path, feature)` tuple - none of the callers were using the path anyway.

### DIFF
--- a/sno/dataset2.py
+++ b/sno/dataset2.py
@@ -259,6 +259,9 @@ class Schema:
         self._columns = tuple(columns)
         # Creating a legend validates the primaryKeyIndex field.
         self._legend = self._to_legend()
+        self._pk_columns = tuple(
+            c for c in sorted(columns, key=pk_index_ordering) if c.pk_index is not None
+        )
 
     @property
     def columns(self):
@@ -267,6 +270,10 @@ class Schema:
     @property
     def legend(self):
         return self._legend
+
+    @property
+    def pk_columns(self):
+        return self._pk_columns
 
     def __getitem__(self, i):
         """Return the _i_th ColumnSchema."""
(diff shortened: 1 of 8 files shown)


sno/dataset2.py
```diff
@@ -593,3 +618,18 @@ def import_iter_feature_blobs(self, resultset, source):
         for feature in resultset:
             feature_path, feature_content = self.encode_feature(feature, schema)
             yield self.repo_path(feature_path), feature_content
+
+    @property
+    def primary_key(self):
+        # TODO - datasets v2 model supports more than one primary key.
+        # This function needs to be changed when we have a working copy v2 that does too.
+        for column in self.schema:
+            if column.pk_index == 0:
```
craigds, line 627: I think this should raise a NotImplementedError if there are >1, rather than silently returning the first one

sno/dataset2.py
```diff
@@ -348,6 +355,23 @@ def is_pk_compatible(self, other):
...
+        as text and so had to be parsed from text:
+        1. pk_values should be a list / tuple, with one value per primary key column.
+           (There is hardly ever >1 primary key column, but as this is what our model supports, we need a list.)
+        2. integer columns need int values, so if the values were supplied as text, we need to cast to int here.
+        """
+        if not isinstance(pk_values, (list, tuple)):
+            pk_values = [pk_values]
+        for i, (value, column) in enumerate(zip(pk_values, self.pk_columns)):
+            if column.data_type == "integer" and isinstance(value, str):
+                if not isinstance(pk_values, list):
```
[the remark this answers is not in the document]
olsen232: Done - for some reason I was trying to optimise for the case where the input is already fine, I didn't want to create more objects. But, didn't really make sense

sno/dataset2.py
olsen232: Good point. Done

PR #155, opened 2020-07-07
Merge sno-v2-internal-data-format (so far) to master

olsen232:
## Description

Implements the Datasets V2 layout/structure, and adds import and upgrade functionality for the same. Very limited functionality beyond this, but V1 as the default for new repositories and leaves V1 functionality intact.

## Related links:

## Checklist:


sno/dataset2.py
```diff
@@ -0,0 +1,634 @@
...
+    def srs_definitions(self):
+        """Return all stored srs definitions in a dict."""
+        for blob in find_blobs_in_tree(self.tree / self.SRS_PATH):
+            # -4 -> Remove ".wkt"
+            yield blob.name[:-4], _text(blob.data)
+
+    @functools.lru_cache()
+    def get_legend(self, legend_hash=None, *, path=None):
+        """Load the legend with the given hash / at the given path from this dataset."""
+        if path is None:
```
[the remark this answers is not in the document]
olsen232: Short answer no we don't. Removed.

Will take it away for now but I wonder is there some value in standardising on the structure of these methods - ie, all methods should be called `(encode|get)_(feature|raw_feature_dict|legend)` and have a signature of `(self, <which>, *, path=None, data=None)` - where argument `<which>` is the most natural way of specifying which item, but path or path+data can be used if the caller doesn't know which item, they just know which file it is stored in.

But, the different methods dealing with features/raw_feature_dicts/legends will always have to have slightly different signatures anyway, so maybe there's no point tp pretending that someone could call all of them without even bothering to look up the different signatures.

sno/dataset1.py
```diff
@@ -218,68 +218,70 @@ def tupleizer(pk, blob):
...
+                    self.decode_pk(blob.name) if c == -1 else bin_feature[c]
                     for c in ftuple_order
                 ]
             )
 
         return tupleizer
 
-    def _features(self, feature_builder, fast):
-        top_tree = self.tree / ".sno-table"
+    def _iter_feature_blobs(self, feature_builder, fast=False):
```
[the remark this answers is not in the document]
olsen232: Makes sense. Done

sno/dataset1.py
```diff
@@ -218,68 +218,70 @@ def tupleizer(pk, blob):
...
 
     def features(self, *, ogr_geoms=False, **kwargs):
-        """ Feature iterator yielding (feature-key, feature-dict) pairs """
-        return self._features(
-            lambda pk, blob: (
-                pk,
-                self.repo_feature_to_dict(pk, blob, ogr_geoms=ogr_geoms),
-            ),
-            fast=False,
+        """ Feature iterator yielding (pk, feature-dict) pairs """
```
rcoup, line 269: pk here could be a tuple itself?

rcoup: A: no, looks like it's an encoded PK value[s]

sno/gpkg_adapter.py
```diff
@@ -0,0 +1,251 @@
...
+    ]
+
+
+def wkt_to_srs_str(wkt):
+    """Given a WKT srs definition, generate a sensible name for it."""
+    return osgeo_to_srs_str(SpatialReference(wkt))
+
+
+def osgeo_to_srs_str(spatial_ref):
+    """Given a osgeo SpatialReference, generate a sensible name for it."""
```
rcoup, line 121:
```suggestion
    """Given a osgeo SpatialReference, generate a sensible identifier for it."""
```

FYI, a human-readable name/description comes from the top-level PROJCS/GEOGCS element, and it's available via `srs.GetName()`.

```python
>>> srs = osr.SpatialReference()
>>> srs.ImportFromEPSG(2193)
>>> srs.GetName()
'NZGD2000 / New Zealand Transverse Mercator 2000'
```

sno/init.py
```diff
@@ -888,7 +965,10 @@ def import_table(
             xml_metadata=info.get('xmlMetadata'),
         )
 
-    structure.fast_import_tables(repo, loaders, message=message, version=version)
+    extra_blobs = []
+    if version == 2:
+        extra_blobs = [(VERSION_BLOB_PATH, b"0.5\n")]
```
rcoup, line 970: Feels like we don't want 0.5/0.6/0.12 stored in the repos if the datasets are compatible? 

Ie. I think this should be a version of breaking changes to repository structure — upgrading Sno (with a compatible repo structure) shouldn't mean you have to (or are even encouraged to) commit.

So dataset v1 vs v2 rather than Sno v0.5 vs v0.6.

sno/dataset2.py
```diff
@@ -593,3 +618,17 @@ def import_iter_feature_blobs(self, resultset, source):
...
+        # TODO - datasets v2 model supports more than one primary key.
+        # This function needs to be changed when we have a working copy v2 that does too.
+        if len(self.schema.pk_columns) == 1:
+            return self.schema.pk_columns[0].name
+        raise ValueError(f"No single primary key: {self.schema.pk_columns}")
+
+    def encode_feature_blob(self, feature):
+        # TODO - the dataset interface still needs some work:
+        # - maybe encode methods shouldn't be classmethods.
+        # - having a _blob version of encode_feature is too many similar methods.
```
rcoup, line 632: Best approach is probably to profile various dataset operations (checkout/import/diff) and fix the hotspots — that's what I did first time around and why there was a proliferation of similar-but-not-quite methods. 

I'd hope our new structure is slightly cleaner and they're maybe needed less.

sno/dataset1.py
olsen232: It looks like I made two changes here:

1. Change "feature-key" to "pk" in the comment, in line with the variable name in the code.
But, this was actually misleading since the "pk" variable did not store a pk value, but the blob name (ie, an encoded pk value).
2. Refactored the code so that the variable name "pk" disappeared. (This is dataset v1 so not valueS.)

Will fix the docstring.

sno/gpkg_adapter.py
olsen232: Done

sno/init.py
olsen232: Changed to 1, 2 etc in ".sno-format"

sno/dataset2.py
olsen232: This particular method will hopefully go away if we fix diffs to not manually compare blob hashes for rename detection, and leave that to pygit2, since that is the only caller.

Issue #157, opened 2020-07-08
Progress output for Spatial Index creation


UberMouse:
Creation of the Spatial Index can be quite lengthy for larger datasets. The LINZ NZ Property Titles for example took 70 seconds to create it. It would be helpful if some progress indication was output during this so better clone progress reporting could be created.

craigds:
I know the ticket's about progress indication, but  if we could speed it up significantly we might not need progress?

part of the slowness might be that the geometries we import don't include envelopes (since cbcb4c8bfef7f215e24769dbb9515324dc557ee2 )

so to build a spatial index we have to build OGR geometries and then ask OGR for an envelope for each feature, which is probably quite slow.

To fix that we could reverse that decision and start adding envelopes to our stored geometries. cc @rcoup

rcoup:
> To fix that we could reverse that decision and start adding envelopes to our stored geometries. cc @rcoup

isn't that in the v2 spec? Haven't decided on a format.

olsen232:
Just to clarify - is this the per-dataset spatial index created in the GPKG working copy?

There's (I think) another one that you can create at sno-repo/`ice_seed.sno-idxd` if you run `sno query index` but it's mostly not used

PR #160, opened 2020-07-09
Add bboxes to stored geometries, except points

craigds:
## Description

Add bboxes to stored geometries, except points

(backward compatible; bboxes aren't *required* for anything)

bboxes are helpful for building spatial indexes quickly:
 * during checkout
 * via `dataset.build_spatial_index()` for querying (experimental)

I didn't add them for points since it would >double the storage required
for them, and they don't add much value for points.

I did this to improve the speed of spatial index creation during checkout (#157) but with the test layer I was using ([NZ Property Titles](https://data.linz.govt.nz/layer/50804-nz-property-titles/), 2M features with fairly simple polygons) I only got about 10% wins.

Wins are likely to be much larger on layers with more complex geometries, so if there is resistance to this being merged I can try some other layers and try and quantify the wins.

## Checklist:


sno/gpkg.py
```diff
@@ -317,25 +294,95 @@ def hex_wkb_to_gpkg_geom(hex_wkb, **kwargs):
 WKB_POINT_EMPTY_LE = b"\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xF8\x7F\x00\x00\x00\x00\x00\x00\xF8\x7F"
 
 
+def _gpkg_envelope_size(flags):
+    envelope_typ = (flags & 0b00001110) >> 1
```
rcoup, line 298: this needs comments — presumably different types are 2/3/4D?

sno/gpkg.py
```diff
@@ -317,25 +294,95 @@ def hex_wkb_to_gpkg_geom(hex_wkb, **kwargs):
...
+    else:
+        raise ValueError("Invalid envelope contents indicator")
+
+
+def _wkb_endianness_and_geometry_type(buf, wkb_offset=0):
+    """
+    Given a buffer containing some WKB at the given offset,
+    returns a two-tuple:
+        * is_little_endian (bool)
+        * geometry type as an integer.
```
rcoup, line 316: OGR Geometry type enum? WKB Geometry Type identifier?

sno/gpkg.py
```diff
@@ -317,25 +294,95 @@ def hex_wkb_to_gpkg_geom(hex_wkb, **kwargs):
...
+
+        # is this a point? if so, we don't *want* an envelope
+        wkb_offset = 8 + envelope_size
+        wkb_is_le, geom_type = _wkb_endianness_and_geometry_type(
+            gpkg_geom, wkb_offset=wkb_offset
+        )
+        want_envelope = geom_type != ogr.wkbPoint
+
+        if wkb_is_le and has_envelope == want_envelope:
+            # everything is fine, no need to roundtrip via OGR
```
rcoup, line 352: what about empty geometry collections (or geometries)? Presumably they don't need envelopes either?

craigds:
> Feels like this is a breaking change — should we roll it into the planned 0.5 geometry encoding changes instead?

I added it to the upgrade script...

> I can't see any as-it-was preservation code, so if we checkout an existing without-envelope geometry into a working copy, will it immediately be marked as a change with the repo?

I hadn't thought of this, but yes that sounds likely. So perhaps it should happen only in dataset2 and workingcopy2.

I think maybe it's better to drop this. It's a minor perf improvement, but not worth the extra complexity, and will need revisiting soon anyway since we plan to switch away from using gpkg geometries internally, presumably for sno v0.5.

Reopen if you disagree.

craigds:
Reopening this.

We discussed future geometry formats, specifically:

- [GSerialized](https://github.com/postgis/postgis/blob/master/liblwgeom/gserialized.txt)
- just little-endian ISO WKB with a mandatory 2D bounding box prepended

We didn't see any major reasons to switch. The main problem we see with GPKG is that too much stuff is optional (endianness/envelopes/etc) which is mostly the case already with WKB.

If we encode all our GPKG geometries with all the optionality removed, we avoid that problem:
* always little-endian
* always little-endian embedded WKB
* always set the is-empty flag correctly
* always have an N-dimensional envelope (for N-dimensional geometries). Except in well-defined useless cases (points/empty geometries don't need envelopes!)
* always set SRS IDs to 0 (this is somewhat controversial, but means v2 datasets can change SRS ID without re-encoding features)

This basically sets us up with what GSerialized has anyway, and means we don't need to bother changing. It also means we can continue to write GPKG working copies trivially, which are likely to be the main WC format for the foreseeable future.

With that in mind, we can reopen this ticket because it gets us most of the way there

sno/gpkg.py
craigds: afaict those are the same thing :) but i'll clarify this is the type extracted directly from the wkb

sno/gpkg.py
craigds: correct, I guess I can exclude them. I don't want the logic for this to get _complicated_ but "empty or point" isn't too bad

rcoup: I think since this is used for hashing we need to get it definitively one way, and it's not going to be trivial to change

rcoup:
Need to update https://github.com/koordinates/sno/pull/224#pullrequestreview-472303962 to match whatever we have here.

sno/gpkg.py
rcoup: From https://github.com/koordinates/sno/pull/224#pullrequestreview-472303962 (but that's me writing down what I remember/made sense, not canonical):

1.  So either we go with "Envelopes must be populated for all non-empty geometries except for `POINT[ZM]` types." ie:

- `POINT(0 0)` → no
- `MULTIPOINT ((0 0))` → yes*
- `MULTIPOINT ((0 0), (1 1))` → yes
- `GEOMETRYCOLLECTION(POINT(0 0))` → yes*
- `GEOMETRYCOLLECTION(MULTIPOINT(0 0))` → yes*
- `GEOMETRYCOLLECTION(POINT(0 0), POINT(1 1))` → yes

where yes* means (bbox top==bottom, left==right)

Or technically "better" but maybe slightly more complex to implement — "Envelopes must be populated for all geometries containing less than two vertices. In practise this means non-empty geometries and single POINT geometries". ie:

- `POINT(0 0)` → no
- `MULTIPOINT ((0 0))` → no
- `MULTIPOINT ((0 0), (1 1))` → yes
- `GEOMETRYCOLLECTION(POINT(0 0))` → no
- `GEOMETRYCOLLECTION(MULTIPOINT(0 0))` → no
- `GEOMETRYCOLLECTION(POINT(0 0), POINT(1 1))` → yes



2. Envelopes must have the same number of dimensions as the geometry (ie: a `POLYGONZM` geometry shall have a 4 dimensional envelope). For a `GEOMETRYCOLLECTION` the envelope shall have a dimensionality equal to the maximum dimensionality of the collected geometries.

- `GEOMETRYCOLLECTION(POINTZM(0 0), LINESTRING(1 1, 2 2))` → 2D
- `GEOMETRYCOLLECTION(POINTZM(0 1 2 3), LINESTRING(1 1, 2 2))` → 4D

craigds: > Or technically "better" but maybe more complex to implement

definitely too complex. The optimisation for points was due to not wanting to double the storage size for a whole entire points layer (which are really _common_) and because it's fast to parse a point and simple to reason about. Niche cases like geometry collections with one point in them are going to break this simplicity.

What we don't want is to have to do serious work to figure out whether a geometry should or should not have an envelope added.

craigds:
slightly different topic: I just realised that [requirement 152](http://www.geopackage.org/spec/#r152) says that empty geometries shall not have an envelope, so I'll definitely implement that.

I think everything else in the above (e.g. multipoints with one point in them) we should just add an envelope.

sno/gpkg.py
craigds: > GEOMETRYCOLLECTION(POINTZM(0 1 2 3), LINESTRING(1 1, 2 2)) → 4D


🤯 I had no idea geoemtrycollection could contain geometries with differing dimensionality.
At present this PR always just adds 2D envelopes. I'll try and make it add 3d/4d ones.

rcoup: > definitely too complex

ok, so all the cases in (1), and that definition into the dataset format documentation.

> I had no idea geoemtrycollection could contain geometries with differing dimensionality.

they can contain anything 😄 

> At present this PR always just adds 2D envelopes. I'll try and make it add 3d/4d ones.

ya, lets not hobble ZM geometries for no real reason.

craigds: > no real reason

It's going to mean a sizable slowdown here: https://github.com/koordinates/sno/blob/geom-bboxes/sno/geometry.py#L124-L142

because we need to parse the WKB of every geometry to determine what size envelope to give it. Including iterating over all subcomponents of collections.

I don't _really_ want to implement wkb parsing myself, so essentially it negates all the shortcuts there, and the whole normalise method becomes a roundtrip via `ogr_to_gpkg_geom(gpkg_geom_to_ogr(g))`.

craigds: I'll merge this PR because it's growing too many limbs, but if we need to tackle 3d/4d envelopes I can do that in a followup.

Issue #159, opened 2020-07-09
datasetV2 checkout is very slow


craigds:
`sno -v checkout` on a largeish v1 dataset (2M polygon features) is much slower than the equivalent v1 dataset.

### 1. `dataset.feature_count()` is very slow in v2.

```
# v1 - initial feature count took 3s
Jul 09 12:00:07 INFO:WorkingCopy_GPKG_1.write_full:Creating features...
Jul 09 12:00:10 INFO:WorkingCopy_GPKG_1.write_full:0.4% 10000/2272386 features... @2.9s (+2.9s, ~17095 F/s)

# v2 - initial feature count took 5 *minutes* (!)
Jul 09 12:03:46 INFO:WorkingCopy_GPKG_1.write_full:Creating features...
Jul 09 12:09:05 INFO:WorkingCopy_GPKG_1.write_full:0.4% 10000/2272347 features... @319.6s (+319.6s, ~156 F/s)
```

### 2. feature iteration during checkout is 3-4x slower in v2:

```
# v1
Jul 09 12:01:58 INFO:WorkingCopy_GPKG_1.write_full:Added 2272386 features to GPKG in 110.5s
Jul 09 12:01:58 INFO:WorkingCopy_GPKG_1.write_full:Overall rate: 20572 features/s

# v2
Jul 09 12:16:12 INFO:WorkingCopy_GPKG_1.write_full:Added 2272347 features to GPKG in 746.9s
Jul 09 12:16:12 INFO:WorkingCopy_GPKG_1.write_full:Overall rate: 3042 features/s
```

note: 746.9s includes the feature count time of 319s. So actual iteration time is 427s, for an *actual* rate of ~5300 features/s. Hence why I said v2 is 3-4x slower.

craigds:
I've done some profiling and found out that `blob.data` is >4x as slow in v2 repos as it is in v1 repos, at least for the NZ property titles repo.

Profiling script: https://gist.github.com/craigds/52574a9413fb70175461241f34cf6f7f

### v2 results:
```
~/layer-50804-v2 (BARE:master) $ /Users/cdestigter/c/sno/profile-blob-data.py . 1000000
         8057895 function calls (3028958 primitive calls) in 138.234 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000  138.234  138.234 <string>:1(<module>)
[...]

PR #163, opened 2020-07-13
Don't create deltas when doing imports

craigds:
## Description

Creating deltas slows down imports by quite a lot.
It may also cause checkouts to perform suboptimally, especially when
iterating over all features (checkouts, large diffs, etc).

In #159 this improved checkout speed by 250%.

This is likely inferior to doing `git gc --aggressive` after
importing, but also *much* faster.

If the caller wants to do an aggressive gc *after* the import to improve
the efficiency of the repo even more, that's their prerogative.

## Related links:

- #159 

## Checklist:

### DIFF
--- a/sno/fast_import.py
+++ b/sno/fast_import.py
@@ -25,6 +25,7 @@ def fast_import_tables(
     message=None,
     limit=None,
     max_pack_size="2G",
+    max_delta_depth=0,
     extra_blobs=(),
 ):
     structure_version = int(structure_version)
@@ -55,6 +56,7 @@ def fast_import_tables(
         "--quiet",
         "--done",
         f"--max-pack-size={max_pack_size}",
+        f"--depth={max_delta_depth}",
     ]
 
     if header is None:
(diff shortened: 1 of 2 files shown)


rcoup:
> This is likely inferior to doing `git gc --aggressive` after importing

Can you quantify this? Would `--depth=1` or some other low number achieve a better balance?

craigds:
### `--depth=0`:
1174656 KB of objects
20563 features/s checkout speed

### `--depth=2:`

1097816 KB of objects (-7%)
18283 features/s checkout speed (-11%)



We could expose this as a (hidden?) CLI flag if you think it's useful for tuning. Otherwise ... 
it's subjective, but feels like disk space is generally cheaper than users time so I prefer depth=0

craigds:
^ thar

rcoup:
> We could expose this as a (hidden?) CLI flag if you think it's useful for tuning.

Nah, we should respect whatever is set (`gc.aggressiveDepth`? `pack.depth`?) via git config mechanisms, and set the default otherwise.

> I prefer depth=0

👍

PR #164, opened 2020-07-13
Fix versioning to be per repo, not per dataset

olsen232:
Some changes had already been made to make versioning per repo - for instance, a .sno-format blob existed already for version 2 - but this finishes up that change, with a variety of small changes:

- Initialising a dataset no longer checks what the `meta/version` file in the dataset contains, but just depends on the `.sno-format` file for the whole repo. The absence of this file means that version=1.
- Using a RepositoryStructure to obtain a DatasetStructure means you automatically get the right one. Even test code should generally use a RepositoryStructure to obtain a DatasetStructure, instead of instantiating one directly
- Dataset V2 no longer contains a `meta/version` file, since this is redundant. Dataset1 continues to contain this file but it is no longer used: instead, we know that version=1 because there is no .sno-format file in v1.
- Importing a dataset automatically creates a .sno-format file if one is needed and if this is the first commit. If this is not the first commit, we make sure the new dataset is compatible with the existing repo version.
- Structure versions are now 1 and 2, not "1.0" or "2.0". There are no minor versions for breaking structural changes.
- Some code moved out of Dataset superclass and into Dataset1 since it is not common to both

## Related links:

https://github.com/koordinates/sno/issues/142

## Checklist:


sno/init.py
```diff
@@ -686,19 +686,19 @@ def iter_features(self):
...
         yield from dbcur
 
-    def build_meta_info(self):
+    def iter_gpkg_meta_items(self):
         """
         Returns metadata from the gpkg_* tables about this GPKG.
         """
         # We use the OGR implementation for everything it knows about.
-        # The `gpkg.get_meta_info()` can produce mostly the same stuff,
+        # The `gpkg.iter_gpkg_meta_items()` can produce mostly the same stuff,
```
craigds, line 694: this is wrong, it's still called `get_meta_info()` in `sno/gpkg.py`

sno/init.py
```diff
@@ -686,19 +686,19 @@ def iter_features(self):
...
         # but it doesn't know about overrides from the `--titles` and `--description` options.
         done_keys = []
-        for k, v in super().build_meta_info():
+        for k, v in super().iter_gpkg_meta_items():
             done_keys.append(k)
             yield k, v
 
-        # For everything not handled by the OGRImporter implementation, get it from gpkg.get_meta_info()
+        # For everything not handled by the OGRImporter implementation, get it from gpkg.iter_gpkg_meta_items()
```
craigds, line 701: same here

PR #165, opened 2020-07-13
Fix merge / conflicts / resolve for V2.

olsen232:
## Description

Moved decode_path into dataset superclass, which was enough to get `sno merge`, `sno conflicts` and `sno resolve` working.

Conversely, moved `cast_primary_key` out of datasets superclass and into v1 - datasets V2 has its own, schema.sanitise_pk_values, and right now these are not part of the public interface.

## Related links:

## Checklist:

### DIFF
--- a/sno/structure.py
+++ b/sno/structure.py
@@ -87,15 +87,16 @@ class RepositoryStructure:
         """Returns the dataset version to use for this entire repo."""
         return get_structure_version(self.repo, self.tree)
 
-    def decode_path(self, path):
+    def decode_path(self, full_path):
         """
-        Given a path in the sno repository - eg "table_A/.sno-table/49/3e/Bg==" -
+        Given a path in the sno repository - eg "path/to/dataset/.sno-table/49/3e/Bg==" -
         returns a tuple in either of the following forms:
-        1. (table, "feature", primary_key)
-        2. (table, "meta", metadata_file_path)
+        1. (dataset_path, "feature", primary_key)
+        2. (dataset_path, "meta", meta_item_path)
         """
-        table, table_path = path.split("/.sno-table/", 1)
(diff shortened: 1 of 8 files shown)


sno/structure.py
```diff
@@ -87,15 +87,15 @@ def version(self):
...
+        Given a path in the sno repository - eg "path/to/dataset/.sno-table/49/3e/Bg==" -
         returns a tuple in either of the following forms:
-        1. (table, "feature", primary_key)
-        2. (table, "meta", metadata_file_path)
+        1. (dataset_path, "feature", primary_key)
+        2. (dataset_path, "meta", meta_item_path)
         """
-        table, table_path = path.split("/.sno-table/", 1)
-        return (table,) + self.get(table).decode_path(table_path)
+        dataset_path, rel_path = re.split("(?=/.sno-table/)", full_path, 1)
```
craigds, line 97: should pre-compile this regex outside the function body. Also it should be a raw string

sno/structure.py
```diff
@@ -275,6 +275,9 @@ def instantiate(cls, tree, path, version):
         if not isinstance(tree, pygit2.Tree):
             raise TypeError(f"Expected Tree object, got {type(tree)}")
 
+        if not ".sno-table" in tree:
```
craigds, line 278:
```suggestion
        if ".sno-table" not in tree:
```
my linter would dislike this immensely :)

sno/structure.py
olsen232: Timed this - precompiling a regex is 25% faster, using simple string split and then re-prepending it is 75% faster. Using simple string operations now. No guarantee that it makes any difference to the code overall, but no harm in it.

sno/structure.py
olsen232: Done

PR #168, opened 2020-07-17
Speed up feature iteration using memoryviews

craigds:
## Description

This is a ~10% speedup for object-iteration commands (eg checkout / show / diff)

`blob.data` is slow, partly because it copies the data into a
bytestring.
But in most cases we throw away the bytestring after passing it through
`msgpack.unpackb()`. msgpack can accept a memoryview object instead,
so just switching to memoryview avoids the copy and thus gives us a
sizeable speedup.

## Related links:

none

## Checklist:

### DIFF
--- a/sno/dataset1.py
+++ b/sno/dataset1.py
@@ -145,12 +145,12 @@ class Dataset1(DatasetStructure):
         feature_path = self.encode_1pk_to_path(pk)
         index.remove(feature_path)
 
-    def repo_feature_to_dict(self, blob_path, blob_data, ogr_geoms=False):
+    def repo_feature_to_dict(self, blob_path, blob_memoryview, ogr_geoms=False):
         feature = {
             self.primary_key: self.decode_path_to_1pk(blob_path),
         }
         bin_feature = msgpack.unpackb(
-            blob_data,
+            blob_memoryview,
             ext_hook=self._msgpack_unpack_ext_ogr
             if ogr_geoms
             else self._msgpack_unpack_ext,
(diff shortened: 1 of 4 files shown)


craigds:
Benchmark results for this:

https://github.com/koordinates/sno/compare/master...craigds:bench-data

```
$ sno benchmark blob-memoryview blob-data
[starting profiler]
[stopped profiler]
         2272407 function calls in 24.265 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  2272403   24.265    0.000   24.265    0.000 /Users/cdestigter/checkout/sno/sno/benchmark.py:90(_get_data)
        1    0.000    0.000    0.000    0.000 /usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py:116(__exit__)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.next}
        1    0.000    0.000    0.000    0.000 /Users/cdestigter/checkout/sno/sno/benchmark.py:39(_profile)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}



[starting profiler]
[stopped profiler]
         2272407 function calls in 33.216 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  2272403   33.216    0.000   33.216    0.000 /Users/cdestigter/checkout/sno/sno/benchmark.py:76(_get_data)
        1    0.000    0.000    0.000    0.000 /usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py:116(__exit__)
        1    0.000    0.000    0.000    0.000 {built-in method builtins.next}
        1    0.000    0.000    0.000    0.000 /Users/cdestigter/checkout/sno/sno/benchmark.py:39(_profile)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

craigds:
verbally approved by @olsen232 👍 

discussed: we might be able to use memoryviews for _all_ blobs, if we
- switch to ujson (it handles memoryviews in `loads`. might be a good thing to do anyway for generally fast jsonning)
- use `str(mymemoryview, encoding='utf-8')` where we need text. memoryview doesn't have a `.decode()` method but str() still works

however, meta blobs are such a small fraction of blobs that it doesn't seem worth doing anything for them right now, so I'll merge this and forget about the rest for the moment.

Issue #170, opened 2020-07-20
Better machine-readable log output


UberMouse:
As is it stands currently we shell out to git-log when calling `sno log`. This means we are limited to the settings `git-log` supports for creating machine-readable log output, namely, `--pretty`. Current log output for machine consumption is generated with the following `git-log` arguments

```
  const prettyFormat = [
    "%H", // SHA
    "%h", // short SHA
    "%s", // summary
    "%b", // body
    // author identity string, matching format of GIT_AUTHOR_IDENT.
    //   author name <author email> <author date>
    // author date format dependent on --date arg, should be raw
    "%an <%ae> %ad",
    "%cn <%ce> %cd",
    "%P", // parent SHAs,
    "%(trailers:unfold,only)",
  ].join(`%x${delimiter}`);
```
`--stat --date=raw -z --no-show-signature --no-color --pretty=${prettyFormat}`.

Currently, everything in `prettyFormat` is parsed out of a commit, except for the [trailers](https://www.git-scm.com/docs/git-interpret-trailers), trailers are currently ignored.

## Problems with `--pretty`

1. It is only aware of git concepts so we can't extract any sno specific information (ie also show me which datasets have been modified in this commit)
2. It's a pretty esoteric format
3. We can't extend it because it's part of git-log

I don't have any thoughts or opinions on _how_ this should be done at this stage, or what format would be better. This is just to collect issues that need to be resolved as part of improving what machine-readable information we can pull out.

## Problems found through usage

1. We want the ability to commits differently depending on their "location", ie whether they are a new commit and have not been pushed, or a commit that has been pushed, or a commit that has been fetched and not merged. Currently, this is handled by performing 3x calls to `sno log` filtering the commits by that criteria and merging the results
2. There's no way to extract this information with `--pretty`. Something workable is to include `--stat` and then parse the stat information but it' pretty awful, delimiters aren't in the right places so the stat output ends up on the wrong line and it's human-readable text

olsen232:
Somewhat addressed by https://github.com/koordinates/sno/pull/177 - try `sno log -o json --dataset-changes`

olsen232:
I'm closing this for now -
- machine readability is handled with -o json, and --dataset-changes
- three simultaneous calls to log is not ideal, but not the worst either. A single call to kart log will suffice if you are willing to parse the parents to split it into three parts. It's tricky (and not git-standard) to find another way to output logs that not only makes parsing easier, but also works well for edge cases like multiple merge-parent candidates.

PR #177, opened 2020-07-24
Add support for json output to `sno log`

olsen232:
## Description

`sno log` now supports JSON output - the JSON produced is very similar to other commands like `sno show` or `sno status`.

## Checklist:

### DIFF
--- a/sno/log.py
+++ b/sno/log.py
@@ -1,0 +1,172 @@
+from datetime import datetime, timezone, timedelta
+import itertools
+import subprocess
+import sys
+
+import click
+
+from .exec import execvp
+from .exceptions import SubprocessError
+from .output_util import dump_json_output
+from .timestamps import datetime_to_iso8601_utc, timedelta_to_iso8601_tz
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True,))
+@click.pass_context
+@click.option(
+    "--output-format",
+    "-o",
+    type=click.Choice(["text", "json"]),
+    default="text",
+    help=(
+        "Output format. 'quiet' disables all output and implies --exit-code.\n"
+        "'html' attempts to open a browser unless writing to stdout ( --output=- )"
+    ),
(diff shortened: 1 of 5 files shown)


sno/log.py
```diff
@@ -0,0 +1,75 @@
...
+
+@click.command(context_settings=dict(ignore_unknown_options=True,))
+@click.pass_context
+@click.option(
+    "--output-format",
+    "-o",
+    type=click.Choice(["text", "json"]),
+    default="text",
+    help=(
+        "Output format. 'quiet' disables all output and implies --exit-code.\n"
```
craigds, line 21: help text is wrong, there's no quiet or html option

sno/log.py
```diff
@@ -0,0 +1,75 @@
...
+def log(ctx, output_format, json_style, args):
+    """ Show commit logs """
+    if output_format == "text":
+        execvp("git", ["git", "-C", ctx.obj.repo.path, "log"] + list(args))
+
+    elif output_format == "json":
+        repo = ctx.obj.repo
+        try:
+            cmd = ["git", "-C", repo.path, "log", "--pretty=format:%H"] + list(args)
+            r = subprocess.run(cmd, encoding="utf8", check=True, capture_output=True)
```
craigds, line 41: Ideally we'd iterate over the output of this while it's still running and dump json output in a generative way rather than capturing all the output before dumping any of it. I think we probably have all the pieces (`json.dump` apparently starts producing output right away.) But doing that would add a lot of complexity, and probably we'll need some really long-lived repos for the size of these logs to become a problem. Maybe add a TODO for that?

UberMouse:
I'm not sure what the plan is with this, if more work is going to be done soon, but as this currently stands I can't make use of it. It doesn't provide me the information I need (#170). Which is fine for now, I've already written the code to parse all the information out of the existing --pretty-format format I'm using.

PR #179, opened 2020-07-29
Streaming of diffs (text output)

olsen232:
## Description

This greatly reduces the time until the first change is shown for large diffs with lots of changes, eg the first import - by not reading the relevant blob from the repo until it is needed to show the change.

Currently only works for text output - for json and geojson, a dict containing the entire diff is generated first before json.dump is called. This can be fixed, will work on it in a follow up PR.

Rename detection in revision<>revision diffs is removed as of this change - it seems that is has never worked anyway. Will also re-add that in a follow up PR.

## Related links:

https://github.com/koordinates/sno/issues/156

## Checklist:

### DIFF
--- a/sno/diff_structs.py
+++ b/sno/diff_structs.py
@@ -6,7 +6,10 @@ class Conflict(Exception):
 
 
 class KeyValue(namedtuple("KeyValue", ("key", "value"))):
-    """A key-value pair. A delta is made of two of these - one old, one new."""
+    """
+    A key-value pair. A delta is made of two of these - one old, one new.
+
+    """
 
     @staticmethod
     def of(obj):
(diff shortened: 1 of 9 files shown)


sno/diff_structs.py
```diff
@@ -17,6 +20,13 @@ def of(obj):
             return KeyValue(*obj)
         raise ValueError(f"Expected (key, value) tuple - got f{type(obj)}")
 
+    def get_lazy_value(self):
+        """Deltas can be created so that the values are only generated when needed."""
```
craigds, line 24: Should this be memoized? If you call it twice will it call get_feature twice?

olsen232: Done

olsen232:
Added a test that checks that the lazy evaluation works at a basic level

PR #185, opened 2020-08-04
Support tables being added / deleted / structurally changed during checkout

olsen232:
## Description

Working copies are now able to follow along when checking out commits that have completely different structure to HEAD, including when: 
- tables that exist at HEAD need to be deleted
- tables that do not exist at HEAD need to be created and fully written
- tables have structural changes that can't be done without rewriting the entire table

The only change to schema that can be done without rewriting the entire table is when a column is renamed. The reasons for this are the following:
- if a new column is inserted, we need to rewrite every row. Simply executing an ALTER TABLE ... ADD COLUMN statement would create the column, but wouldn't populate it - and, it would only work if the column is after every other column.
- if an existing column is deleted, we need to recreate the table. Sqlite doesn't support ALTER TABLE ... DROP COLUMN.

Although there could be optimisations where we only sort-of rewrite the entire table, depending on whether columns are added, dropped, or both, this is the easiest solution for now.

Still TODO: maybe optimise the cases where the only schema changes are new columns at the end of the schema, or, the schema has been structurally changed but there are no new columns.

## Related links:

## Checklist:

### DIFF
--- a/sno/working_copy.py
+++ b/sno/working_copy.py
@@ -334,6 +334,34 @@ class WorkingCopyGPKG(WorkingCopy):
         with self.session() as db:
             yield from gpkg.get_meta_info(db, dataset.name)
 
+    def delete_meta(self, dataset):
+        with self.session() as db:
+            dbcur = db.cursor()
+
+            # FOREIGN KEY constraints are still active, so we delete in a particular order:
+            for table in (
+                "gpkg_metadata_reference",
+                "gpkg_geometry_columns",
+                "gpkg_contents",
+            ):
+                dbcur.execute(
+                    f"""DELETE FROM {table} WHERE table_name = ?;""", (dataset.name,)
+                )
+
+            gpkg_metadata = dataset.get_meta_item("gpkg_metadata") or {}
+            for row in gpkg_metadata:
+                params = dict(row.items())
+                params.pop("id")
+
+                keys, values = zip(*params.items())
+                sql = f"""
+                    DELETE FROM gpkg_metadata WHERE
+                        ({','.join([gpkg.ident(k) for k in keys])})
+                    =
+                        ({','.join(['?']*len(keys))});
+                    """
+                dbcur.execute(sql, values)
+
     def iter_meta_items_v2(self, dataset):
         from .gpkg_adapter import wkt_to_srs_str, gpkg_to_v2_schema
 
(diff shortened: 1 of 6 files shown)


sno/working_copy.py
```diff
@@ -674,6 +732,25 @@ def delete_features(self, dbcur, dataset, pk_iter):
...
+    def drop_table(self, target_tree_or_commit, *datasets):
+        with self.session() as db:
+            dbcur = db.cursor()
+            for dataset in datasets:
+                if dataset.has_geometry:
+                    self._drop_spatial_index(dataset)
+                table = dataset.name
+                dbcur.execute(f"""DROP TABLE {gpkg.ident(table)};""")
+
+        with self.session() as db:
```
craigds, line 744: any reason these with/cursor blocks are separated?

sno/working_copy.py
```diff
@@ -836,19 +913,34 @@ def update_meta_table(self, new_tree):
...
-        for patch in wc_meta_diff:
+        # Sanity check:
+        for patch in meta_diff:
             delta = patch.delta
-            assert delta.old_file.path == delta.new_file.path
+            if delta.old_file and delta.new_file:
+                assert delta.old_file.path == delta.new_file.path
+
+        if src_ds.version < 2:
+            # Dataset1 doesn't support meta changes at all
```
craigds, line 933: comment is kinda incorrect now

rcoup:
I'm wondering whether the user should approve messing with schemas? Thinking if you have FKs to Sno tables or triggers or something else then you might not want changes to be applied accidentally.

rcoup:
Particularly in stricter environments like a Postgres database. And conversely in a flat file environment like Shapefiles you don't care at all.

olsen232:
@rcoup that would have to be a different change - this change means that, given the schema DID change between rev X and rev Y, and the user was at rev X and checks out rev Y - then we update the tables to match the schema at rev Y. If we don't do that, it means we can't properly update the features to be those at rev Y, and nothing will work properly.

But by all means write up a feature request describing what you have in mind - certain repos should be schema-locked and there we don't allow schema changes to be committed until/unless the user unlocks them?

rcoup:
Ah, I meant aborting the checkout and staying at revision X, rather than half-applying Y.

Eg: "Y has a different schema from X, apply changes? [Yn]"

sno/working_copy.py
olsen232: Updated comment.

sno/working_copy.py
olsen232: I thought there was but apparently not. Fixed.

olsen232:
Yeah we can do that too

PR #194, opened 2020-08-07
Use workingcopy V2 when repo is V2

### LINKED ISSUE #63: Rename meta tables in geopackage to `gpkg_*`
When dragging a sno working copy GPKG into QGIS, you're presented with this dialog:

Screen Shot 2020-04-30 at 7 52 39 AM

OGR does similar:

```bash
$ ogrinfo -ro points.gpkg
INFO: Open of `points.gpkg'
      using driver `GPKG' successful.
1: nz_pa_points_topo_150k (Point)
2: .sno-meta (None)
3: .sno-track (None)
```

It'd be best if we could hide those .sno tables. We could just rename them to `gpkg_sno_track` and `gpkg_sno_meta` and OGR/QGIS would ignore them:

```
$ sqlite3 points2.gpkg
SQLite version 3.28.0 2019-04-15 14:49:49
Enter ".help" for usage hints.
sqlite> alter table `.sno-meta` rename to `gpkg_sno_meta`;
sqlite> alter table `.sno-track` rename to `gpkg_sno_track`;


$ ogrinfo -ro points2.gpkg
INFO: Open of `points2.gpkg'
      using driver `GPKG' successful.
1: nz_pa_points_topo_150k (Point)
```

craigds:
## Description

This renames the meta/track tables so they're not presented by OGR and
QGIS as datasets. The new tables are
 * `gpkg_sno_meta`
 * `gpkg_sno_track`

When using V1 repos, working copy is unchanged.

## Related links:

Fixes #63

## Checklist:

### DIFF
--- a/sno/working_copy.py
+++ b/sno/working_copy.py
@@ -16,14 +16,14 @@ from .exceptions import InvalidOperation
 from .filter_util import UNFILTERED
 from .schema import Schema
 from .structure import RepositoryStructure
+from .structure_version import get_structure_version
 
 L = logging.getLogger("sno.working_copy")
 
 
-_DEFAULT_VERSION = 1
-
-
 class WorkingCopy:
+    VALID_VERSIONS = (1, 2)
+
     @classmethod
     def get(cls, repo, create_if_missing=False):
         if create_if_missing:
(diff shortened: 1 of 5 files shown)


rcoup:
All the other GPKG tables are snake case, can we follow that?

sno/working_copy.py
```diff
@@ -230,6 +232,11 @@ def delete(self):
         # https://www.sqlite.org/shortnames.html
         for f in Path(self.full_path).parent.glob(Path(self.path).name + "-*"):
             f.unlink()
+        try:
```
olsen232, line 235: Maybe don't delete this config - see https://github.com/koordinates/sno/issues/192 - this config is now written during `sno init` and so can be present even when the working copy is not. Deleting a working copy shouldn't necessarily clear the config - it depends what is required: should the next `sno checkout` recreate a working copy using the existing config, or if it should create a new working copy using the default config. Check to make sure callers would want the config to be lost, and maybe document, or add an extra arg eg clear_config=False

craigds: I didn't write this, I think this was a bad merge. will remove

craigds:
done

sno/working_copy.py
```diff
@@ -428,11 +438,16 @@ def iter_meta_items_v2(self, dataset):
             f"{gpkg_name}/{dataset.name}",
         ).to_column_dicts()
 
-    def write_full(self, target_tree_or_commit, *datasets, safe=True):
-        raise NotImplementedError()
+    def save_config(self, **kwargs):
```
olsen232, line 441: I think this is a bad merge too?

PR #200, opened 2020-08-11
better schema diffs

OrangeOranges:
## Description

Improves text diff of schema by marking columns which are deleted, inserted or updated, rather than marking the entire schema as deleted/inserted.

## Checklist:


sno/diff_output.py
```diff
@@ -67,20 +68,83 @@ def _out(dataset, diff):
...
+            if key == "schema" and delta.old and delta.new:
+                # Schema is handled below
+                schema_delta = delta
+                continue
             if delta.old:
                 click.secho(prefix_json(delta.old_value, "- "), fg="red", **pecho)
             if delta.new:
                 click.secho(prefix_json(delta.new_value, "+ "), fg="green", **pecho)
 
+        if schema_delta:
```
olsen232, line 86: Can you move the schema-diff specific stuff out of output_text and into a helper function?

olsen232:
(Craig has already had a look at this to try since I helped a bit with the design of it too, to try and avoid having only one perspective)

sno/diff_output.py
olsen232: This is done

PR #204, opened 2020-08-13
Rewrite working_copy.reset to fix issue 191

olsen232:
## Description

A working copy has a dataset at a particular revision which is its "base"  - this is found by looking up the tree in the sno-meta table.
The working copy can differ from its base in two ways.
1) it can have meta or structural changes from its base. We can't track those, so we need to generate a full diff of them.
2) it can have feature changes. To save us diffing 1000s of features, we maintain a tracking table of which features have been edited since the table was last reset.

Before meta changes were supported, the logic for resetting a WC to a given target dataset was as follows:

1. If we are not okay with discarding changes: Make sure nothing is in the tracking table 
2. Reset everything in the tracking table to match the base dataset and delete the tracking records
3. Make all feature changes that are in the diff from base dataset to target dataset. 
4. Either record these as edits in the tracking table, or update the WC's base to be the target dataset.

Now that we support meta changes, there's more to do:

1. If we are not okay with discarding changes: Make sure nothing is in the tracking table 
AND: Make sure there are no meta changes
2. Revert any meta changes so that we match the base dataset
3. Reset everything in the tracking table to match the base dataset and delete the tracking records 
(this is only guaranteed to work now that the schema matches the base dataset)
4. Make all meta changes that are in the diff from base dataset to target dataset
5. Make all feature changes that are in the diff from base dataset to target dataset. 
6. Either record these as edits in the tracking table, or update the WC's base to be the target dataset.

Even so, this doesn't always work - if the schema changes in any non-trivial way, we just have to delete and recreate
the dataset. In that case, we don't yet support tracking the edits we made to the dataset - so 
`sno restore <commit-C>` doesn't work if commit C has a schema that differs from ours in some way other than just column renames. But, you `sno reset <commit-C>` or `sno checkout <commit-C>` will work at least.

Side effects:
This code fixes meta diff generation so it can detect when metadata has changed and apply those changes to the working copy. As a side effect, meta diffs now work everywhere, even for V1 datasets.

## Related links:

https://github.com/koordinates/sno/issues/191

## Checklist:

### DIFF
--- a/sno/working_copy.py
+++ b/sno/working_copy.py
@@ -855,7 +855,7 @@ class WorkingCopyGPKG(WorkingCopy):
         return ds_diff
 
     def can_find_renames(self, meta_diff):
-        """There's no point looking for renames if the schema has changed."""
+        """Can we find a renamed (aka moved) feature? There's no point looking for renames if the schema has changed."""
         if "schema" not in meta_diff:
             return True
 
@@ -863,9 +863,13 @@ class WorkingCopyGPKG(WorkingCopy):
         if not schema_delta.old_value or not schema_delta.new_value:
             return False
 
-        old_col_ids = [c["id"] for c in schema_delta.old_value]
-        new_col_ids = [c["id"] for c in schema_delta.new_value]
-        return old_col_ids == new_col_ids
+        old_schema = Schema.from_column_dicts(schema_delta.old_value)
+        new_schema = Schema.from_column_dicts(schema_delta.new_value)
+        dt = old_schema.diff_type_counts(new_schema)
+        # We could still recognise a renamed feature in the case of type updates (eg int32 -> int64),
+        # but basically any other type of schema modification means there's no point looking for renames.
+        dt.pop("type_updates")
+        return sum(dt.values()) == 0
 
     def find_renames(self, feature_diff, dataset):
         """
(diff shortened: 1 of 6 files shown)


sno/working_copy.py
```diff
@@ -1024,78 +1020,98 @@ def _apply_meta_schema(self, src_ds, dest_ds, src_value, dest_value, dbcur):
...
-                # GIT_DELTA_RENAMED
-                # GIT_DELTA_COPIED
-                # GIT_DELTA_IGNORED
-                # GIT_DELTA_TYPECHANGE
-                # GIT_DELTA_UNMODIFIED
-                # GIT_DELTA_UNREADABLE
-                # GIT_DELTA_UNTRACKED
-                raise NotImplementedError(f"Delta status: {d.status_char()}")
+
+        first_unsupported_delta = next(
```
craigds, line 1063: This change means we're iterating the entire diff 3 times.

I'd prefer to not do that - it could be millions of features...

what about using a single iteration? Maybe use a chunk iterator to batch the deletes/inserts to sqlite every 10000 deltas or something.

And if we encounter an unsupported delta we can rollback the sqlite transaction before raising

sno/working_copy.py
```diff
@@ -1159,117 +1189,105 @@ def reset(
...
             table_inserts,
             table_deletes,
             table_updates,
         )
 
+        structural_changes = table_inserts | table_deletes
+        if track_changes_as_dirty and structural_changes:
+            # We don't yet support tracking changes as dirty if we delete, create, or rewrite an entire table.
+            structural_changes_text = '\n'.join(structural_changes)
+            raise NotNotYetImplemented(
```
craigds, line 1202: Too many negatives

sno/working_copy.py
```diff
@@ -1159,117 +1189,105 @@ def reset(
...
+            if not track_changes_as_dirty:
                 # update the tree id
                 dbcur.execute(
                     f"UPDATE {self.META_TABLE} SET value=? WHERE table_name='*' AND key='tree';",
                     (target_tree.hex,),
                 )
 
+    def _filter_by_paths(self, datasets, paths):
+        """Filters the datasets so that only those matching the paths are returned."""
+        # FIXME - looks like this only keeps things that match *all* paths, instead of *any* paths.
```
craigds, line 1243: yeah this impl looks a bit broken, need to fix

sno/working_copy.py
```diff
@@ -1159,117 +1189,105 @@ def reset(
...
+        if paths:
+            for path in paths:
+                datasets = [
+                    ds for ds in datasets if os.path.commonpath([ds.path, path]) == path
```
craigds, line 1247: is this just the same as 
```suggestion
                    ds for ds in datasets if ds.path.startswith(path)
```

craigds: in fact since startswith takes a tuple, isn't this whole method just 

```python
     def _filter_by_paths(self, datasets, paths): 
         paths = tuple(paths)
         return [ds for ds in datasets if ds.path.startswith(paths)]
```

olsen232: Thanks! Done

sno/working_copy.py
olsen232: Done using your fix

sno/working_copy.py
olsen232: Done

sno/working_copy.py
olsen232: Done - appended all the PKS to lists before resetting them. There doesn't seem to be any real downside to doing it that way - allocating a million element list itself is very cheap compared to generating a million-element diff, which is what we did at the top of this method.

sno/working_copy.py
```diff
@@ -1024,78 +1020,98 @@ def _apply_meta_schema(self, src_ds, dest_ds, src_value, dest_value, dbcur):
...
+        dbcur - DB cursor.
+        """
+        L.debug("Meta diff: %s changes", len(meta_diff))
+        table_name = dataset.name
+        for key in meta_diff:
             try:
-                func = getattr(self, f'_apply_meta_{path.replace("/", "__")}')
+                func = getattr(self, f'_apply_meta_{key.replace("/", "__")}')
+                delta = meta_diff[key]
+                func(table_name, delta.old_value, delta.new_value, dbcur)
```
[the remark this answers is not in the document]
olsen232: Done

PR #206, opened 2020-08-13
Rename the table formerly known as .sno-meta to gpkg_sno_state

olsen232:
A good time to make this breaking change since we already changed it this week, so we can just forget this intermediate change ever happened

### DIFF
--- a/sno/working_copy.py
+++ b/sno/working_copy.py
@@ -108,16 +108,19 @@ class WorkingCopyGPKG(WorkingCopy):
         """ Return a full absolute path to the working copy """
         return (Path(self.repo.path) / self.path).resolve()
 
+    TRACKING_NAME = "track"
+    STATE_NAME = "state"
+
     @property
     def TRACKING_TABLE(self):
-        return self._meta_name("track")
+        return self._sno_table(self.TRACKING_NAME)
 
     @property
-    def META_TABLE(self):
-        return self._meta_name("meta")
+    def STATE_TABLE(self):
+        return self._sno_table(self.STATE_NAME)
 
-    def _meta_name(self, name, suffix=""):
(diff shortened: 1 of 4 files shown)


sno/output_util.py
```diff
@@ -61,7 +61,7 @@ class ExtendedJsonLexer(
         ex_json_lexer = ExtendedJsonLexer()
         # The LexerContext stores the state of the lexer after each call to get_tokens_unprocessed
         lexer_context = pygments.lexer.LexerContext("", 0)
-        formatter = pygments.formatters.TerminalFormatter()
+        formatter = pygments.formatters.Terminal256Formatter(style="colorful")
```
craigds, line 64: unrelated? what's this do? feel free to leave it in if it's useful

sno/working_copy.py
```diff
@@ -286,11 +286,11 @@ def create(self):
 
             dbcur.execute(
                 f"""
-                CREATE TABLE {self.META_TABLE} (
+                CREATE TABLE {self.STATE_TABLE} (
                     table_name TEXT NOT NULL,
                     key TEXT NOT NULL,
                     value TEXT NULL,
-                    CONSTRAINT {self._meta_name('meta', 'pk')} PRIMARY KEY (table_name, key)
+                    CONSTRAINT {self._sno_table('meta', 'pk')} PRIMARY KEY (table_name, key)
```
craigds, line 293: constraint name doesn't really _matter_ but we should keep it consistent with the table going forward, might as well change this to `state` too

sno/output_util.py
olsen232: Ooops was just messing with this last night. It isn't useful

sno/working_copy.py
olsen232: Done

PR #208, opened 2020-08-14
Json highlighting 2

OrangeOranges:
## Description

Changed colours of json syntax highlighting when printing to terminal, to make it more readable.

## Checklist:

        Existing tests pass, and manually checked colours in terminal


olsen232:
So far we don't import anything pygments unless it is used - not sure if it's expensive but we may as well keep this pattern. You can have a lazily created formatter by doing something like:
```
_json_formatter = None

def _get_json_formatter():
  if _json_formatter is None:
    import pygments
    _json_formatter = pygments.formatters.TerminalFormatter(...)
  return _json_formatter
```

PR #213, opened 2020-08-18
diff/show: Add reprojection

craigds:
## Description

diff/show: Add reprojection using the  `--crs` option.

The form of the argument is quite flexible. Equivalent forms:

* `--crs 2193`
* `--crs epsg:2193`
* `--crs EPSG:2193`

* `--crs '+proj=tmerc +lat_0=0 +lon_0=173 +k=0.9996 +x_0=1600000 +y_0=10000000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'`
* `--crs 'PROJCS["NZGD2000 / New Zealand Transverse Mercator 2000",GEOGCS["NZGD2000",DATUM["New_Zealand_Geodetic_Datum_2000",SPHEROID["GRS 1980",6378137,298.257222101,AUTHORITY["EPSG","7019"]],TOWGS84[0,0,0,0,0,0,0],AUTHORITY["EPSG","6167"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4167"]],PROJECTION["Transverse_Mercator"],PARAMETER["latitude_of_origin",0],PARAMETER["central_meridian",173],PARAMETER["scale_factor",0.9996],PARAMETER["false_easting",1600000],PARAMETER["false_northing",10000000],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AUTHORITY["EPSG","2193"]]'`
* `--crs @./nztm.wkt`
* `--crs @./nztm.prj`

I have not added any handling for cases where reprojection fails. The process will error (`RuntimeError` normally) in that case.

## Related links:

none

## Checklist:

### DIFF
--- a/sno/diff.py
+++ b/sno/diff.py
@@ -173,9 +173,14 @@ def diff_with_writer(
                 crs_wkt = dataset.crs_wkt
                 if crs_wkt is not None:
                     src_crs = make_crs(crs_wkt)
-                    dataset_geometry_transforms[
-                        dataset.name
-                    ] = osr.CoordinateTransformation(src_crs, target_crs)
+                    try:
+                        transform = osr.CoordinateTransformation(src_crs, target_crs)
+                    except RuntimeError as e:
+                        raise InvalidOperation(
+                            f"Can't reproject dataset {dataset_path!r} into target CRS: {e}"
+                        )
+
+                    dataset_geometry_transforms[dataset.name] = transform
 
         writer_params = {
             "repo": repo,
@@ -232,8 +237,10 @@ class CoordinateReferenceString(StringFromFile):
 
         try:
             return make_crs(value)
-        except RuntimeError:
-            self.fail(f"Invalid or unknown coordinate reference system: {value}")
+        except RuntimeError as e:
+            self.fail(
+                f"Invalid or unknown coordinate reference system: {value!r} ({e})"
+            )
 
 
 @click.command()
@@ -256,7 +263,7 @@ class CoordinateReferenceString(StringFromFile):
 @click.option(
     "--crs",
     type=CoordinateReferenceString(encoding="utf-8"),
-    help="Reproject geometries into the given coordinate reference system ('EPSG:<code>', proj text or WKT)",
+    help="Reproject geometries into the given coordinate reference system. Accepts: 'EPSG:<code>'; proj text; OGC WKT; OGC URN; PROJJSON.)",
 )
 @click.option(
     "--output",
--- a/sno/diff.py
+++ b/sno/diff.py
@@ -4,7 +4,9 @@ import sys
 from pathlib import Path
 
 import click
+from osgeo import osr
 
+from .cli_util import StringFromFile
 from .diff_output import (  # noqa - used from globals()
     diff_output_text,
     diff_output_json,
@@ -20,6 +22,7 @@ from .exceptions import (
     UNCATEGORIZED_ERROR,
 )
 from .filter_util import build_feature_filter, UNFILTERED
+from .geometry import make_crs
 from .repo_files import RepoState
 from .structure import RepositoryStructure
 
@@ -100,6 +103,7 @@ def diff_with_writer(
     json_style="pretty",
     commit_spec,
     filters,
+    target_crs=None,
 ):
     """
     Calculates the appropriate diff from the arguments,
@@ -114,6 +118,7 @@ def diff_with_writer(
       exit_code:   If True, the process will exit with code 1 if the diff is non-empty.
       commit_spec: The commit-ref or -refs to diff.
       filters:     Limit the diff to certain datasets or features.
+      target_crs:  An osr.SpatialReference object, or None
     """
     from .working_copy import WorkingCopy
 
@@ -161,6 +166,17 @@ def diff_with_writer(
         if feature_filter is not UNFILTERED:
             all_datasets = all_datasets.intersection(feature_filter.keys())
 
+        dataset_geometry_transforms = {}
+        if target_crs is not None:
+            for dataset_path in all_datasets:
+                dataset = base_rs.get(dataset_path) or target_rs.get(dataset_path)
+                crs_wkt = dataset.crs_wkt
+                if crs_wkt is not None:
+                    src_crs = make_crs(crs_wkt)
+                    dataset_geometry_transforms[
+                        dataset.name
+                    ] = osr.CoordinateTransformation(src_crs, target_crs)
+
         writer_params = {
             "repo": repo,
             "base": base_rs,
@@ -168,6 +184,7 @@ def diff_with_writer(
             "output_path": output_path,
             "dataset_count": len(all_datasets),
             "json_style": json_style,
+            "dataset_geometry_transforms": dataset_geometry_transforms,
         }
 
         L.debug(
@@ -209,6 +226,16 @@ def diff_with_writer(
             sys.exit(1)
 
 
+class CoordinateReferenceString(StringFromFile):
+    def convert(self, value, param, ctx):
+        value = super().convert(value, param, ctx)
+
+        try:
+            return make_crs(value)
+        except RuntimeError:
+            self.fail(f"Invalid or unknown coordinate reference system: {value}")
+
+
 @click.command()
 @click.pass_context
 @click.option(
@@ -227,6 +254,11 @@ def diff_with_writer(
     help="Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.",
 )
 @click.option(
+    "--crs",
+    type=CoordinateReferenceString(encoding="utf-8"),
+    help="Reproject geometries into the given coordinate reference system ('EPSG:<code>', proj text or WKT)",
+)
+@click.option(
     "--output",
     "output_path",
     help="Output to a specific file/directory instead of stdout.",
@@ -240,7 +272,9 @@ def diff_with_writer(
 )
 @click.argument("commit_spec", required=False, nargs=1)
 @click.argument("filters", nargs=-1)
-def diff(ctx, output_format, output_path, exit_code, json_style, commit_spec, filters):
+def diff(
+    ctx, output_format, crs, output_path, exit_code, json_style, commit_spec, filters
+):
     """
     Show changes between two commits, or between a commit and the working copy.
 
@@ -270,4 +304,5 @@ def diff(ctx, output_format, output_path, exit_code, json_style, commit_spec, fi
         json_style=json_style,
         commit_spec=commit_spec,
         filters=filters,
+        target_crs=crs,
     )
--- a/sno/diff_output.py
+++ b/sno/diff_output.py
@@ -13,7 +13,7 @@ from sno.schema import Schema
 
 import click
 
-from .geometry import gpkg_geom_to_ogr, gpkg_geom_to_hex_wkb
+from .geometry import gpkg_geom_to_ogr, gpkg_geom_to_hex_wkb, ogr_to_hex_wkb
 from .output_util import dump_json_output, resolve_output_path
 from .utils import ungenerator
 
@@ -318,7 +318,14 @@ def text_row_field(row, key, prefix):
 
 
 @contextlib.contextmanager
-def diff_output_geojson(*, output_path, dataset_count, json_style="pretty", **kwargs):
+def diff_output_geojson(
+    *,
+    output_path,
+    dataset_count,
+    json_style="pretty",
+    dataset_geometry_transforms=None,
+    **kwargs,
+):
     """
     Contextmanager.
 
@@ -375,13 +382,22 @@ def diff_output_geojson(*, output_path, dataset_count, json_style="pretty", **kw
             )
 
         features = []
+        geometry_transform = dataset_geometry_transforms.get(dataset.name)
         for key, delta in sorted(diff.get("feature", {}).items()):
             if delta.old:
                 change_type = "U-" if delta.new else "D"
-                features.append(LazyGeojsonFeatureOutput(change_type, delta.old))
+                features.append(
+                    LazyGeojsonFeatureOutput(
+                        change_type, delta.old, geometry_transform=geometry_transform
+                    )
+                )
             if delta.new:
                 change_type = "U+" if delta.old else "I"
-                features.append(LazyGeojsonFeatureOutput(change_type, delta.new))
+                features.append(
+                    LazyGeojsonFeatureOutput(
+                        change_type, delta.new, geometry_transform=geometry_transform
+                    )
+                )
 
         fc = {"type": "FeatureCollection", "features": features}
         dump_json_output(fc, fp, json_style=json_style)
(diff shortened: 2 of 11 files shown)


tests/test_diff.py
```diff
@@ -263,6 +264,72 @@ def test_diff_points(
...
+            assert features[0]['geometry']['coordinates'] == [
+                1958227.0621957763,
+                5787640.540304652,
+            ]
+            assert features[1]['geometry']['coordinates'] == [
+                1958227.0621957763,
+                5787640.540304652,
+            ]
+
+        print("STDOUT", repr(r.stdout))
```
olsen232, line 305: debug print?

sno/diff.py
```diff
@@ -161,13 +166,28 @@ def diff_with_writer(
...
             all_datasets = all_datasets.intersection(feature_filter.keys())
 
+        dataset_geometry_transforms = {}
+        if target_srs is not None:
+
+            dest_srs = make_srs(target_srs)
+            for dataset_path in all_datasets:
+                dataset = base_rs.get(dataset_path) or target_rs.get(dataset_path)
+                # is there a better way to do this?
+                gc = dataset.get_meta_item("gpkg_geometry_columns")
```
olsen232, line 176: You could add an "srs_id" field to dataset - seems good to have some useful things like that as part of the interface, as opposed to making callers figure out how to extract it from metadata.

There are some examples of these eg "geom_column_name" (defined for both datasets at once in structure.py) and "primary_key" (defined differently for each dataset in dataset1.py and dataset2.py)

rcoup:
Can we use `crs` instead of `srs` (everywhere)? Seems like OGC is standardising on CRS now, no reason we shouldn't as well.

rcoup:
So, these are [the formats `proj_create()` accepts](https://proj.org/apps/projinfo.html#projinfo), I don't think we need to support them all right now, but I don't think we should add anything _not_ in this list:

> 1. a proj-string,
> 2. a WKT string,
> 3. an object code (like “EPSG:4326”, “urn:ogc:def:crs:EPSG::4326”, “urn:ogc:def:coordinateOperation:EPSG::1671”),
> 4. an Object name. e.g “WGS 84”, “WGS 84 / UTM zone 31N”. In that case as uniqueness is not guaranteed, heuristics are applied to determine the appropriate best match.
> 5. a OGC URN combining references for compound coordinate reference systems (e.g “urn:ogc:def:crs,crs:EPSG::2393,crs:EPSG::5717” or custom abbreviated syntax “EPSG:2393+5717”),
> 6. a OGC URN combining references for references for projected or derived CRSs e.g. for Projected 3D CRS “UTM zone 31N / WGS 84 (3D)”: “urn:ogc:def:crs,crs:EPSG::4979,cs:PROJ::ENh,coordinateOperation:EPSG::16031” (added in 6.2)
> 7. a OGC URN combining references for concatenated operations (e.g. “urn:ogc:def:coordinateOperation,coordinateOperation:EPSG::3895,coordinateOperation:EPSG::1618”)
> 8. a PROJJSON string. The jsonschema is at https://proj.org/schemas/v0.2/projjson.schema.json (added in 6.2)
> 9. a compound CRS made from two object names separated with ” + “. e.g. “WGS 84 + EGM96 height” (added in 7.1)

---

> * `--srs 2193`

I'm not keen on this one, c/- the above logic.

> * `--srs EPSG:2193`
> * `--srs EPSG:2193`

Meant to be the same?

> * `--srs '+proj=tmerc +lat_0=0 +lon_0=173 +k=0.9996 +x_0=1600000 +y_0=10000000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'`

Presumably this is any proj-string, so eg: `--srs '+init=epsg:2193'` works too?

Idle not-for-now query: if I add a `+nadgrids=/path/to/some.gsb` presumably that isn't helpful to anyone. I guess this goes away now new proj basically has every grid in it and prefers them.

> `--srs @./nztm.wkt`
> `--srs @./nztm.prj`

Can we make the content of this just parsed like the command-line argument? Ie could contain a proj string, `EPSG:1234`, or anything else we support?

And related to axis-order handling ideas, can we add URN support? `--crs=urn:ogc:def:crs:EPSG::4326`

sno/diff.py
```diff
@@ -161,13 +166,27 @@ def diff_with_writer(
...
+
+            dest_srs = make_srs(target_srs)
+            for dataset_path in all_datasets:
+                dataset = base_rs.get(dataset_path) or target_rs.get(dataset_path)
+                srs_wkt = dataset.srs_wkt
+                if srs_wkt is not None:
+                    src_srs = make_srs(srs_wkt)
+                    dataset_geometry_transforms[
+                        dataset.name
+                    ] = osr.CoordinateTransformation(src_srs, dest_srs)
```
rcoup, line 180: Presumably this can error?

sno/diff.py
```diff
@@ -209,6 +228,16 @@ def diff_with_writer(
...
 
 
+class SpatialReferenceString(StringFromFile):
+    def convert(self, value, param, ctx):
+        value = super().convert(value, param, ctx)
+
+        try:
+            return make_srs(value)
+        except RuntimeError:
+            self.fail(f"Invalid or unknown spatial reference system: {value}")
```
rcoup, line 238: any way to get an error message from proj here?

sno/diff.py
```diff
@@ -226,6 +255,11 @@ def diff_with_writer(
     is_flag=True,
     help="Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.",
 )
+@click.option(
+    "--srs",
+    type=SpatialReferenceString(encoding="utf-8"),
+    help="Reproject geometries into the given spatial reference system ('EPSG:<code>', proj text or WKT)",
```
rcoup, line 261: Help text needs to summarise the things we end up supporting.

sno/diff.py
```diff
@@ -226,6 +255,11 @@ def diff_with_writer(
     is_flag=True,
     help="Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.",
 )
+@click.option(
+    "--srs",
```
rcoup, line 259: Can we DRY the option, since presumably we'll end up using it everywhere?

sno/diff_output.py
```diff
@@ -392,19 +408,23 @@ def _out(dataset, diff):
 class LazyGeojsonFeatureOutput:
     """Wrapper of KeyValue that lazily serialises it as GEOJSON when sent to json.dumps"""
 
-    __slots__ = ("change_type", "key_value")
+    __slots__ = ("change_type", "key_value", "geometry_transform")
```
rcoup, line 411: isn't there one transform per dataset? Do we need to carry it around with every feature?

sno/diff_output.py
```diff
@@ -420,6 +440,9 @@ def geojson_row(row, pk_value, change=None):
         v = row[k]
         if isinstance(v, bytes):
             g = gpkg_geom_to_ogr(v)
+            if geometry_transform is not None:
+                # reproject
+                g.Transform(geometry_transform)
             f["geometry"] = json.loads(g.ExportToJson())
```
rcoup, line 446: We should really fix the GDAL Python bindings to avoid this encode+decode dance 😄  File a feature request? 🤞

sno/geometry.py
```diff
@@ -1,11 +1,34 @@
...
 from osgeo import ogr, osr
 
 
+_epsg_re = re.compile(r'^(?:epsg:)?(\d+)$', re.IGNORECASE)
+
+
+def make_srs(srs_text):
+    """
+    Creates an OGR SpatialReference object from the given string.
+    Accepted input is either WKT or 'EPSG:<id>'
```
rcoup, line 16: Should we use [`SetFromUserInput()`](https://gdal.org/api/ogrspatialref.html?highlight=setfromuserinput#_CPPv4N19OGRSpatialReference16SetFromUserInputEPKc) here?

sno/diff_output.py
```diff
@@ -495,29 +524,47 @@ def _out(dataset, ds_diff):
...
     """
     for k, v in row.items():
         if isinstance(v, bytes):
-            v = gpkg_geom_to_hex_wkb(v)
+            if geometry_transform is None:
+                v = gpkg_geom_to_hex_wkb(v)
+            else:
+                # reproject
+                ogr_geom = gpkg_geom_to_ogr(v)
+                ogr_geom.Transform(geometry_transform)
```
rcoup, line 552: does this raise exceptions correctly?

craigds:
> 
> > * `--srs 2193`
> 
> I'm not keen on this one, c/- the above logic.

ok, will remove

> Presumably this is any proj-string, so eg: `--srs '+init=epsg:2193'` works too?

yes it does

 
> Idle not-for-now query: if I add a `+nadgrids=/path/to/some.gsb` presumably that isn't helpful to anyone. I guess this goes away now new proj basically has every grid in it and prefers them.

yeah, that'll cause issues, but let's not deal with it. User problem

> > `--srs @./nztm.wkt`
> > `--srs @./nztm.prj`
> Can we make the content of this just parsed like the command-line argument? Ie could contain a proj string, `EPSG:1234`, or anything else we support?

it is - I wasn't suggested the filename here is important. It just reads the file and parses it as a string like everything else

> And related to axis-order handling ideas, can we add URN support? `--crs=urn:ogc:def:crs:EPSG::4326`

will see if I can get GDAL to handle them. Otherwise I'm not terribly familiar with them and don't really want to figure out how to parse them correctly

sno/diff_output.py
craigds: Would it necessarily be faster? It would require the bindings to produce a ton of python objects, whereas at present they just handle one string. I'm happy with leaving this as-is until we identify it as some kind of bottleneck

sno/diff.py
craigds: At present only in diff and show, and it's quite tiny. Happy to DRY it if it gets bigger or we start using it in a few more places (where do you have in mind?)

sno/diff.py
craigds: ok, added in the RuntimeError's message. The two I've been able to produce are:

```
Error: Invalid value for '--crs': Invalid or unknown coordinate reference system: 'burgers' (OGR Error: Corrupt data)

Error: Invalid value for '--crs': Invalid or unknown coordinate reference system: 'EPSG:999999' (PROJ: proj_create_from_database: crs not found)
```

sno/diff_output.py
craigds: yes, there's only one of these objects (all features get the same one).

but the geometry isn't turned into an OGR geometry until it's being streamed out as JSON. So we need to pass the transform down into the lazily-serialized object so it can be used after we have an OGR geometry.

Doing it this way is quite elegant because it ensures minimal memory overhead; we don't have to store millions of OGR geometries in memory (we just create one at a time while streaming JSON out)

sno/diff_output.py
craigds:
```points2 (BARE:master) $ sno diff -o json | grep geom
            "geom": "POINT (173.831599213632 -87.1267244418588)",
points2 (BARE:master) $ sno diff -o json --crs epsg:4326 | grep geom
            "geom": "POINT (173.831599213632 -87.1267244418588)",
points2 (BARE:master) $ sno diff -o json --crs epsg:3857 | grep geom
            "geom": "POINT (19350845.108242 -23507975.1660679)",
```

sno/diff.py
craigds: I haven't managed to get it to error, but looking at the GDAL code it looks like it can theoretically raise a RuntimeError. I'll wrap it with an InvalidArgument

sno/diff_output.py
craigds: There doesn't seem to be a way to make this fail. It does eventually produce `inf` values, but even detecting that will be slow on our end and thus not worth it. I can't see a way to make OGR/proj reject these.

sno/diff_output.py
rcoup: sure, I doing the transform at emit-time makes total sense. But why do we need to pass N million pointers to the same geometry_transform object about with each feature?

sno/diff_output.py
rcoup: >  It would require the bindings to produce a ton of python objects

Sure, but the user ends up with those same Python objects _anyway_.

Emitting JSON is relatively trivial, but parsing _anything_ is much less so.

rcoup:
>> And related to axis-order handling ideas, can we add URN support? --crs=urn:ogc:def:crs:EPSG::4326
>
> will see if I can get GDAL to handle them. Otherwise I'm not terribly familiar with them and don't really want to figure out how to parse them correctly


`SetFromUserInput()` handles them, so done?

sno/diff.py
rcoup: spatial indexing, query, export, vtiles, import, load/replace, checkout...

sno/diff_output.py
rcoup: From https://github.com/OSGeo/PROJ/blob/master/test/gie/builtins.gie#L2548-L2549

```
>>> from osgeo import osr, ogr
>>> osr.UseExceptions()
>>> ogr.UseExceptions()
>>> srs = osr.SpatialReference()
>>> srs.SetFromUserInput("+proj=laea +ellps=GRS80")
0
>>> g_ok = ogr.CreateGeometryFromWkt("POINT(2 1)")
>>> g_err = ogr.CreateGeometryFromWkt("POINT(180 0)")
>>> ll = osr.SpatialReference()
>>> ll.SetWellKnownGeogCS('WGS84')
0
>>> tx = osr.CreateCoordinateTransformation(ll, srs)
>>> g_ok.Transform(tx)
0
>>> g_ok.ExportToWkt()
'POINT (111267.664536143 221146.644064019)'
>>> g_err.Transform(tx)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/osgeo/ogr.py", line 7791, in Transform
    return _ogr.Geometry_Transform(self, *args)
RuntimeError: OGR Error: General Error
```

sno/diff_output.py
craigds: the alternative being a global variable? I try to avoid those where it's not absolutely necessary

sno/diff_output.py
craigds: OK. That happens here too. I'll wrap it with InvalidOperation. One awkward thing is that errors here happen halfway through a JSON stream. But there's not much we can do about that apart from processing everything up front, which we're trying hard to avoid.

rcoup: Sure, I think that's somewhat inevitable.

Though, modern proj knows a _lot_ more about coordinate systems, areas of use, etc in order to do the best possible transform...

https://proj.org/faq.html#can-i-transform-from-abc-to-xyz
https://proj.org/operations/operations_computation.html

Wonder if we can do the equivalent of that in GDAL and issue a warning if it's using Ballpark transforms?

```console
$ projinfo --summary -s 'epsg:27200' -t 'epsg:3416'  # 3416 = ETRS89 / Austria Lambert
Candidate operations found: 1
unknown id, Inverse of New Zealand Map Grid + Ballpark geographic offset from NZGD49 to ETRS89 + Austria Lambert, unknown accuracy, World, has ballpark transformation
$ projinfo --summary -s 'epsg:27200' -t 'epsg:3416' --hide-ballpark
Candidate operations found: 0
```

sno/diff_output.py
craigds: okay, have filed this as OSGeo/gdal#2877

sno/diff.py
craigds: can DRY it when we do some of those.

sno/diff_output.py
rcoup: Create an output context containing that?

I'm not familiar enough with how this is working though, I'd have to dig in tomorrow — eg: I thought we were generating/producing a feature at a time, but that doesn't seem to be the case, we're building up big features lists.

craigds: sort of - we're building up lists of dicts, but only transforming & converting during `json.dump()`. It's a lot less work up-front than it used to be.

PR #224, opened 2020-08-21
Add documentation for datasets v2 format

olsen232:
## Description

Add documentation for datasets v2 format - based on internal Google Doc etc

## Related links:

Fixes https://github.com/koordinates/sno/issues/201

### DIFF
--- a/DATASETS_v2.md
+++ b/DATASETS_v2.md
@@ -1,0 +1,262 @@
+Datasets V2
+-----------
+
+### Background 
+
+Sno 0.2 introduced Datasets V1, and Sno 0.5 introduced Datasets V2.
+Datasets V1 and V2 are storage formats for database tables where each table row is stored in a separate file.
+This means table rows can be stored using git-style version control, resulting in a storage format for a database-table which has version history.
+
+Datasets V2 is very similar to Datasets V1 - the main difference is that the schema of a Datasets V2 table can be changed in isolation without having
+to rewrite every row in the table. Rows that were written with a previous schema are adapted to fit the current schema when read.
+
+In Sno 0.4, only Datasets V1 is supported. In Sno 0.5, both Datasets V1 and Datasets V2 are supported, but a particular Sno repository must be entirely one or the other. When creating a new Sno repository, this can be selected by specifying either `sno init --repo-version=1` or 
+
+The following is a technical description of Datasets V2.
+
+### Overall structure
+
+A V2 dataset is a folder named `.sno-dataset` that contains two folders. These are `meta` which contains information about the database table itself - its title, description, and schema - and 'feature' which contains the table rows. The schema file contains the structure of the table rows, ie the names and type of each column. The "name" of the V2 Dataset is the path to the `.sno-dataset` folder. 
+
+For example, here is the basic folder structure of a dataset named contours/500m:
+
+```
+contours/
+contours/500m/
+contours/500m/.sno-dataset/
+contours/500m/.sno-dataset/meta/
+contours/500m/.sno-dataset/meta/title              # Title of the dataset
+contours/500m/.sno-dataset/meta/description        # Description of the dataset
+contours/500m/.sno-dataset/meta/schema.json        # Schema of the dataset 
+contours/500m/.sno-dataset/meta/...                # Other dataset metadata
+
+contours/500m/.sno-dataset/feature/...             # Database table rows
+```
+
+### Meta items
+
+The following items are stored in the meta part of the dataset, and have the following structure.
+
+#### `meta/title` 
+Contains the title of the dataset, encoded using UTF-8. The title is freeform text, clients could if they desired include HTML tags or markdown in the title.
+
+#### `meta/description`
+Contains the title of the dataset, encoded using UTF-8. The description is freeform text, clients could if they desired include HTML tags or markdown in the title.
+
+#### `meta/schema.json`
+Contains the current schema of the table, as a JSON array. Each item in the array represents a column in the table, and contains the name and type of that database column. A more in-depth explanation is found below under the heading "Syntax".
+
+For example, here is the schema for a dataset containing building outlines:
+```json
+[
+  {
+    "id": "500f5ecc-c02c-e8db-052a-84efb4d08a00",
+    "name": "fid",
+    "dataType": "integer",
+    "primaryKeyIndex": 0,
+    "size": 64
+  },
+  {
+    "id": "b8ae8ff3-1691-7f6f-bbe0-c03c648b6d67",
+    "name": "geom",
+    "dataType": "geometry",
+    "primaryKeyIndex": null,
+    "geometryType": "MULTIPOLYGON",
+    "geometryCRS": "EPSG:4326"
+  },
+  {
+    "id": "4324e825-ffed-8aff-ceac-53a7dde7e2f7",
+    "name": "building_id",
+    "dataType": "integer",
+    "primaryKeyIndex": null,
+    "size": 32
+  },
+  {
+    "id": "878f8e4e-433a-b7bb-74d5-b360ccfb3607",
+    "name": "name",
+    "dataType": "text",
+    "primaryKeyIndex": null,
+    "length": 250
+  },
+  {
+    "id": "c8e75111-0506-a898-4d0e-ed1aa8c81280",
+    "name": "last_modified",
+    "dataType": "date",
+    "primaryKeyIndex": null
+  }
+]
+```
+
+##### Syntax
+Every JSON object in the array represents a column of the database table, and these objects are listed in the same order as the columns in the table. Each of these objects has at least following fields:
+
+###### `id`
+This is a unique ID used internally, the contents of the ID have no specific meaning. However, the ID of a column remains constant over its lifetime, even as its name or position in the array changes, so they can are used to recognise a column even if it has been renamed and moved.
+
+###### `name`
+This is the name of the column in the database table, as would be used in a SELECT statement. Column names must be unique within a dataset.
+
+###### `dataType`
+This is the type of data which is stored in this column. A complete list of allowed types is found in the "Data types" section below.
+
+###### `primaryKeyIndex`
+This controls whether or not this column is a primary key. If this value is `null`, then the column is not a primary key. If it is any non-negative integer, then the column is a primary key. The first primary key column should have a `primaryKeyIndex` of `0`, the next primary key column should have `1`, and so on.
+
+Those are all of the required fields. Certain dataTypes can have extra fields that help specify the type of data that the column should hold - see the "Extra type info" section below.
+
+##### Data types
+
+The following data types are supported by sno. When a versioned sno dataset is converted to a database table (ie, when `sno checkout` updates the working copy) then these sno data types will be converted to equivalent data types in the database table, depending on what is supported by the database in question.
+
+* `boolean`
+  - stores `true` or `false`.
+* `blob`
+  - stores a string of bytes.
+* `date`
+  - stores year + month + day. The timezone that should be used interpret this is not stored.
+* `datetime`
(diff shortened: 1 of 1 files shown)


DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+Another legend entry is added whenever an update to the schema is committed, and they are never modified or deleted. They are part of the internal structure of the dataset and they need not be viewed by the end user.
+
+Each legend has a unique filename based on its hash. Legends are not stored using JSON, but an binary encoding called [msgpack](https://msgpack.org/index.html) that has equivalent capabilities.
+
+#### `meta/crs/{identifier}.wkt`
+A dataset may contain well-known-text CRS definitions of any of the CRS that are needed to interpret its geometry. The identifier that is part of the filename here should be the same as the `geometryCRS` identifier in the schema.
+
+#### `.sno-table/meta/metadata/dataset.json`
```
craigds, line 188: .sno-dataset/...

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+### Features
+
+Every database table row is stored in its own file. It is stored as an array of values plus the name of the legend that should be used to read it. This array is serialised using messagepack (a binary format equivalent to JSON), but for the sake of readability, the example below is shown in JSON.
+
+For instance, a single feature might be stored as the following:
+```json
+[
+  "204b9886d5dbd9fe3a7edb9a7a7dba699b5202f7",  # legend
```
craigds, line 207: can't have comments in json, so most formatters will object to this. IMO you don't need to have a comment in the json blob itself

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+  "500f5ecc-c02c-e8db-052a-84efb4d08a00": 1445288,
+  "b8ae8ff3-1691-7f6f-bbe0-c03c648b6d67": 
+    "GP0001e61000000101cce1b0dce@7fx8f4Dc0",
+  "4324e825-ffed-8aff-ceac-53a7dde7e2f7": 1260047, 
+  "878f8e4e-433a-b7bb-74d5-b360ccfb3607": "Pukerua Bay Police Station",
+  "c8e75111-0506-a898-4d0e-ed1aa8c81280": "2018-11-05"
+}
+```
+
+Finally, the current schema is consulted to find out what the current position and name of the columns with those IDs, so that a database row can be constructed. There is the possibility that some of those columns are no longer part of the schema, in which case these values will be dropped from the feature. There is also the possibility that new columns have been added to the schema since this feature was written, in which cases extra `NULL` values will be inserted into the feature in the appropriate position. The end result will be a feature that conforms to the current database schema - something like the following:
```
craigds, line 231:
```suggestion
Finally, the current schema is consulted to find out the current position and name of the columns with those IDs, so that a database row can be constructed. There is the possibility that some of those columns are no longer part of the schema, in which case these values will be dropped from the feature. There is also the possibility that new columns have been added to the schema since this feature was written, in which case extra `NULL` values will be inserted into the feature in the appropriate position. The end result will be a feature that conforms to the current database schema - something like the following:
```

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+{
+  "http://www.isotc211.org/2005/gmd": {
+    "text/xml": "<gmd:MD_Metadata …"
+  }
+}
+```
+
+### Features
+
+Every database table row is stored in its own file. It is stored as an array of values plus the name of the legend that should be used to read it. This array is serialised using messagepack (a binary format equivalent to JSON), but for the sake of readability, the example below is shown in JSON.
```
craigds, line 202:
```suggestion
Every database table row is stored in its own file. It is stored as an array of values plus the name of the legend that should be used to read it. This array is serialised using [messagepack](https://msgpack.org/index.html) (a binary format equivalent to JSON), but for the sake of readability, the example below is shown in JSON.
```

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+  "c8e75111-0506-a898-4d0e-ed1aa8c81280",
+]
+```
+
+Another legend entry is added whenever an update to the schema is committed, and they are never modified or deleted. They are part of the internal structure of the dataset and they need not be viewed by the end user.
+
+Each legend has a unique filename based on its hash. Legends are not stored using JSON, but an binary encoding called [msgpack](https://msgpack.org/index.html) that has equivalent capabilities.
+
+#### `meta/crs/{identifier}.wkt`
+A dataset may contain well-known-text CRS definitions of any of the CRS that are needed to interpret its geometry. The identifier that is part of the filename here should be the same as the `geometryCRS` identifier in the schema.
```
craigds, line 186:
```suggestion
A dataset may contain well-known-text CRS definitions for any CRS needed to interpret its geometry. The identifier that is part of the filename here should be the same as the `geometryCRS` identifier in the schema.
```

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+
+###### Extra type info for `"dataType": "text"`
+- `length`
+  * Eg: `"length": 100`
+  * The maximum length of the text in characters.
+  * Can be null if the maximum length is unbounded.
+
+###### Extra type info for `"dataType": "numeric"`
+- `precision`
+  * The maximum number of digits for the numeric type. For
```
craigds, line 155:
```suggestion
  * The total number of total digits for the numeric type.
```

DATASETS_v2.md
olsen232: Done

DATASETS_v2.md
olsen232: Done

DATASETS_v2.md
olsen232: Done

DATASETS_v2.md
olsen232: Done

DATASETS_v2.md
olsen232: Done

DATASETS_v2.md
olsen232: Done

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+Contains the title of the dataset, encoded using UTF-8. The description is freeform text, clients could if they desired include HTML tags or markdown in the title.
+
+#### `meta/schema.json`
+Contains the current schema of the table, as a JSON array. Each item in the array represents a column in the table, and contains the name and type of that database column. A more in-depth explanation is found below under the heading "Syntax".
+
+For example, here is the schema for a dataset containing building outlines:
+```json
+[
+  {
+    "id": "500f5ecc-c02c-e8db-052a-84efb4d08a00",
```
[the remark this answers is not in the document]
olsen232: No... in python I keep construct these in column order (ie, first column first), which python3 preserves, so that if I output them again they remain in column order for readability reasons. 

Here in the documentation I could sort or shuffle them without it making the documentation incorrect, but, seems best to keep them in column order still since they *are* in that order in the code and in the database row - putting them in some other order won't make it clearer.

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+
+###### Extra type info for `"dataType": "numeric"`
+- `precision`
+  * The maximum number of digits for the numeric type. For 
+- `scale`
+  * How many of the digits are to the right of the decimal point.
+For example, the number "1234.5678" can be stored in a numeric type with a precision of 8 and a scale of 4.
+
+###### Extra type info for `"dataType": "timestamp"`
+- `timeZone`
```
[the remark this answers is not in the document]
olsen232: Not really sure: perhaps there's a bug in your schema.json gist, which has timezone as metadata you can add to an entire column:
https://gist.github.com/rcoup/2774c12af9f3997088dd13a8a84eff98#file-schema-js

Perhaps that's correct, or perhaps the metadata we need is not the timezone, but whether the column's complete type is `TIMESTAMP WITH TIMEZONE` or `TIMESTAMP WITHOUT TIMEZONE`.

GPKG doesn't have timestamps and no extra work has been done on them so this spec is actually only aspirational at this point. I've just added bug tracking this: https://github.com/koordinates/sno/issues/225
So, I've just taken the extra type info section out for timestamp. I've left timestamp in the list of types, since we will support it, but it's actually not supported as of 0.5

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+  "b8ae8ff3-1691-7f6f-bbe0-c03c648b6d67",
+  "4324e825-ffed-8aff-ceac-53a7dde7e2f7",
+  "878f8e4e-433a-b7bb-74d5-b360ccfb3607",
+  "c8e75111-0506-a898-4d0e-ed1aa8c81280",
+]
+```
+
+Another legend entry is added whenever an update to the schema is committed, and they are never modified or deleted. They are part of the internal structure of the dataset and they need not be viewed by the end user.
+
+Each legend has a unique filename based on its hash. Legends are not stored using JSON, but an binary encoding called [msgpack](https://msgpack.org/index.html) that has equivalent capabilities.
```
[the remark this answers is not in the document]
olsen232: Done

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+    1445288,
+    "GP0001e61000000101cce1b0dce@7fx8f4Dc0",
+    1260047, 
+    "Pukerua Bay Police Station",
+    "2018-11-05"
+  ]
+]
+```
+
+Decoding a feature works as follows - first, look up the legend with the given name from the `meta/legend/` directory. This will contain a list of column IDs. There will be the same number of column IDs as values, so they can be zipped together into key-value pairs:
```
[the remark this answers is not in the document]
olsen232: Done

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+
+```json
+{
+  "500f5ecc-c02c-e8db-052a-84efb4d08a00": 1445288,
+  "b8ae8ff3-1691-7f6f-bbe0-c03c648b6d67": 
+    "GP0001e61000000101cce1b0dce@7fx8f4Dc0",
```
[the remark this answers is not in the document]
olsen232: Done

DATASETS_v2.md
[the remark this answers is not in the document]
olsen232: Done

DATASETS_v2.md
```diff
@@ -0,0 +1,244 @@
...
+* interval
+* numeric
+* text
+* time
+* timestamp
+
+##### Extra type info
+
+Certain types have extra fields that help specify how the type should be stored in a database. They don't necessarily affect how sno stores the data - and they don't necessarily affect all database types - for instance, setting a maximum length of 10 characters in a column with `"dataType": "text"` won't be enforced in a SQLite since it doesn't enforce maximum lengths, and if a user then modifies a field in this column to contain an overlength string, this can still be committed to sno.
+
```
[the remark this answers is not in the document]
olsen232: Done - by adding descriptions to every type - although I realise now that timestamps are not really supported, so it's a little bit open-ended what we think the difference is. Went with a pretty standard SQL kind of distinction.

Added a tracking bug: https://github.com/koordinates/sno/issues/225

olsen232:
Some of the things you wrote in the geometry encoding format section are not true, or not yet true - 

- we don't use any MessagePack extensions cos we don't really need them for anything
- the envelopes change isn't in yet. And, it's not tracked as a blocker for 0.5, but maybe it should be:
https://github.com/koordinates/sno/pull/160
So right now, there is no particular requirement for storing envelopes or not in datasets V2.

I think I need to hold off on making a 0.5 release for another day until we settle whether these are blockers for 0.5, since they are breaking changes. For now I'll document how things are currently, and merge this PR, bringing HEAD up to an untagged "possible v0.5.0 unless we need to do a bit more".

PR #227, opened 2020-08-26
Make upgrade work easily and automatically

olsen232:
## Description

Until now, upgrading a repo looked like this:
`sno upgrade 00-02 [source] [dest] [layer]`
`sno upgrade 02-05 [source] [dest]`

To upgrade from v0 to v2, you would need to run both, and you could only upgrade one layer.

This fixes it so that upgrade automatically upgrades 
- from: whatever the current version is
- to: the latest version
- all layers

so both those commands would just become:
`sno upgrade [source] [dest]`

## Related links:

https://github.com/koordinates/sno/issues/169

## Checklist:


sno/upgrade/__init__.py
```diff
@@ -1,15 +1,39 @@
...
+    source_repo = pygit2.Repository(str(source))
+    if not source_repo or not source_repo.is_bare:
+        raise click.BadParameter(
+            f"'{source}': not an existing repository", param_hint="SOURCE"
+        )
+
+    version = get_repo_version(source_repo)
+
+    if version == 0:
+        upgrade_v0.upgrade_v0(ctx, source, dest)
```
craigds, line 35: This upgrades it from v0 to v1, right? Shouldn't it then call upgrade_v1 to go the rest of the way?

sno/upgrade/upgrade_v0.py
```diff
@@ -0,0 +1,265 @@
...
+    source_repo = pygit2.Repository(str(source))
+    if not source_repo or not source_repo.is_bare:
+        raise click.BadParameter(
+            f"'{source}': not an existing repository", param_hint="SOURCE"
+        )
+
+    version = get_repo_version(source_repo)
+    if version != 0:
+        raise InvalidOperation(
+            "source repo is not at sno version 0.2 - (dataset version 1)"
```
craigds, line 40: wrong version in message

sno/upgrade/upgrade_v0.py
```diff
@@ -0,0 +1,265 @@
...
+    dest = Path(dest)
+
+    if dest.exists():
+        raise click.BadParameter(f"'{dest}': already exists", param_hint="DEST")
+
+    source_repo = pygit2.Repository(str(source))
+    if not source_repo or not source_repo.is_bare:
+        raise click.BadParameter(
+            f"'{source}': not an existing repository", param_hint="SOURCE"
+        )
```
craigds, line 35: This boilerplate can all go in the main `upgrade` function; these versioned scripts can just take pre-validated Path objects

craigds (changes requested):

sno/upgrade/__init__.py
olsen232: Nup I changed it to upgrade straight to V2.

sno/upgrade/upgrade_v0.py
olsen232: Good call. There's a lot of duplicate code, rather get rid of it now than later

sno/upgrade/upgrade_v0.py
olsen232: Fixed - no longer need this check anyway now that this part of the code is the same for both versions.

olsen232:
Please Take Another Look

craigds (approved):

PR #228, opened 2020-08-28
commit/apply: Faster large commits

craigds:
## Description

We were using Index objects to create trees from diffs.

Simply, we'd:
 - create an Index
 - read the full tree of the repo into it
 - call .remove() and .add() a bunch of times to apply a diff
 - write the index back into a tree
 - create a commit object with the new tree.

Unfortunately it turns out this performs poorly at scale, because
.remove() and .add() spend an awful lot of time doing memmove()
operations.
It turns out that indexes seem to be basically sorted vectors, so
inserting and removing is slow.

This change replaces indexes with TreeBuilder objects.
They have a somewhat lower-level API. each TreeBuilder represents
a single directory in a tree. So to modify a large swath of subtrees,
you have to use treebuilders to create each tree, and then add each
tree to it's parent tree.

This requires quite a bit of recursive support code so we can write
features and meta blobs. The result is an 80-90% reduction in time
to apply large patches (10000+ small features. I haven't tested
large features) and the same will be true of regular large commits.
## Related links:

none

## Checklist:

- [ ] Have you included test(s)? (no, but this is a refactor and covered by a lot of existing tests)

### DIFF
--- a/sno/structure.py
+++ b/sno/structure.py
@@ -20,6 +20,7 @@ from .geometry import geom_envelope
 from .schema import Schema
 from .serialise_util import ensure_bytes, json_pack
 from .repository_version import get_repo_version
+from .tree_util import replace_subtree
 
 
 L = logging.getLogger("sno.structure")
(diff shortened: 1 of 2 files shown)


sno/structure.py
```diff
@@ -602,11 +599,152 @@ def diff_meta(self, other, reverse=False):
         meta_new = dict(new.iter_meta_items()) if new else {}
         return DeltaDiff.diff_dicts(meta_old, meta_new)
 
-    def write_index(self, dataset_diff, index, repo):
+    def _replace_subtree(self, repo, root_tree, subpath, subtree_or_blob):
```
olsen232, line 602: Move at least _replace_subtree into a new file:
tree_util.py

_recursive_build_feature_tree is a bit of a hybrid - it has some code that is feature-aware, but also some code for dealing with trees. Ideally we would split this into two halves, the feature part that can stay here and the tree part that would also go to tree_util, and we would figure out what sort of diff to send to tree_util to give it the info it needs. Something like a DeltaDiff, but lower-level - I guess something more like a pygit2.Diff although that probably is a horrible type with horrible performance.

For now you can just move _replace_subtree.

olsen232 (approved):

PR #229, opened 2020-08-31
import: Add `--replace-existing` option

### LINKED ISSUE #112: import: Add a way to replace a full existing dataset
Many agencies publish data in regular releases without tracking changes. It'd be helpful if these datasets could be easily imported into a sno repository, replacing what data is already there.

I suggest a `--replace-existing` option:

```bash
$ sno import --replace-existing --table=mytable src.gpkg
```

For datasets without PKs, this would delete an object and create a new object for each row in the dataset. Those commits will be unreadable, but at least the contents have been tracked.

For datasets with PKs or at least FIDs exposed via OGR, we could trivially implement some basic change detection to avoid doing that full delete & recreate. This would enable third parties to easily track changes in releases of data imported from legacy sources.

For datasets without PKs, it's possible that git will de-dupe the blobs for us? If so...  💯

craigds:
## Description

Adds `--replace-existing` option to the `import` command.

Replaces existing dataset(s) with the same name.

All meta and features are replaced. Git takes care of deduplicating the feature objects, so this can be used as a way of adding versioning to third-party data (as long as it was produced with a stable primary key field)

## Related links:

Fixes #112 

## Checklist:

### DIFF
--- a/sno/fast_import.py
+++ b/sno/fast_import.py
@@ -1,6 +1,7 @@
 import logging
 import subprocess
 import time
+from enum import Enum, auto
 
 import click
 import pygit2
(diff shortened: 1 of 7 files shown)


sno/fast_import.py
```diff
@@ -22,10 +22,10 @@ def fast_import_tables(
     quiet=False,
     header=None,
     message=None,
+    replace_existing=False,
```
olsen232, line 25: Let's merge "replace_existing" and "incremental" into a single three-state flag - something like:
start_from
replace_existing= one of:
REPLACE_NONE,
REPLACE_GIVEN,
REPLACE_ALL,

olsen232 (approved):

PR #230, opened 2020-08-31
Make `--replace-existing` efficient with minor schema changes

craigds:
## Description
`sno import --replace-existing`  (#229) currently rewrites all features if
any schema change occurs, because it passes all features directly to
fast-import.

However, Datasets V2 theoretically supports many schema changes without
rewriting features:
 * re-named columns
 * re-ordered columns
 * dropped columns

This commit adds some checks to the importer to re-use existing feature
blobs in those cases.

This will slow down imports if those schema changes are present,
because the importer has to perform additional checks on each feature
to compare them against the corresponding feature in the existing HEAD.

However, it should cause sizable reductions in unnecessary repo
bloat over time.

The importer now deserialises the existing feature from the repo, and
compares it against the new feature. If the results are the same, the
old feature blob is re-used directly, and no new blob is created.

## Related links:

#229 

## Checklist:

### DIFF
--- a/sno/fast_import.py
+++ b/sno/fast_import.py
@@ -6,9 +6,10 @@ from enum import Enum, auto
 import click
 import pygit2
 
+from . import core
 from .exceptions import SubprocessError, InvalidOperation
 from .import_source import ImportSource
-from .structure import DatasetStructure
+from .structure import DatasetStructure, RepositoryStructure
 from .repository_version import get_repo_version, extra_blobs_for_version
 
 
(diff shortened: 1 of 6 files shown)


sno/fast_import.py
```diff
@@ -104,8 +105,57 @@ def fast_import_tables(
...
+                    replacing_dataset = RepositoryStructure(repo)[source.dest_path]
+                except KeyError:
+                    pass
+                else:
+                    # If the schemas are *identical, there's no need to pass this dataset object on.
+                    # Features won't be rewritten anyway.
+                    if replacing_dataset.schema == source.schema:
+                        replacing_dataset = None
+                    else:
+                        # Check that the schemas are compatible.
```
olsen232, line 127: Look at Schema.diff_type_counts - that has some code that categorises schema changes.

You could use it here, or you could add another function to Schema that does the check that you care about.

sno/fast_import.py
```diff
@@ -104,8 +105,57 @@ def fast_import_tables(
                 raise ValueError(f"{blob_path} already exists")
 
         for source in sources:
+            replacing_dataset = None
             if replace_existing == ReplaceExisting.GIVEN:
                 p.stdin.write(f"D {source.dest_path}\n".encode('utf8'))
+
+                # Optimisation: Try to avoid rewriting features for compatible schema changes.
```
olsen232, line 112: Move some or all of this optimisation into its own function

sno/fast_import.py
```diff
@@ -104,8 +105,57 @@ def fast_import_tables(
...
+                        new_types = {col.id: col.data_type for col in source.schema}
+                        old_types = {
+                            col.id: col.data_type
+                            for col in replacing_dataset.schema
+                            if col.id in new_col_ids
+                        }
+                        if new_col_ids - old_col_ids or new_types != old_types:
+                            # a column has been added, or a column's data_type has changed.
+                            # In this case all rows will be rewritten; can't avoid that.
+                            replacing_dataset = None
```
olsen232, line 139: There is actually only one schema change that guarantees that we need to rewrite all the data - changes to which column is the primary key. These probably don't work so great right now anyway, but you can test for them with Schema.is_pk_compatible.
- major type changes are supposed to be impossible - you can't just change a column from geometry to int
- minor type changes don't mean we need to rewrite values - if someone changes a column from int32 to int64, we don't need to rewrite all the ints. These minor types are just to help us generate working copies of the appropriate type
- column adds don't necessarily mean that we need to rewrite all the data either. Rewrites are only required for those rows which have non-null values in the new column - if a huge dataset with many rows has a new column added, it is highly likely that the user won't immediately have any data for the new column for the majority of existing rows, only for a few new rows.

rcoup:
1. Are column ordering changes handled correctly? eg: `[colA, colB, colC, colD]` then `[colB, colA, colC, colD]` should be a different schema, the same legend, and no feature blob changes.

2. Column adds: any features with NULL values in new columns are free from rewriting blobs, is that the case here?

3. We can't detect column renames via this approach? We should document that renaming columns will result in (unfixable) repo bloat.

craigds:
1: yes - the test checks this.

2: not yet handled but yes okay, I'll add support for that.

3: this already deals with renames, but I'll add a test.

sno/fast_import.py
craigds: ok so basically as long as the PK fields don't change, I'll use `replacing_dataset`

olsen232 (approved):

PR #233, opened 2020-09-01
Better support for custom CRS

olsen232:
## Description

Fix a few loose ends to do with CRS import / working copy generation, so that custom CRS can imported and roundtripped properly.

- don't call SpatialReference.AutoIdentifyEPSG() - this can change the CRS definition.
- in fact, don't roundtrip through SpatialReference(wkt).ExportToWkt() at all, if it can be helped - this can change the CRS definition too.

To avoid this second one, we now pull the CRS definition straight from GPKG for a GPKGImportSource, not through OGR.

## Related links:

Fixes https://github.com/koordinates/sno/issues/148

## Checklist:

### DIFF
--- a/sno/crs_util.py
+++ b/sno/crs_util.py
@@ -1,14 +1,34 @@
 from osgeo.osr import SpatialReference
 
 
-def get_identifier(crs):
+def get_identifier_str(crs):
     """
-    Given a CRS, generate a unique idenfier for it. Eg: "EPSG:2193"
+    Given a CRS, generate a stable, unique identifier for it of type 'str'. Eg: "EPSG:2193"
     """
     if isinstance(crs, str):
         crs = SpatialReference(crs)
-    if isinstance(crs, SpatialReference):
-        auth_name = crs.GetAuthorityName(None)
-        auth_code = crs.GetAuthorityCode(None)
+    if not isinstance(crs, SpatialReference):
+        raise RuntimeError(f"Unrecognised CRS: {crs}")
+    auth_name = crs.GetAuthorityName(None)
+    auth_code = crs.GetAuthorityCode(None)
+    if auth_name and auth_code:
         return f"{auth_name}:{auth_code}"
-    raise RuntimeError(f"Unrecognised CRS: {crs}")
+    code = auth_name or auth_code
+    if code and code.strip() not in ("0", "EPSG"):
+        return code
+    return f"CUSTOM:{get_identifier_int(crs)}"
+
+
+def get_identifier_int(crs):
+    """
+    Given a CRS, generate a stable, unique identifer for it of type 'int'. Eg: 2193
+    """
+    if isinstance(crs, str):
+        crs = SpatialReference(crs)
+    if not isinstance(crs, SpatialReference):
+        raise RuntimeError(f"Unrecognised CRS: {crs}")
+    auth_code = crs.GetAuthorityCode(None)
+    if auth_code and auth_code.isdigit() and int(auth_code) > 0:
+        return int(auth_code)
+    # Stable code that fits easily in an int32 and won't collide with EPSG codes.
+    return (hash(crs.ExportToWkt()) & 0xFFFFFFF) + 1000000
--- a/sno/gpkg.py
+++ b/sno/gpkg.py
@@ -2,7 +2,7 @@ import collections
 
 import apsw
 
-from sno import spatialite_path
+from sno import gpkg_adapter, spatialite_path
 
 
 def ident(identifier):
@@ -66,7 +66,40 @@ def db(path, **kwargs):
     return db
 
 
-def get_meta_items(db, layer, keys=None):
+def get_gpkg_meta_items_obj(db, layer):
+    """
+    Returns an object that supports .gpkg_meta_items and .get_gpkg_meta_item, for a single layer -
+    to be used with gpkg_adapter
+    """
+
+    class GpkgTableMetaItems:
+        """Collection of gpkg_meta_items for a particular table."""
+
+        def __init__(self, db, layer):
+            self.db_repr = repr(db)
+            self.layer_repr = repr(layer)
+            self._gpkg_meta_items = dict(get_gpkg_meta_items(db, layer))
+
+        def gpkg_meta_items(self):
+            yield from self._gpkg_meta_items.items()
+
+        def get_gpkg_meta_item(self, name):
+            try:
+                return self._gpkg_meta_items[name]
+            except KeyError:
+                if name in gpkg_adapter.GPKG_META_ITEM_NAMES:
+                    return None
+                raise
+
+        def __repr__(self):
+            return f'{self.__class__.__name__}({self.db_repr}, {self.layer_repr})'
+
+        __str__ = __repr__
+
+    return GpkgTableMetaItems(db, layer)
+
+
+def get_gpkg_meta_items(db, layer, keys=None):
     """
     Returns metadata from the gpkg_* tables about this GPKG.
     Keep this in sync with OgrImportSource.gpkg_meta_items for other datasource types.
--- a/sno/gpkg_adapter.py
+++ b/sno/gpkg_adapter.py
@@ -114,7 +114,7 @@ def all_v2_crs_definitions(gpkg_obj):
         definition = gsrs["definition"]
         if not definition or definition == "undefined":
             continue
-        yield crs_util.get_identifier(definition), definition
+        yield crs_util.get_identifier_str(definition), definition
 
 
 def get_table_name(gpkg_obj):
@@ -203,11 +203,9 @@ def _gpkg_srs_id(v2_obj):
 
 def wkt_to_gpkg_spatial_ref_sys(wkt):
     """Given a WKT crs definition, generate a gpkg_spatial_ref_sys meta item."""
-    # TODO: Better support for custom WKT. https://github.com/koordinates/sno/issues/148
     spatial_ref = SpatialReference(wkt)
-    spatial_ref.AutoIdentifyEPSG()
     organization = spatial_ref.GetAuthorityName(None) or "NONE"
-    srs_id = spatial_ref.GetAuthorityCode(None) or 0
+    srs_id = crs_util.get_identifier_int(spatial_ref)
     return [
         {
             "srs_name": spatial_ref.GetName(),
@@ -221,7 +219,7 @@ def wkt_to_gpkg_spatial_ref_sys(wkt):
 
 
 def wkt_to_v2_name(wkt):
-    identifier = crs_util.get_identifier(wkt)
+    identifier = crs_util.get_identifier_str(wkt)
     return f"crs/{identifier}.wkt"
 
 
@@ -355,7 +353,7 @@ def _gkpg_geometry_columns_to_v2_type(ggc, gsrs):
 
     crs_identifier = None
     if gsrs and gsrs[0]["definition"]:
-        crs_identifier = crs_util.get_identifier(gsrs[0]["definition"])
+        crs_identifier = crs_util.get_identifier_str(gsrs[0]["definition"])
 
     extra_type_info = {
         "geometryType": f"{geometry_type} {z}{m}".strip(),
--- a/sno/ogr_import_source.py
+++ b/sno/ogr_import_source.py
@@ -437,20 +437,12 @@ class OgrImportSource(ImportSource):
             return self.get_crs_definition(name)
         raise KeyError(f"No meta item found with name: {name}")
 
-    def get_crs_identifer(self):
-        spatial_ref = self.ogrlayer.GetSpatialRef()
-        if not spatial_ref:
-            return None
-        spatial_ref.AutoIdentifyEPSG()
-        return crs_util.get_identifier(spatial_ref)
-
     def crs_definitions(self):
         if self.is_spatial:
             spatial_ref = self.ogrlayer.GetSpatialRef()
-            if spatial_ref:
-                spatial_ref.AutoIdentifyEPSG()
+            if spatial_ref is not None:
                 yield (
-                    crs_util.get_identifier(spatial_ref),
+                    crs_util.get_identifier_str(spatial_ref),
                     spatial_ref.ExportToWkt(),
                 )
 
@@ -482,9 +474,9 @@ class OgrImportSource(ImportSource):
             "geometryType": f"{geometry_type} {z}{m}".strip(),
         }
 
-        crs_identifier = self.get_crs_identifer()
-        if crs_identifier:
-            extra_type_info["geometryCRS"] = crs_identifier
+        crs_definitions = list(self.crs_definitions())
+        if crs_definitions:
+            extra_type_info["geometryCRS"] = crs_definitions[0][0]
 
         return ColumnSchema(
             ColumnSchema.new_id(), name, "geometry", None, **extra_type_info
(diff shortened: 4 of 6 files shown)


sno/gpkg_adapter.py
```diff
@@ -203,11 +203,9 @@ def _gpkg_srs_id(v2_obj):
 
 def wkt_to_gpkg_spatial_ref_sys(wkt):
     """Given a WKT crs definition, generate a gpkg_spatial_ref_sys meta item."""
-    # TODO: Better support for custom WKT. https://github.com/koordinates/sno/issues/148
     spatial_ref = SpatialReference(wkt)
-    spatial_ref.AutoIdentifyEPSG()
```
rcoup, line 208: It's pretty common to have WKT _without_ authority codes, but otherwise identical to an EPSG definition. This is particularly common with classic ESRI tools. If the parameters exactly match EPSG:1234 then we should use that for it's CRS identifier and that's what `AutoIdentifyEPSG` tries to do.

eg. WKT2 for NZTM (EPSG:2193):
```PROJCRS["NZGD2000 / New Zealand Transverse Mercator 2000",
    BASEGEOGCRS["NZGD2000",
        DATUM["New Zealand Geodetic Datum 2000",
            ELLIPSOID["GRS 1980",6378137,298.257222101,
                LENGTHUNIT["metre",1]]],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["degree",0.0174532925199433]],
        ID["EPSG",4167]],
    CONVERSION["New Zealand Transverse Mercator 2000",
        METHOD["Transverse Mercator",
            ID["EPSG",9807]],
        PARAMETER["Latitude of natural origin",0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8801]],
        PARAMETER["Longitude of natural origin",173,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8802]],
        PARAMETER["Scale factor at natural origin",0.9996,
            SCALEUNIT["unity",1],
            ID["EPSG",8805]],
        PARAMETER["False easting",1600000,
            LENGTHUNIT["metre",1],
            ID["EPSG",8806]],
        PARAMETER["False northing",10000000,
            LENGTHUNIT["metre",1],
            ID["EPSG",8807]]],
    CS[Cartesian,2],
        AXIS["northing (N)",north,
            ORDER[1],
            LENGTHUNIT["metre",1]],
        AXIS["easting (E)",east,
            ORDER[2],
            LENGTHUNIT["metre",1]],
    USAGE[
        SCOPE["unknown"],
        AREA["New Zealand - onshore"],
        BBOX[-47.33,166.37,-34.1,178.63]],
    ID["EPSG",2193]]
```

ESRI WKT for the same CRS:
```
PROJCS["NZGD_2000_New_Zealand_Transverse_Mercator",GEOGCS["GCS_NZGD_2000",DATUM["D_NZGD_2000",SPHEROID["GRS_1980",6378137.0,298.257222101]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Transverse_Mercator"],PARAMETER["False_Easting",1600000.0],PARAMETER["False_Northing",10000000.0],PARAMETER["Central_Meridian",173.0],PARAMETER["Scale_Factor",0.9996],PARAMETER["Latitude_Of_Origin",0.0],UNIT["Meter",1.0]]
```

rcoup (changes requested):
I think `AutoIdentifyEPSG` needs resurrecting

sno/gpkg_adapter.py
olsen232: Ok. I'll investigate more and re-add AutoIdentifyEPSG.

I might end up calling it and checking the result to see if it did what we wanted, if not keep the original.

sno/gpkg.py
```diff
@@ -66,7 +66,41 @@ def db(path, **kwargs):
...
+    class GpkgTableMetaItems:
+        """Collection of gpkg_meta_items for a particular table."""
+
+        def __init__(self, db, layer):
+            self.db_repr = repr(db)
+            self.layer_repr = repr(layer)
+            self._gpkg_meta_items = dict(get_gpkg_meta_items(db, layer))
+
+        def gpkg_meta_items(self):
+            for key, value in self._gpkg_meta_items.items():
```
craigds, line 84: use `yield from`

sno/gpkg_adapter.py
rcoup: I think maybe call `AutoIdentifyEPSG()` and use the auth name+code for the identifier if it returns one. Then keep the original WKT?

rcoup:
> this can change the CRS definition.

Trying to understand what the _downside_ of this is? GDAL expands/updates/fleshes-out assumed/default bits, but the underlying maths will work the same.

olsen232:
I looked into this some more - I have a memory that AutoIdentifyEPSG() had done something destructive once, but I'm no longer sure - maybe it was actually just the roundtrip through a SpatialRef that had changed the WKT, or maybe I was doing something weird anyway like trying to put custom WKT in a file with a standard EPSG code, which I probably shouldn't do.

The roundtrip through a SpatialRef does a couple of things that we may or may not want - 

- Authorities are removed if they don't have an organisation eg:
`AUTHORITY["","PA27"]` -> ∅
This one we could preserve somewhat by pre-processing - we could swap all empty string organisations for "NONE" if we really want to keep these.

- Non-standard tags are removed eg
`EXTENSION["tx_authority","PA27"]` ->  ∅

It's hard for me to be sure which changes are good or bad, and how destructive SpatialRef normalisation seems to be probably depends a lot on what kind of WKT files you start with - how outlandish or junky they are. Before I was using some I found on spatialreference.org, now I'm mostly making sure it works for Panama.

rcoup:
Sure. I'm kinda thinking we preserve the original WKT and if we can identify the authority code, just use it. As long as we reject bad WKT that can't be parsed at all. That way if there's weird extensions in there then they're at least preserved.

Wondering for a fallback identifier about hashing the proj4 string  (`ExportToProj4()` or `projinfo -o PROJ`) rather than just creating a random number, then taking the first few bits of it? AFAICT it only needs to really be unique to a repository anyway?

Proj string for EPSG:2193 is `+proj=tmerc +lat_0=0 +lon_0=173 +k=0.9996 +x_0=1600000 +y_0=10000000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +type=crs`

```python
auth_name = crs.GetAuthorityName(None) or "CUSTOM"
auth_code = crs.GetAuthorityCode(None)

if not auth_code:
  crs_proj = crs.ExportToProj4()
  crs_hash = hashlib.sha256(crs_proj.encode('ascii'))
  crs_hash_code = int.from_bytes(crs_hash.digest()[:3], byteorder='little')
  auth_code = 100000 + crs_hash_code

crs_id = f"{auth_name}:{auth_code}"
```

olsen232:
> preserve the original WKT
Ok then I think this change should mostly be kept as-is - the biggest part of it is to avoid parsing and reserialising WKT if we don't have to (ie for GPKG import where we know how to read the original) - I'll just work harder on
- modifying the definition to include the EPSG code, if we can determine what it is
- using the EPSG code as the srs_id, instead of some hash

If there really is no EPSG code I could indeed use proj4 as a hash. This is a bit counter-productive however - they are two broad approaches we can take:
1. prefer to preserve WKT
2. prefer to normalise WKT

Since we are going with approach 1 - preserve the WKT - it makes more sense to me that the hash should be based on whatever the WKT actually is (maybe it should actually be a hash of the entire WKT string, with whitespace removed).
Making the hash based on the proj4 - ie, a normalised representation of the WKT - would make more sense if we were normalising the WKT too. Using a normalised hash will increase the chance of collisions, in the case that two different custom WKTs have the same proj4 definition, so one of the two (functionally identical) custom CRS will be lost, folded into the other one. Desiring this exact behaviour is the only reason we should switch to a normalised hash - if we don't want this to happen, we should keep an un-normalised hash and preserve both custom CRS. What would we prefer here?

sno/crs_util.py
```diff
@@ -1,14 +1,32 @@
...
+def get_identifier_int(crs):
+    """
+    Given a CRS, generate a stable, unique identifer for it of type 'int'. Eg: 2193
+    """
+    if isinstance(crs, str):
+        crs = SpatialReference(crs)
+    if not isinstance(crs, SpatialReference):
+        raise RuntimeError(f"Unrecognised CRS: {crs}")
+    auth_code = crs.GetAuthorityCode(None)
+    if auth_code.isnumeric():
```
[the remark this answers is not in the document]
olsen232: Done

sno/crs_util.py
```diff
@@ -1,14 +1,32 @@
...
+        return int(auth_code)
+    auth_name = crs.GetAuthorityName(None)
+    # Stable code that fits easily in an int32 and won't collide with EPSG codes.
+    return (hash((auth_name, auth_code)) & 0xFFFFFF) + 1000000
```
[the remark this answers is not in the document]
olsen232: Done

sno/crs_util.py
```diff
@@ -1,14 +1,32 @@
...
-        auth_name = crs.GetAuthorityName(None)
-        auth_code = crs.GetAuthorityCode(None)
+    if not isinstance(crs, SpatialReference):
+        raise RuntimeError(f"Unrecognised CRS: {crs}")
+    auth_name = crs.GetAuthorityName(None)
+    auth_code = crs.GetAuthorityCode(None)
+    if auth_name and auth_code:
         return f"{auth_name}:{auth_code}"
-    raise RuntimeError(f"Unrecognised CRS: {crs}")
+    return auth_name or auth_code
```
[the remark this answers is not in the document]
olsen232: Done - but only if the authority is useless like "0" or "EPSG" - a user could well have written a semi-descriptive unique authority and we in general respect that whatever they have called it, while it may not be the best name, might be better than an opaque hash.
AUTHORITY["", "PA27"]
or
AUTHORITY["olsen232", ""]

sno/gpkg.py
olsen232: Done

rcoup:
> [there] are two broad approaches we can take:
> 1. prefer to preserve WKT
> 2. prefer to normalise WKT

Yeah, seems until we clearly decide this it'll be hard to get anything correct.

My reckons, opinions @hamishcampbell @craigds?

* EPSG codes are canonical and immutable, so there's no ambiguity in the _definition_ of EPSG:2193. Feels to me like we should normalise these.
* Some WKT (ESRI) maps directly onto EPSG definitions, but excludes the codes (because 🤷 ) and/or other defaults.  Feels like we should normalise these too.
* You can define anything you want in Custom WKT. Feels like we shouldn't normalise these in principle

For Custom WKT, do we do a content-hash? Even with normalising whitespace, field ordering and other things could easily affect it unless there's a parser. [`SpatialReference.ExportToWkt()`](https://gdal.org/doxygen/classOGRSpatialReference.html#ae986da88649783b5c194de55c46890a5) does support some options including extensions, does that turn out to be the best way to format it anyway?

sno/gpkg_adapter.py
craigds: the answer here seems to be that AutoIdentifyEPSG just ... doesn't work. 

For both of those inputs you gave us above (if you strip off the ID from the WKT2 so it's not trivial), `AutoIdentifyEPSG` raises a `RuntimeError`...

One way to get something useful is to call `FindMatches` instead:

```python
>>> s.FindMatches()[0]
(<osgeo.osr.SpatialReference; proxy of <Swig Object of type 'OSRSpatialReferenceShadow *' at 0x10e9f06c0> >, 100)
>>> s.FindMatches()[0][0].GetAuthorityCode(None)
'2193'
```

This seems to work for me for both inputs, maybe we should just use that instead?

craigds: With the ESRI example, calling `AutoIdentifyEPSG()` _then_ `FindMatches()` reduces the FindMatches confidence to 25% (it's 100 if you don't call `AutoIdentifyEPSG` first)

Based on this it seems to me that @olsen232 is correct that `AutoIdentifyEPSG` is destructive and fairly useless and should be avoided

olsen232:
> With the ESRI example, calling AutoIdentifyEPSG(), catching the RuntimeError, then calling FindMatches() reduces the FindMatches confidence to 25% (it's 100 if you don't call AutoIdentifyEPSG first)
> 
> Based on this it seems to me that @olsen232 is correct that AutoIdentifyEPSG is destructive and fairly useless and should be avoided

I'm a bit confused by this though - seems like AutoIdentifyEPSG is something you've used a lot in dandelion and it should work?

rcoup:
Here's the various docs:

[`AutoIdentifyEPSG()`](https://gdal.org/doxygen/classOGRSpatialReference.html#acb0373c83927bfd694048da6f79e33ea):
> Set EPSG authority info if possible.
>
> This method inspects a WKT definition, and adds EPSG authority nodes where an aspect of the coordinate system can be easily and safely corresponded with an EPSG identifier. In practice, this method will evolve over time. In theory it can add authority nodes for any object (i.e. spheroid, datum, GEOGCS, units, and PROJCS) that could have an authority node. Mostly this is useful to inserting appropriate PROJCS codes for common formulations (like UTM n WGS84).
>
> ...
>
> Since GDAL 2.3, the FindMatches() method can also be used for improved matching by researching the EPSG catalog.

[`FindMatches()`](https://gdal.org/doxygen/classOGRSpatialReference.html#ab669e0ebf7e0f19e1942172662a50419):
> Try to identify a match between the passed SRS and a related SRS in a catalog.
>
> Matching may be partial, or may fail. Returned entries will be sorted by decreasing match confidence (first entry has the highest match confidence).
> 
> The exact way matching is done may change in future versions. Starting with GDAL 3.0, it relies on PROJ' `proj_identify()` function.

[`proj_identify()`](https://proj.org/development/reference/functions.html#_CPPv413proj_identifyP10PJ_CONTEXTPK2PJPKcPPCKcPPi):
> Identify the CRS with reference CRSs.
>
> The candidate CRSs are either hard-coded, or looked in the database when it is available.
> 
> Note that the implementation uses a set of heuristics to have a good compromise of successful identifications over execution time. It might miss legitimate matches in some circumstances.
> 
> The method returns a list of matching reference CRS, and the percentage (0-100) of confidence in the match. The list is sorted by decreasing confidence.
> 
> 100% means that the name of the reference entry perfectly matches the CRS name, and both are equivalent. In which case a single result is returned. Note: in the case of a GeographicCRS whose axis order is implicit in the input definition (for example ESRI WKT), then axis order is ignored for the purpose of identification. That is the CRS built from `GEOGCS[“GCS_WGS_1984”,DATUM[“D_WGS_1984”,SPHEROID[“WGS_1984”,6378137.0,298.257223563]], PRIMEM[“Greenwich”,0.0],UNIT[“Degree”,0.0174532925199433]]` will be identified to EPSG:4326, but will not pass a `isEquivalentTo(EPSG_4326, util::IComparable::Criterion::EQUIVALENT)` test, but rather `isEquivalentTo(EPSG_4326, util::IComparable::Criterion::EQUIVALENT_EXCEPT_AXIS_ORDER_GEOGCRS)`
> 
> 90% means that CRS are equivalent, but the names are not exactly the same.
> 70% means that CRS are equivalent, but the names are not equivalent.
> 25% means that the CRS are not equivalent, but there is some similarity in the names.

rcoup:
> I'm a bit confused by this though - seems like AutoIdentifyEPSG is something you've used a lot in dandelion and it should work?

@olsen232 I think we have a pile of heuristics, of which AutoIdentifyEPSG() is a part, but it's all pretty old and predates newer PROJ/FindMatches()/etc.

rouault:
>  is `FindMatches()` preferred over `AutoIdentifyEPSG()` in the general case now?

yes, more powerful (but also costly)

olsen232:
I've decided to stop this change here for the time being - the changes that I have so far mean custom CRS with or without an authority has a place to be stored, and the partially filled in authorities from Panama will be preserved. So, that is an improvement, and fixes the linked bug. We don't have any open bugs related to WKT not being normalised: from the discussion above, it seems like something we should probably be doing, but taking the opposite approach and never normalising is also a reasonable solution that doesn't cause any obvious bugs.

The problem with normalising WKT is it hasn't been done until now and so it will count as a diff for existing repos unless we are careful about it - ie, we also normalise the stored definitions before comparing them to the normalised working copy. I can do this if that is what we want, in a follow up change. It will mean that for pre-existing repos, the CRS will forever be stored in a non-canonical form that is normalised every time it is read. But, that is not terrible. What do you think?

For example - our "points" test repo has the following definition:
```
GEOGCS["WGS 84",
    DATUM["WGS_1984",
        SPHEROID["WGS 84",6378137,298.257223563,
            AUTHORITY["EPSG","7030"]],
        AUTHORITY["EPSG","6326"]],
    PRIMEM["Greenwich",0,
        AUTHORITY["EPSG","8901"]],
    UNIT["degree",0.0174532925199433,
        AUTHORITY["EPSG","9122"]],
    AUTHORITY["EPSG","4326"]]
```

Which normalises to the following:
```
 GEOGCS["WGS 84",
     DATUM["WGS_1984",
         SPHEROID["WGS 84",6378137,298.257223563,
             AUTHORITY["EPSG","7030"]],
         AUTHORITY["EPSG","6326"]],
     PRIMEM["Greenwich",0,
         AUTHORITY["EPSG","8901"]],
     UNIT["degree",0.0174532925199433,
         AUTHORITY["EPSG","9122"]],
+    AXIS["Longitude",EAST],
+    AXIS["Latitude",NORTH],
     AUTHORITY["EPSG","4326"]]
```

So, anyone using that repo or a similar one would either have to commit a change that normalises the definition - probably not a good solution - or, that definition will be kept un-normalised on disk forever, but will be normalised whenever it is read. Probably the right solution. Or we do some kind of breaking change.

craigds (approved):
👍 

It's not clear to me that anything else needs to be done re normalising WKT, and if it does need to be done it probably doesn't need to be done _now_

PR #235, opened 2020-09-01
Windows: Fix missing PROJ data in builds

rcoup:
## Description

* PROJ data files don't appear to be picked up by PyInstaller's osgeo/gdal
hook, so include them explicitly.
* add a distcheck PowerShell script for Windows and run it during CI. Currently checks for GDAL & PROJ data files only. Rough corollary of the existing `distcheck.sh` script.
* report PROJ version via `sno --version`

## Related links:

Similar to #42 but for Windows.

## Checklist:


craigds (approved):
should also add a `--crs=epsg:3857` to the `sno diff` call in the e2e tests, to exercise this a bit. otherwise 👍

PR #240, opened 2020-09-03
Pretty wkt output

OrangeOranges:
## Description

Adds syntax highlighting to well-known text output in function get_meta (when writing to a terminal).

Existing tests pass.
Checked syntax highlighting manually in terminal.

### DIFF
--- a/sno/output_util.py
+++ b/sno/output_util.py
@@ -114,6 +114,11 @@ def wkt_whitespace_format(token_iter):
     yield pygments.token.Whitespace, "\n"
 
 
+def write_with_indent(fp, text, indent=""):
+    for line in text.splitlines():
+        fp.write(f"{indent}{line}\n")
+
+
 def wrap_text_to_terminal(text, indent=""):
     """
     Wraps block text to the current width of the terminal.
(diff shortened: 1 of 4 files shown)


sno/output_util.py
```diff
@@ -77,7 +79,47 @@ def format_json_for_output(output, fp, json_style="pretty"):
         return json.dumps(output, **JSON_PARAMS[json_style]) + "\n"
 
 
-def wrap_text_to_terminal(text, indent=''):
+def format_wkt_for_output(output, fp):
+    """
+    If the given filelike object is a terminal, adds whitespace and syntax highlighting to the output.
```
craigds, line 84: IMO the whitespace is useful even if it's not going to a terminal

craigds (commented):
screenshot from `sno meta get` output:

Screen Shot 2020-09-03 at 3 23 38 PM


looks nice!

I think it needs a quick test for the whitespace changes at least. i.e. just invoke `sno meta get <datasetname> metadata/dataset.json` and check the output has the expected whitespacing :)

craigds:
macOS build failure appears to be a code-signing thing; I guess it doesn't work on third-party PRs. Can ignore that

rcoup:
> macOS build failure appears to be a code-signing thing; I guess it doesn't work on third-party PRs. Can ignore that

Should skip rather than failing, I'll get it fixed on master

PR #245, opened 2020-09-08
import: No empty commits without --allow-empty

craigds:
## Description
This adds the --allow-empty option, which has the same meaning as
in `commit` and `apply`.

This behaviour mostly only affects `--replace-existing`,
since `import` will never produce an empty commit if you're not
replacing anything.

Because import uses fast-import, this was a little tricky. fast-import will happily do empty commits and there's no easy way to stop it from doing so. 

So instead, I made fast-import use a temporary branch to do its importing on. Then afterwards I checked whether the commit was empty before updating the original branch ref to the new commit.

## Related links:

#243 did the same thing for `apply`

## Checklist:

### DIFF
--- a/sno/fast_import.py
+++ b/sno/fast_import.py
@@ -95,11 +95,10 @@ def fast_import_tables(
         import_branch = f'refs/heads/{uuid.uuid4()}'
 
         # may be None, if head is detached
-        orig_branch = get_head_branch(repo, allow_detached=True)
-        header = generate_header(repo, sources, message, branch=import_branch)
-    else:
         orig_branch = get_head_branch(repo)
-        import_branch = orig_branch
+        header = generate_header(repo, sources, message, import_branch)
+    else:
+        import_branch = None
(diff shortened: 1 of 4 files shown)


sno/fast_import.py
```diff
@@ -87,16 +89,26 @@ def fast_import_tables(
         f"--depth={max_delta_depth}",
     ] + list(extra_cmd_args)
 
+    orig_branch = get_head_branch(repo)
     if header is None:
-        header = generate_header(repo, sources, message)
+        # import onto a temp branch. then reset the head branch afterwards.
+        # this allows us to check the result before updating the orig branch.
+        import_branch = f'refs/heads/{uuid.uuid4()}'
```
olsen232, line 96: This code looks like it would be a better workaround than the existing workaround I have for doing imports at detached HEAD - see temporary_branch in init.py
If you fix this code so that it also works when importing at detached HEAD, you could get rid of my workaround, and fast_import.py would have a simpler interface that supports all cases.

PR #246, opened 2020-09-08
Support patches that create or delete datasets

olsen232:
## Related links:

Fixes https://github.com/koordinates/sno/issues/239

## Checklist:

### DIFF
--- a/sno/dataset2.py
+++ b/sno/dataset2.py
@@ -169,12 +169,6 @@ class Dataset2(DatasetStructure):
 
     @property
     @functools.lru_cache(maxsize=1)
-    def schema(self):
-        """Load the current schema from this dataset."""
-        return Schema.loads(self.get_data_at(self.SCHEMA_PATH))
-
-    @property
-    @functools.lru_cache(maxsize=1)
     def crs_identifier(self):
         for col in self.schema:
             if col.data_type == "geometry":
@@ -401,9 +395,11 @@ class Dataset2(DatasetStructure):
         if not meta_diff:
             return
 
-        conflicts = False
+        # Applying diffs works even if there is no tree yet created for the dataset,
+        # as is the case when the dataset is first being created right now.
+        meta_tree = self.meta_tree if self.tree is not None else ()
 
-        meta_tree = self.meta_tree
+        conflicts = False
         with tree_builder.cd(self.META_PATH):
             for delta in meta_diff.values():
                 name = delta.key
--- a/sno/git_util.py
+++ b/sno/git_util.py
@@ -6,6 +6,38 @@ import pygit2
 from .timestamps import tz_offset_to_minutes
 
 
+def get_head_tree(repo):
+    """Returns the tree at the current repo HEAD."""
+    if repo.is_empty:
+        return None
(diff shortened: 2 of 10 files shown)


craigds (approved):
LGTM

sno/dataset2.py
```diff
@@ -401,9 +395,11 @@ def apply_meta_diff(self, meta_diff, tree_builder):
         if not meta_diff:
             return
 
-        conflicts = False
+        # Applying diffs works even if there is no tree yet created for the dataset,
+        # as is the case when the dataset is first being created right now.
+        meta_tree = self.meta_tree if self.tree is not None else ()
 
-        meta_tree = self.meta_tree
+        conflicts = False
```
rcoup, line 402: would prefer to use `has_`/`is_`/`can_`/etc as a prefix for boolean variable names

sno/git_util.py
```diff
@@ -6,6 +6,38 @@
 from .timestamps import tz_offset_to_minutes
 
 
+def get_head_tree(repo):
+    """Returns the tree at the current repo HEAD."""
```
rcoup, line 10: Can the docstring be expanded with the comments about when/why `None` is returned?

sno/git_util.py
```diff
@@ -6,6 +6,38 @@
...
+        return None
+    try:
+        return repo.head.peel(pygit2.Tree)
+    except pygit2.GitError:
+        # This happens when the repo is not empty, but the current HEAD has no commits.
+        return None
+
+
+def get_head_commit(repo):
+    """Returns the commit at the current repo HEAD."""
```
rcoup, line 21: Can the docstring be expanded with the comments about when/why `None` is returned?

sno/rich_tree_builder.py
```diff
@@ -13,18 +16,27 @@ class RichTreeBuilder:
     Conflicts are not detected.
     """
 
-    def __init__(self, repo, initial_root_tree):
+    def __init__(self, repo, initial_root_tree, auto_include_version_blobs=True):
         """
         The repo and an initial root tree which will be updated.
         All paths are specified relative to this tree - the root tree at a particular commit is a good choice.
```
rcoup, line 22: document what `auto_include_version_blobs` does, since it's not obvious?

rcoup (commented):
Nice, some minor doc/style comments

sno/dataset2.py
olsen232: Done

sno/git_util.py
olsen232: Done. Also fixed the try/except pattern - we should have just been checking if head_is_unborn

sno/git_util.py
olsen232: Done

sno/rich_tree_builder.py
olsen232: Actually, this should be at a different level - the tree builder should just build the tree its told to build and not be clever about it.
I've moved it to the caller - RepositoryStructure.create_tree_from_diff - and added a comment

Issue #130, opened 2020-06-21
Populate reflogs


craigds:
`git reflog` in a sno repo is empty. So it's difficult to double-check complex changes after the fact.

We should populate the reflogs and add a `sno reflog` command (it can probably just exec `git reflog` for now)

PR #251, opened 2020-09-10
Reflogs

OrangeOranges:
## Description

Enables reflogs and adds sno reflog command to display them.
Thanks to Andrew for finding the core.logAllRefUpdates="always" setting, which magically enables writing of reflogs.

## Checklist:


OrangeOranges:
Oh no all the checks failed...

craigds (commented):
This looks great. much simpler than I thought, nice work 👍 

Fixes #130 

I checked the test failures and was able to reproduce the failure on my laptop:

```
$ pytest -k test_e2e
...
E                 At index 4 diff: 'HEAD@{4}: ' != 'HEAD@{4}: fast-import'
E                 Full diff:
E                   [
E                    'HEAD@{0}: commit (merge): merge-1',
E                    'HEAD@{1}: checkout: moving from edit-1 to master',
E                    'HEAD@{2}: commit: commit-1',
E                    'HEAD@{3}: checkout: moving from master to edit-1',
E                 -  'HEAD@{4}: fast-import',
E                 +  'HEAD@{4}: ',
```

sno/init.py
```diff
@@ -343,6 +343,9 @@ def init(
     write_repo_version_config(repo, repo_version)
     WorkingCopy.write_config(repo, wc_path, bare)
 
+    # Enable writing to reflogs
+    repo.config["core.logAllRefUpdates"] = "always"
```
rcoup, line 347: so this automatically persists/writes into the repo `.git/config`? #til

OrangeOranges: Yes it writes into the config file.

tests/test_branch.py
```diff
@@ -51,7 +51,7 @@ def test_branches(
         r = cli_runner.invoke(["checkout", "HEAD~1"])
         assert r.exit_code == 0, r
 
-        assert text_branches(cli_runner) == ["* (no branch)", "  master"]
+        assert text_branches(cli_runner) == ["* (HEAD detached at 63a9492)", "  master"]
```
rcoup, line 54: so this changes because git reads it back from the reflog? 🤯 👍

rcoup (approved):
Nice work 👏

rcoup:
@OrangeOranges GTG, but lets get CI green before merging.

https://github.com/koordinates/sno/pull/253 should sort that, then we can rebase this one and it should be green

tests/test_branch.py
craigds: it's inexplicable to me, but yes git seems to do this using the reflog... somehow.

PR #262, opened 2020-09-22
Tidy up sno repo directories

olsen232:
## Description

- Moves git internals into hidden `.sno` directory, so they don't clutter up the main folder / get deleted or modified accidentally.
- Preserves tools that auto-detect git repositories (eg git-prompt.sh) by making `.git` file point to `.sno` directory.
- Prevents git commands that shouldn't work from working with a "locked index file" - the binary format is extended with an extension called ".sno" that git does not support. This causes error messages like the following:

```
$ git checkout master
error: index uses .sno extension, which we do not understand
fatal: index file corrupt
```
(Which is probably about the most helpful error message we can provide to the user, given we didn't write git and can't fully control its error messages.)

## Related links:

First part of:
https://github.com/koordinates/sno/issues/147

Not done: maintaining a working tree of version-controlled data that doesn't belong in datasets, letting the user check files into it.

## Checklist:


craigds (changes requested):
This all makes sense to me except we need a way to create legitimately-bare repos for remotes.

ie. `sno init --no-checkout` needs to produce a repo with `core.bare = true`

Otherwise 👍

rcoup:
Will review more this morning, but we should check all the files/dirs are properly set to hidden on Windows (which is a file attribute).

Might require solving cross-platform pip-compile for installing pywin32 though 🐇🕳

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+            ...      # etc
+    """
+
+    # Note: pygit2 can read the .sno directory okay, but it won't create it, so we do it ourselves using Path.rename()
+    # Note: We could also call `git init --separate-git-dir=.sno` to crete it but it doesn't work very well - it
+    # converts all paths to be absolute, which would mean sno repos wouldn't be renameable.
+
+    # Move .git to .sno but add a reference at .git so git can find it.
+    dot_git_path = repo_root_path / ".git"
+    assert dot_git_path.is_dir()
```
rcoup, line 82: `assert` statements are a bit of a danger in production Python code, since `python -O` compiles them away, is better to conditionally raise an `AssertionError` explicitly.

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+    repo_files.write_repo_file(repo, repo_files.INDEX, LOCKED_EMPTY_GIT_INDEX)
+
+    # Force writing to reflogs.
+    repo.config["core.logAllRefUpdates"] = "always"
+    return repo
+
+
+def is_sno_repo(pygit_repo):
+    if pygit_repo is not None:
+        # Older sno repos were just bare repos, newer sno repos are non-bare git repos that use ".sno" instead of ".git"
```
rcoup, line 102:
```suggestion
        # Older sno repos were just bare git repos, newer sno repos are non-bare git repos that use ".sno" instead of ".git"
```

rcoup: We can still have bare sno repos though? (ie: for server-side repo hosting)

But that behaviour is set through git config using the `sno.workingcopy.*` values? (ie: do i have a working copy)

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def get_sno_workdir(pygit_repo):
+    # Older sno repos are bare and use repo.path as the "workdir" - that is, where working copies are written -
+    # Newer ones have an actual workdir.
+    return Path(pygit_repo.path) if pygit_repo.is_bare else Path(pygit_repo.workdir)
+
+
+@contextlib.contextmanager
+def unlocked_git_index(pygit_repo):
```
rcoup, line 114: we should probably start to resurrect `fsck` again to allow us to do things like "re-lock" an index if this process crashes hard and the index doesn't get reset.

sno/checkout.py
```diff
@@ -105,7 +105,6 @@ def checkout(ctx, branch, force, discard_changes, refish):
     reset_wc_if_needed(repo, commit, discard_changes=discard_changes)
 
     repo.set_head(head_ref)
-    repo.reset(commit.oid, pygit2.GIT_RESET_SOFT)
```
rcoup, line 108: Why aren't these needed any more?

sno/sno_repo.py
rcoup: Would prefer some more specific tests around locking and unlocking behaviours

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+
+    _remove_locked_sno_extension(index_path)
+    try:
+        yield
+    finally:
+        _add_locked_sno_extension(index_path)
+
+
+def _remove_locked_sno_extension(index_path):
+    index_contents = index_path.read_bytes()
```
rcoup, line 136: indexes can potentially be GiB-sized, right? 🤨

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+    index_contents = index_contents[:-20]  # Trim off checksum
+    if index_contents[-8:] == LOCKED_SNO_EXTENSION:
```
rcoup, line 138: Should we be logging a warning if it's not how we expect?

sno/sno_repo.py
rcoup: Oh, is always empty. Should we check it's not become GiB-size before we read it all into memory?

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+
+    repo = pygit2.init_repository(str(repo_root_path.resolve()))
+    repo = convert_git_repo_to_sno_repo(repo_root_path)
+    write_repo_version_config(repo, repo_version)
+    assert is_sno_repo(repo)
+    return repo
+
+
+def convert_git_repo_to_sno_repo(repo_root_path):
```
rcoup, line 61: again, think we should make sure we have a clear recovery off-ramp for this aborting part-way through. 

eg: no index or no reflog config would be annoying and non-obvious, even though `.sno/` is there and it'd otherwise seem to "work"

Maybe just a case of re-ordering the steps?

rcoup (changes requested):
Looking good :)

1. Could we have a way to upgrade an existing sno repo to the new WC structure?
2. Should we start subclassing PyGit2 classes (eg. Repository) rather than having loads of utility methods?

sno/clone.py
```diff
@@ -88,15 +86,14 @@ def clone(ctx, bare, wc_path, wc_version, do_progress, depth, url, directory):
     except subprocess.CalledProcessError as e:
         sys.exit(translate_subprocess_exit_code(e.returncode))
 
-    repo = pygit2.Repository(str(repo_path.resolve()))
+    repo = sno_repo.convert_git_repo_to_sno_repo(repo_path)
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy.py
```diff
@@ -78,10 +79,16 @@ def del_repo_cfg(key):
...
             return
 
-        path = path or f"{Path(repo.path).resolve().stem}.gpkg"
+        path = path or cls.default_path(repo)
         repo_cfg["sno.workingcopy.path"] = str(path)
         del_repo_cfg("sno.workingcopy.bare")
 
+    @classmethod
+    def default_path(cls, repo):
+        workdir = sno_repo.get_sno_workdir(repo)
```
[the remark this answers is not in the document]
olsen232: Done

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+import struct
+from pathlib import Path
+
+import pygit2
+
+from .repository_version import write_repo_version_config
+from . import repo_files
+
+
+# Utilities for maintaining file structure of a sno repo - not the versioned controlled blobs,
```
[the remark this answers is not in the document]
olsen232: Done

sno/sno_repo.py
```diff
@@ -0,0 +1,146 @@
...
+# but the .git file, .sno directory, .sno/index file. pygit2 mostly takes care of this for us,
+# but since we differ slightly from a standard git repository, there are a few extra things to take care of.
+
+
+def _append_checksum(data):
+    return data + pygit2.hash(data).raw
+
+
+INDEX_VERSION = 2
+EMPTY_GIT_INDEX = struct.pack(">4sII", b"DIRC", INDEX_VERSION, 0)
```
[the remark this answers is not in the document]
olsen232: Done

sno/sno_repo.py
olsen232: ... Right now it is always empty since there should be no way for git to mess with it. Potentially one day we will actually use the index - ie if we do worktrees + sparse checkouts. This code would work as-is for when we do, except, it reads everything into memory. Unfortunately the index has a checksum at the end which means that doing tiny edits to it is tricky / inefficient. For now, I'll just drop support for edits to the index: I'll just hard-code a locked and an unlocked version of the index, and write those, so I don't have to read it into memory.

sno/sno_repo.py
olsen232: Done

sno/sno_repo.py
olsen232: Re-ordered the steps. Moving .git to .sno is now the first step, and creating the .git file that points to .sno is now the last step. Any failure between these two steps now leaves you with a repo that nothing can find unless they specifically know to look in the .sno directory - which neither git or sno do automatically without the .git file.

Of course, we could potentially fail to rename .git to .sno - attempted to ameliorate this by making any command that would call convert_git_repo_to_sno_repo give up before creating a git repo if .git or .sno already exist, which is the most likely reason the rename would fail.

sno/checkout.py
olsen232: There's nothing to reset that isn't taken care of by `repo.set_head(head_ref)`
- we don't want to modify anything in refs/heads - just HEAD, which set_head takes care of
- we don't have an index
- we don't have a worktree

I think GIT_RESET_SOFT does the right thing here (ie, nothing we didn't already do) but it also checks the index file, which means the index needs to be unlocked. Easier just to get rid of it since it's a no-op.

sno/sno_repo.py
olsen232: Added some more specific tests in test_sno_repo.py

If it seems best, we could eagerly re-lock the index every time we open up a SnoRepo, even if we expect it to be already locked - as opposed to now, where we leave it how it is until next time we need to unlock / relock it.

sno/clone.py
```diff
@@ -88,15 +86,14 @@ def clone(ctx, bare, wc_path, wc_version, do_progress, depth, url, directory):
...
```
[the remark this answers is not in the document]
olsen232: Improved this somewhat:
Re-ordered the steps of convert_git_repo_to_sno_repo to make it more robust.
Added checks here so we don't even start the clone if we can see that convert_git_repo_to_sno_repo won't succeed.
The `git clone` itself is the most likely to fail, in which case, nothing is created.

olsen232:
Root level comments:
@craigds:
I now use core.bare to mark "new+tidy" sno repos as bare. Old+bare sno repos are always bare anyway. So, your requirement is met.

@rcoup 
1. Allow upgrade of sno repos to "new+tidy": Not done. `sno upgrade` almost works for this already, except it is specifically not allowed to upgrade a v2 repo to a v2 repo. Otherwise it would already work. However, it would be much slower than is necessary, and it would not work in-place.

I could special-case `sno upgrade` so it can do in-place v2 "old+bare" -> v2 "new+tidy" repos.
I could even decide that v2 + "new+tidy" is actually v3 - seems mostly unnecessary though, and like it might be more annoying than helpful to make that distinction.

2. Moved away from utility methods, subclassed pygit2.Repository with SnoRepo. Started to move things in here - the SnoRepo object is the logical home for the utility methods in sno_repo but also those in repository_version, repo_files, maybe some of git_util, etc... but I've made a start.

rcoup:
>> Allow upgrade of sno repos to "new+tidy":
>
> Not done. 

Need _some sort_ of solution here that doesn't involve re-cloning. 

> `sno upgrade` almost works for this already, except it is specifically not allowed to upgrade a v2 repo to a v2 repo. Otherwise it would already work. However, it would be much slower than is necessary, and it would not work in-place.

All we really need is a way of calling `SnoRepo.convert_git_repo_to_sno_repo()`, right?

> I could special-case `sno upgrade` so it can do in-place v2 "old+bare" -> v2 "new+tidy" repos.

I think this is my preference, it can always be removed again later. `sno upgrade worktree` or some other incantation?

> I could even decide that v2 + "new+tidy" is actually v3 - seems mostly unnecessary though, and like it might be more annoying than helpful to make that distinction.

I agree, I'm not that keen on this option.

sno/sno_repo.py
rcoup: Not sure about every time, feels like that'd mess up concurrent calls to `sno`?

Whatever we can come up with that is reasonably sane + safe :)

sno/clone.py
```diff
@@ -64,10 +63,13 @@ def get_directory_from_url(url):
     type=click.Path(exists=False, file_okay=False, writable=True),
     required=False,
 )
-def clone(ctx, bare, wc_path, wc_version, do_progress, depth, url, directory):
+def clone(ctx, bare, wc_path, do_progress, depth, url, directory):
     """ Clone a repository into a new directory """
 
     repo_path = Path(directory or get_directory_from_url(url))
+    assert not (repo_path / ".git").exists()
```
rcoup, line 70: same AssertionError thing

sno/sno_repo.py
```diff
@@ -1,146 +1,255 @@
...
+        if isinstance(root_path, Path):
+            root_path = str(root_path.resolve())
+
+        try:
+            super().__init__(root_path)
+        except pygit2.GitError:
+            raise NotFound("Not an existing sno repository", exit_code=NO_REPOSITORY)
+
+        self.gitdir_path = Path(self.path).resolve()
+        assert self.is_old_bare_repo() or self.is_new_tidy_repo()
```
rcoup, line 90: same assertionerror issue

rcoup:
> Will review more this morning, but we should check all the files/dirs are properly set to hidden on Windows (which is a file attribute).

need to check this one as well

sno/sno_repo.py
olsen232: Done

sno/clone.py
olsen232: Done

olsen232:
- `sno upgrade-to-tidy REPO` now works
- Tested on windows, .git and .sno are hidden
- got rid of lots of asserts, replaced with better errors 
- Made clone and init *not* create functional repositories as the first step - or at least not in the main directory. Instead, they create the gitdir in .sno, and they create the workdir in a temporary folder called .init or .clone - the workdir is in any case just a .git file that links to the gitdir. This workdir is deleted immediately, and then the rest of the initialisation steps run on an "unactivated" repo - one where no git commands or sno commands will work. The last step is activating it: writing a .git file is written to the main directory, which causes those commands to start working.

sno/sno_repo.py
```diff
@@ -0,0 +1,318 @@
...
+        True if this sno repo is bare - it has no sno working copy.
+        The repo may or may not also be a bare git repository - this is an implementation detail.
+        That information is at super().is_bare
+        """
+        repo_cfg = self.config
+        bare_key = self.BARE_CONFIG_KEY
+        return repo_cfg.get_bool(bare_key) if bare_key in repo_cfg else False
+
+    @contextlib.contextmanager
+    def unlocked_git_index(self):
```
craigds, line 262: > GIT_INDEX_FILE is the path to the index file (non-bare repositories only).


might be worth using this to change where git subcommands looks for the index, rather than deleting & re-creating it. That will make this safe against concurrent runs and crashes

craigds (approved):

sno/sno_repo.py
olsen232: Done! Not sure how I missed this - it was the kind of thing I was hoping to do

PR #264, opened 2020-09-23
Shard CI test on windows into 10 - green windows ✅

olsen232:
## Description

Windows unit tests normally fail before the end, not for any particular reason.
Maybe if the end was 10x closer, they wouldn't fail.
An experiment: this might green the build for us, which would be good, and if it does green the build, it will also give us insight into why the tests are failing.

### DIFF
--- a/makefile.vc
+++ b/makefile.vc
@@ -62,14 +62,15 @@ test: $(app) venv\.test.installed
 
 ci-test:
 	set CI=true
-	pytest \
-		--verbose \
-		-p no:sugar \
-		--cov-report term \
-		--cov-report html:test-results\coverage/ \
-		--junit-xml=test-results\junit.xml \
-		--benchmark-enable \
-		-p no:xdist
+	FOR /L %%I IN (0, 1, 9) DO \
+		pytest \
+			--num-shards 10 \
+			--shard-id %I \
+			--verbose \
+			--benchmark-enable \
+			--no-cov \
+			-p no:sugar \
+			-p no:xdist
 
 # Packaging
 
--- a/requirements/test.in
+++ b/requirements/test.in
@@ -10,5 +10,6 @@ pytest-sugar
 pytest-helpers-namespace
 pytest-benchmark[aspect]
 pytest-profiling
+pytest-shard
 lovely-pytest-docker
 html5lib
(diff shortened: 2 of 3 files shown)


UberMouse:
I'm curious if this will actually make anything faster? Github CI machines are only dual cores.

craigds:
it isn't for parallelism, they still run serially. It's just because windows CI runs tend to crash before the end, and we wondered if doing it in smaller pieces would help.

makefile.vc
```diff
@@ -62,14 +62,15 @@ test: $(app) venv\.test.installed
...
-		--verbose \
-		-p no:sugar \
-		--cov-report term \
-		--cov-report html:test-results\coverage/ \
-		--junit-xml=test-results\junit.xml \
-		--benchmark-enable \
-		-p no:xdist
+	FOR /L %%I IN (0, 1, 9) DO \
+		pytest \
+			--num-shards 10 \
```
rcoup, line 67: are the shards deterministic?

olsen232: I think yes - 
certainly if you run the exact same command twice in a row without changing anything, the tests assigned to a particular shard is the same each time. But I think there is a stronger guarantee: a test named X will always be in shard `hash(X) % num-shards` - so as long as the test keeps the same name and you keep the same number of shards, it will always be in the same shard.
This guess about how it works is based on a single line of documentation "Shards tests based on a hash of their test name", and the fact that I observe different shards have different numbers of tests in them.

rcoup: If it's using `hash()` then I think it's only guaranteed to be deterministic in the same interpreter instance? Though `PYTHONHASHSEED` affects for scalar objects, hashes of lists/dicts/objects are derived from their memory addresses.

Something based on a hashlib algorithm would be more useful if it's possible, since it'd be repeatable across CI runs.

@craigds Didn't we go through this with dandelion/something?

olsen232: ... well, the name of a test won't be a list/dict/object... just a string. Furthermore the rest of that line of documentation says "Shards tests based on a hash of their test name enabling easy parallelism across machines" - so I think the hash function (whatever it is) is supposed to be relatively machine-and-context independent, although I wouldn't know if it will change if we, for instance, upgrade python. 

I kind of thought you would be happy the windows test is finally green... it seemed preferable to me. Let me know if I should roll it back

rcoup: is all good :)

PR #270, opened 2020-10-02
Implement feature diff: dataset-rev<>postgis-wc

olsen232:
## Description

Adds support for
- writing a dataset to a postgis table
- generating the feature diff between a postgis table and the dataset

Lots of code is shared between postgis working copy and gpkg working copy, much of this code has been factored out and moved up into the superclass, WorkingCopy.

Still TODO in later PRs - add support for:
- meta diffs between postgis table and the dataset
- resetting postgis wc to a particular commit
- configuring repository to have a postgres working copy
- improve tests - should be possible to repurpose GPKG working copy tests, instead of rewriting everything.

## Related links:

### DIFF
--- a/sno/crs_util.py
+++ b/sno/crs_util.py
@@ -1,8 +1,8 @@
 from osgeo.osr import SpatialReference
 
-
 from .cli_util import StringFromFile
 from .geometry import make_crs
+from .serialise_util import uint32hash
 
 
 class CoordinateReferenceString(StringFromFile):
@@ -49,4 +49,29 @@ def get_identifier_int(crs):
     if auth_code and auth_code.isdigit() and int(auth_code) > 0:
         return int(auth_code)
     # Stable code that fits easily in an int32 and won't collide with EPSG codes.
-    return (hash(crs.ExportToWkt()) & 0xFFFFFFF) + 1000000
+    return (uint32hash(crs.ExportToWkt()) & 0xFFFFFFF) + 1000000
+
+
+def get_identifier_int_from_dataset(dataset, crs_name=None):
+    """
+    Get the CRS attached to this dataset with a particular name eg "EPSG:2193",
+    and return an integer to uniquely identify it, eg 2193.
+    (This still works even if the CRS is custom and doesn't have an obvious number embedded in it).
+    crs_name can be ommitted if there is no more than one geometry column.
+    """
+
+    if crs_name is None:
+        geom_columns = dataset.schema.geometry_columns
+        num_geom_columns = len(geom_columns)
+        if num_geom_columns == 0:
+            return None
+        elif num_geom_columns == 1:
+            crs_name = geom_columns[0].extra_type_info.get("geometryCRS", None)
+        else:
+            raise ValueError("Dataset has more than one geometry column")
+
+    if crs_name is None:
+        return None
+
+    definition = dataset.get_crs_definition(crs_name)
+    return get_identifier_int(definition)
--- a/sno/geometry.py
+++ b/sno/geometry.py
@@ -25,6 +25,8 @@ class Geometry(bytes):
 
     @classmethod
     def of(cls, bytes_):
+        if isinstance(bytes_, Geometry):
+            return bytes_
         return Geometry(bytes_) if bytes_ else None
 
     def __str__(self):
@@ -179,7 +181,7 @@ def normalise_gpkg_geom(gpkg_geom):
             # everything is fine, no need to roundtrip via OGR
             # just need to set srs_id to zero if it's not already
             if gpkg_geom[4:8] == b"\x00\x00\x00\x00":
-                return gpkg_geom
+                return Geometry.of(gpkg_geom)
             else:
                 return Geometry.of(gpkg_geom[:4] + b"\x00\x00\x00\x00" + gpkg_geom[8:])
 
@@ -374,7 +376,7 @@ def _bo(is_le):
     return "<" if is_le else ">"
 
 
-def geom_to_ewkb(gpkg_geom):
+def gpkg_geom_to_ewkb(gpkg_geom):
     """
     Parse GeoPackage geometry values to a PostGIS EWKB value
     http://www.geopackage.org/spec/#gpb_format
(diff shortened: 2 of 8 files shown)


sno/crs_util.py
```diff
@@ -50,3 +50,23 @@ def get_identifier_int(crs):
         return int(auth_code)
     # Stable code that fits easily in an int32 and won't collide with EPSG codes.
     return (hash(crs.ExportToWkt()) & 0xFFFFFFF) + 1000000
```
rcoup, line 52: `hash()` in Python isn't stable...

sno/crs_util.py
```diff
@@ -50,3 +50,23 @@ def get_identifier_int(crs):
...
+    Get the CRS attached to this dataset with a particular name eg "EPSG:2193",
+    and return an integer to uniquely identify it, eg 2193.
+    (This still works even if the CRS is custom and doesn't have an obvious number embedded in it).
+    crs_name can be ommitted if there is only one geometry column.
+    """
+
+    if crs_name is None:
+        geom_columns = dataset.schema.geometry_columns
+        if len(geom_columns) == 1:
+            crs_name = geom_columns[0].extra_type_info.get("geometryCRS", None)
```
rcoup, line 66: what happens if we have >1 geometry column?

sno/geometry.py
```diff
@@ -396,19 +400,19 @@ def geom_to_ewkb(gpkg_geom):
...
+        ewkb += struct.pack(f"{bo}I", crs_id)
 
     ewkb += gpkg_geom[(wkb_offset + 5) :]
 
     return ewkb
 
 
-def hexewkb_to_geom(hexewkb):
+def hexewkb_to_gpkg_geom(hexewkb):
     """
```
rcoup, line 416: this method doesn't match up with what we decided on wrt geometry formatting. eg. envelopes & endianness

sno/working_copy/postgis.py
```diff
@@ -116,83 +119,113 @@ def session(self, bulk=0):
...
+            0: "integer",
+            16: "smallint",
+            32: "integer",
+            64: "bigint",
+        },
+        "interval": "interval",
+        "numeric": "numeric",
+        "text": "text",
+        "time": "time",
+        "timestamp": "timestamp",
```
rcoup, line 140: what happened here wrt time zones?

sno/working_copy/postgis.py
```diff
@@ -116,83 +119,113 @@ def session(self, bulk=0):
...
-            cols[col["name"]] = SQL(" ").join(col_spec)
+            if col.pk_index is not None:
+                col_spec.append(SQL("PRIMARY KEY NOT NULL"))
+                pk_field = col.name
+            pg_cols[col.name] = SQL(" ").join(col_spec)
 
-        self.L.debug("Sqlite>Postgres column mapping: %s -> %s", sqlite_cols, cols)
-        return cols, pk_field
+        self.L.debug(
+            "Sqlite>Postgres column mapping: %s -> %s", dataset.schema, pg_cols
```
rcoup, line 205: we're not really mapping sqlite here anymore :)

sno/working_copy/postgis.py
```diff
@@ -251,11 +284,17 @@ def create(self):
...
+                    func=Identifier(self.schema, "_sno_track_trigger"),
                     tracking_table=self.TRACKING_TABLE,
                 )
             )
 
+    def delete(self):
+        """ Delete all tables in the schema"""
+        with self.session() as db:
+            dbcur = db.cursor()
+            dbcur.execute(SQL("DROP SCHEMA {} CASCADE").format(Identifier(self.schema)))
```
rcoup, line 296: mmmmm... not sure we want to cascade drops by default.

if someone makes a foreign key to a sno dataset table, and then we attempt to drop it for whatever reason, we shouldn't erase data from their table too. It might get the repo semi-stuck, but that's better than clearing user data.

would prefer to drop tables individually (with `RESTRICT` to prevent any issues with related tables), then drop the schema afterwards.

sno/working_copy/postgis.py
```diff
@@ -396,13 +427,23 @@ def write_full(self, commit, *datasets, **kwargs):
...
+
                 feat_count = 0
                 t0 = time.monotonic()
                 t0p = t0
 
                 CHUNK_SIZE = 10000
                 for rows in self._chunk(dataset.feature_tuples(col_names), CHUNK_SIZE):
-                    dbcur.executemany(sql_insert_features, ([tuple(r)] for r in rows))
+
+                    # FIXME - find a better way of forcing geometries to have the right CRS id
```
rcoup, line 440: `INSERT INTO mytable (geom, ...) VALUES (SetSRID('...'::geometry, 12345), ...);`?

rcoup (commented):
lot of stuff in here 😄

sno/crs_util.py
olsen232: Fixed.
(This code is not often exercised, it's only used if there is a custom CRS without a code)

sno/crs_util.py
olsen232: Fixed this to raise ValueErrors for >1 of geometry columns. I think much of sno doesn't support having more than 1 geometry column - the new model does, in theory, but GPKG does not and any code older than V2 does not.

sno/geometry.py
olsen232: Fixed by calling normalise_gpkg_geometry. There might be a more efficient way to do it: added a TODO

sno/working_copy/postgis.py
olsen232: Not very sure. Getting this perfectly right will take some more thought - I don't know if we have a plan for what to do when working copies support different things eg one working copy supports "with timezone" and "without timezone" (ie POSTGIS) variants, the other does not (GPKG). I wrote the mapping above because its simple - I'll add a TODO to revisit this, but I can't write the final revision of this code without a plan for how we roundtrip types through less full-featured working copies.

sno/working_copy/postgis.py
olsen232: Done

sno/working_copy/postgis.py
olsen232: That is a good plan - not required for a demo however, so I'll add a TODO.

sno/working_copy/postgis.py
olsen232: Done

sno/crs_util.py
rcoup: sure, I think it'll be a limitation for quite a while.

sno/working_copy/postgis.py
rcoup: Yeah, needs a separate discussion about the general approach, then details for each thing we run into.

rcoup:
I guess at some point we need to decide whether SQLAlchemy/etc would do a better job at DB syntax abstraction than we will. Not that we're running a tonne of SQL, but it does add up. Any thoughts here @craigds?

PR #273, opened 2020-10-08
Makes reset() work for postgis WC, adds some tests.

olsen232:
## Description

Making reset work mostly means moving WC.reset functionality from gpkg.py into base.py, where it is shared by all working copies. A little bit of code that is database specific must be extracted out.

Also - started using register_type to make sure types coming out of the postgis database end up in the right format in python. These conversions were previously done manually for geometries, but now we also need to do it for timestamps. Since there could be arbitrarily many timestamps (or geometries, for that matter) in a single row, the code is tidier if we just use register_type to do it automatically.

Also - added tests, and made some test fixtures - edit_points, edit_polygons, edit_table - work for postgis as well as for gpkg.

## Related links:

### DIFF
--- a/sno/db_util.py
+++ b/sno/db_util.py
@@ -1,0 +1,73 @@
+from psycopg2.sql import Identifier, SQL
+
+from . import gpkg
+
+
+# Utilities for dealing with different database drivers.
+
+
+def execute_insert_dict(dbcur, table, record, gpkg_funcs=None, pg_funcs=None):
+    """
+    Insert the given record into the given table.
+    dbcur - Database cursor.
+    table - the name of a table, or fully qualified name in the form f"{schema}.{name}"
+    record - the record to insert, as a dict keyed by column name.
+    gpkg_funcs, pg_funcs - extra functions needed to adapt the types to fit the columns.
+    """
+    if is_postgis(dbcur):
+        # Table can be a fully-qualified name, including the schema.
+        sql = SQL("INSERT INTO {} ({}) VALUES ({});").format(
+            Identifier(*table.split(".")),
+            SQL(",").join([Identifier(k) for k in record]),
+            SQL(",").join(_postgis_placeholders(record, pg_funcs)),
+        )
+    else:
+        sql = f"""
+        INSERT INTO {table}
+            ({','.join([gpkg.ident(k) for k in record])})
+        VALUES
+            ({','.join(_gpkg_placeholders(record, gpkg_funcs))});
+        """
+
+    dbcur.execute(sql, tuple(record.values()))
+    return changes_rowcount(dbcur)
+
+
+def _gpkg_placeholders(record, gpkg_funcs=None):
+    """
+    Returns ['?', '?', '?', ...] - where the nunber of '?' returned is len(record).
+    gpkg_funcs - a dict keyed by index to override some of the placeholders, eg:
+        {1: "GeomFromEWKT(?)"}
+    """
+    result = ["?"] * len(record)
+    if gpkg_funcs:
+        for index, placeholder in gpkg_funcs.items():
+            result[index] = placeholder
+    return result
+
+
+def _postgis_placeholders(record, pg_funcs=None):
+    """
+    Returns ['%s', '%s', '%s', ...] where the number of '%s' returned is len(record).
+    pg_funcs - a dict keyed by index to override some of the placeholders, eg:
+        {1: "SetSRID(%s, 4326)"}
+    """
+
+    result = [SQL("%s")] * len(record)
+    if pg_funcs:
+        for index, placeholder in pg_funcs.items():
+            result[index] = SQL(placeholder)
+    return result
+
+
+def changes_rowcount(dbcur):
+    """Returns the number of rows that the last .execute*() affected."""
+    # PEP 0249 specifies .rowcount, but it not everyone support it:
+    if hasattr(dbcur, "rowcount"):
+        return dbcur.rowcount
+    # Otherwise, this works for GPKG / sqlite:
+    return dbcur.getconnection().changes()
+
+
+def is_postgis(dbcur):
+    return type(dbcur).__module__.startswith("psycopg2")
--- a/sno/working_copy/postgis.py
+++ b/sno/working_copy/postgis.py
@@ -6,16 +6,16 @@ from urllib.parse import urlsplit, urlunsplit
 
 import pygit2
 import psycopg2
-from psycopg2.extensions import register_adapter, Binary
+from psycopg2.extensions import new_type, register_adapter, register_type, Binary
 from psycopg2.extras import DictCursor
 from psycopg2.sql import Identifier, SQL
 
 
 from .base import WorkingCopy
 from sno import geometry, crs_util
+from sno.db_util import changes_rowcount
 from sno.diff_structs import DeltaDiff
 from sno.filter_util import UNFILTERED
-from sno.geometry import gpkg_geom_to_ewkb
 
 """
 * database needs to exist
@@ -26,18 +26,38 @@ from sno.geometry import gpkg_geom_to_ewkb
     3. create triggers
 """
 
-
-def geometry_ewkb_adapter(g):
-    return Binary(gpkg_geom_to_ewkb(g))
+L = logging.getLogger("sno.working_copy.postgis")
 
 
-register_adapter(geometry.Geometry, geometry_ewkb_adapter)
+def _adapt_geometry_to_db(g):
+    return Binary(geometry.gpkg_geom_to_ewkb(g))
+
+
+register_adapter(geometry.Geometry, _adapt_geometry_to_db)
+
+
+def adapt_geometry_from_db(g, dbcur):
+    return geometry.hexewkb_to_gpkg_geom(g)
+
+
+def adapt_timestamp_from_db(t, dbcur):
+    # TODO - revisit timezones.
+    if isinstance(t, str):
+        # This makes postgis timestamps behave more like GPKG ones.
+        return t.replace(" ", "T") + "Z"
+    return t
+
+
+# See https://github.com/psycopg/psycopg2/blob/master/psycopg/typecast_builtins.c
+TIMESTAMP_OID = 1114
+TIMESTAMP = new_type((TIMESTAMP_OID,), "TIMESTAMP", adapt_timestamp_from_db)
+psycopg2.extensions.register_type(TIMESTAMP)
 
 
 class WorkingCopy_Postgis(WorkingCopy):
     def __init__(self, repo, uri):
         """
-        uri: connection string of the form postgresql://[user[:password]@][netloc][:port][/dbname[/schema]][?param1=value1&...]
+        uri: connection string of the form postgresql://[user[:password]@][netloc][:port][/dbname/schema][?param1=value1&...]
         """
         self.L = logging.getLogger(self.__class__.__qualname__)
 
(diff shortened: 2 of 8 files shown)


sno/db_util.py
```diff
@@ -0,0 +1,61 @@
...
+        VALUES
+            ({','.join(_gpkg_placeholders(record, gpkg_funcs))});
+        """
+
+    dbcur.execute(sql, tuple(record.values()))
+    return rowcount(dbcur)
+
+
+def _gpkg_placeholders(record, gpkg_funcs=None):
+    """Returns '?,?,?,?...' - where the nunber of ? returned is len(record)"""
```
craigds, line 29: docstring says it returns a string but it actually returns an array

sno/db_util.py
```diff
@@ -0,0 +1,61 @@
+from psycopg2.sql import Identifier, SQL
```
craigds, line 1: this whole file feels wrong to me, feels like it's going to grow unholy quickly. Should we look into using a proper multi-database abstraction like sqlalchemy? 

We probably don't want or need an ORM, just the SQL implementation abstraction: https://docs.sqlalchemy.org/en/13/core/tutorial.html . Couple with some [table reflection](https://docs.sqlalchemy.org/en/13/core/reflection.html) for auto-creating the Table instances for user-defined tables)

One minor thing to note with using sqlalchemy is that it doesn't work with the apsw driver (rogerbinns/apsw#107) but I'm not clear if there's any reason we're using apsw to begin with instead of the sqlite module in the stdlib? (cc @rcoup) Feels like switching to something more DBAPI-compliant would make interaction with multiple databases simpler even if we didn't use sqlalchemy.

rcoup: I mentioned this in https://github.com/koordinates/sno/pull/270#issuecomment-703492688 too.

>  I'm not clear if there's any reason we're using apsw to begin with instead of the sqlite module in the stdlib?

1. The sqlite module in stdlib doesn't support loading extensions in platform/version-specific ways.
2. I wanted consistency on sqlite versions across python+spatialite+gdal+proj
3. pysqlite has been abandoned as a standalone project
4. So we'd need to build _Python_ per-platform to get the right build options and an up to date sqlite version
5. And that seemed much harder than swapping to apsw
6. pysqlite's transaction behaviour is abominable ⛄

rcoup: > Should we look into using a proper multi-database abstraction like sqlalchemy?

No objection in principle, seems better than hand-rolling string-substitution & type functions forever & ever

rcoup: aha, there is a https://github.com/coleifer/pysqlite3 now!

rcoup:
This would be _vastly_ easier to review if the code moves were in separate commits from behaviour changes. I understand that it's not feasible to redo here, but for future PRs...

sno/working_copy/postgis.py
```diff
@@ -120,6 +141,20 @@ def session(self, bulk=0):
...
                 L.debug("session: new/done")
 
+    def _register_types(self, db):
+        dbcur = db.cursor()
+        dbcur.execute("SELECT NULL::geometry")
+        geom_oid = dbcur.description[0][1]
+        GEOMETRY = psycopg2.extensions.new_type((geom_oid,), "GEOMETRY", cast_geometry)
+        psycopg2.extensions.register_type(GEOMETRY)
+
+        dbcur.execute("SELECT NULL::timestamp")
```
rcoup, line 151: technically these are all static values for builtin types: https://github.com/psycopg/psycopg2/blob/master/psycopg/typecast_builtins.c

(and there are a couple of flavours of timestamp, which one do we want here?)

tests/test_diff.py
```diff
@@ -88,7 +88,7 @@ def test_diff_points(
                 "+                                     geom = POINT(...)",
                 "+                                  t50_fid = 9999999",
                 "+                               name_ascii = Te Motu-a-kore",
-                "+                               macronated = 0",
+                "+                               macronated = N",
```
rcoup, line 91: did this change types?

sno/working_copy/postgis.py
```diff
@@ -51,19 +65,24 @@ def __init__(self, repo, uri):
...
-            if len(path_parts) > 2:
-                raise ValueError("Expecting postgresql://[host]/dbname/schema")
-            elif len(path_parts) == 2:
-                self.schema = path_parts[1]
-                url_path = f"/{path_parts[0]}"
-            else:
-                self.schema = self._default_schema(repo)
+        path_parts = url_path[1:].split("/", 3) if url_path else []
+        if len(path_parts) != 2:
+            raise ValueError("Expecting postgresql://[host]/dbname/schema")
```
rcoup, line 70: As a user, it's normal to provide [options](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6) in the url for timeouts, application names, tls, etc

sno/working_copy/postgis.py
```diff
@@ -51,19 +65,24 @@ def __init__(self, repo, uri):
...
+        self.schema = path_parts[1]
+
+        url_path = f"/{path_parts[0]}"
+        url_query = f"options=--search_path%3D{self.schema},public"
```
rcoup, line 74: there could be lots of schemas in the search path, defined on a user + db level. PostGIS and its functions could be in a separate schema, etc.

I think the safest is for us to refer to fully-qualified table paths everywhere and don't touch `search_path`

sno/working_copy/postgis.py
rcoup: we should probably always set `fallback_application_name=sno`, makes it much easier to see what's going on in the DB via the status queries if there isn't a sno-specific DB user.

sno/db_util.py
rcoup: and https://github.com/hoehrmann/sqlalchemy-pysqlite3

sno/db_util.py
olsen232: Done

sno/db_util.py
olsen232: Discussed offline - punted until after next week

sno/db_util.py
```diff
@@ -0,0 +1,61 @@
...
+    return result
+
+
+def rowcount(dbcur):
+    """Returns the number of rows that the last .execute*() produced or affected."""
+    # PEP 0249 specifies .rowcount, but it not everyone support it:
+    if hasattr(dbcur, "rowcount"):
+        return dbcur.rowcount
+    # Otherwise, this works for GPKG / sqlite:
+    return dbcur.getconnection().changes()
```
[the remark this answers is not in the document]
olsen232: Changed function to `changes_rowcount`

sno/working_copy/postgis.py
```diff
@@ -26,14 +26,28 @@
...
 def geometry_ewkb_adapter(g):
-    return Binary(gpkg_geom_to_ewkb(g))
+    return Binary(geometry.gpkg_geom_to_ewkb(g))
 
 
 register_adapter(geometry.Geometry, geometry_ewkb_adapter)
 
 
+def cast_geometry(g, dbcur):
+    return geometry.hexewkb_to_gpkg_geom(g)
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/postgis.py
```diff
@@ -120,6 +141,20 @@ def session(self, bulk=0):
                 del self._db
                 L.debug("session: new/done")
 
+    def _register_types(self, db):
+        dbcur = db.cursor()
+        dbcur.execute("SELECT NULL::geometry")
```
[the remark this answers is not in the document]
olsen232: Fixed - this code is replaced with:
`GEOMETRY_OID = 16393`
which doesn't fail even if there is no type with that OID.

sno/working_copy/postgis.py
olsen232: No idea why in postgres' datetime type is called TIMESTAMP, but that builtins.c calls it DATETIME.
Nevertheless, chose to apply this adapter-function to the non-timezone flavour (oid=1114) and the code works - since that is the only type i'm creating right now - GPKG DATETIMES are equivalent to Postgis TIMESTAMP, and right now there's no way to create TIMESTAMPTZs. Still have a TODO to fix this up later.

sno/working_copy/postgis.py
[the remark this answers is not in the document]
rcoup: extension type OIDs aren't fixed. Like, it'll be different on another DB.

db-1
```
dandelion_data=# select oid,* from pg_type where typname = 'geometry';
-[ RECORD 1 ]--+--------------------
oid            | 19178
typname        | geometry
typnamespace   | 2200
```

db-2
```
doubtful2_web=> select oid,* from pg_type where typname = 'geometry';
-[ RECORD 1 ]--+--------------------
oid            | 369573
typname        | geometry
typnamespace   | 2200
```

tests/test_diff.py
olsen232: It did not - it has always been a VARCHAR(1), and in the test data, it is filled with only "N" and "Y".
The test-helper tries to insert a `False` into this column, which in GPKG land, works okay - it gets converted to "0". But in Postgis land it wasn't happy.
I could have split the test-helper into two variants so we put False into it in GPKG, and "N" in Postgis, but it seemed better just to fix the test-helper so it doesn't rely on auto type conversion.

sno/working_copy/postgis.py
olsen232: Fair enough: this was a hack to avoid having to specify the schema in the test-helper functions. But, it doesn't make sense to mess up the production code for that reason only

sno/working_copy/postgis.py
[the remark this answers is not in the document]
rcoup: maybe some/both of these?

```sql
SELECT extname, extversion FROM pg_extension WHERE extname='postgis';
SELECT oid FROM pg_type WHERE typtype='b' AND typecategory = 'U' and typname='geometry';
```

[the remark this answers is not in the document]
rcoup: `SELECT 'geometry'::regtype::oid` might be simpler still 😄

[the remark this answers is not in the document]
olsen232: Brought back the query to find the geometry type, made sure it works when no geometry type is present.
Still using 'TIMESTAMP_OID=1114` for timestamp type

sno/working_copy/postgis.py
olsen232: Done

craigds (approved):

sno/working_copy/postgis.py
```diff
@@ -572,3 +599,58 @@ def _execute_diff_query(self, dbcur, dataset, feature_filter=None):
...
+        )
+        dbcur.execute(sql_changed, (dataset.table_name,))
+
+    def reset_tracking_table(self, reset_filter=UNFILTERED):
+        reset_filter = reset_filter or UNFILTERED
+
+        with self.session() as db:
+            dbcur = db.cursor()
+            if reset_filter == UNFILTERED:
+                dbcur.execute(SQL("DELETE FROM {};").format(self.TRACKING_TABLE))
```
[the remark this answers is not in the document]
olsen232: Seems like a bit of a premature optimisation - we would only use TRUNCATE in this one case, other cases would still use DELETE, TRUNCATE requires a different permission (probably most users will have both permissions or neither so normally won't matter), and TRUNCATE locks the table where DELETE doesn't (it's probably still more efficient, but haven't checked). So, leaving as-is

sno/db_util.py
```diff
@@ -0,0 +1,78 @@
...
+    Insert the given record into the given table.
+    dbcur - Database cursor.
+    table - the name of a table, or fully qualified name in the form f"{schema}.{name}"
+    record - the record to insert, as a dict keyed by column name.
+    gpkg_funcs, pg_funcs - extra functions needed to adapt the types to fit the columns.
+    """
+    if is_postgis(dbcur):
+        # Table can be a fully-qualified name, including the schema.
+        if "." in table:
+            table = SQL(".").join(Identifier(t) for t in table.split("."))
```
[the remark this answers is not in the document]
olsen232: Done

PR #274, opened 2020-10-13
Support HEAD<>PostgisWC diff for schema changes.

olsen232:
## Description

`sno diff` and `sno status` show changes made to the table schema and changes made to the CRS definitions.

For example:
`$ sno diff`
```
+++ ice_seed:meta:crs/EPSG:2113.wkt
+ "PROJCS[\"NZGD2000 / Wellington 2000\",GEOGCS[\"NZGD2000\",DATUM[\"New_Zealand_Geodetic_Datum_2000\",SPHEROID[\"GRS 1980\",6378137,298.257222101,AUTHORITY[\"EPSG\",\"7019\"]],TOWGS84[0,0,0,0,0,0,0],AUTHORITY[\"EPSG\",\"6167\"]],PRIMEM[\"Greenwich\",0,AUTHORITY[\"EPSG\",\"8901\"]],UNIT[\"degree\",0.01745329251994328,AUTHORITY[\"EPSG\",\"9122\"]],AUTHORITY[\"EPSG\",\"4167\"]],UNIT[\"metre\",1,AUTHORITY[\"EPSG\",\"9001\"]],PROJECTION[\"Transverse_Mercator\"],PARAMETER[\"latitude_of_origin\",-41.3011111111111],PARAMETER[\"central_meridian\",174.7763888888889],PARAMETER[\"scale_factor\",1],PARAMETER[\"false_easting\",400000],PARAMETER[\"false_northing\",800000],AUTHORITY[\"EPSG\",\"2113\"],AXIS[\"Easting\",EAST],AXIS[\"Northing\",NORTH]]"
--- ice_seed:meta:crs/EPSG:4326.wkt
- "GEOGCS[\"WGS 84\",DATUM[\"WGS_1984\",SPHEROID[\"WGS 84\",6378137,298.257223563,AUTHORITY[\"EPSG\",\"7030\"]],AUTHORITY[\"EPSG\",\"6326\"]],PRIMEM[\"Greenwich\",0,AUTHORITY[\"EPSG\",\"8901\"]],UNIT[\"degree\",0.0174532925199433,AUTHORITY[\"EPSG\",\"9122\"]],AUTHORITY[\"EPSG\",\"4326\"]]"
--- ice_seed:meta:schema.json
+++ ice_seed:meta:schema.json
  [
    {
      "id": "f450aecc-6796-47ee-a413-cd7ffbfaeb5e",
      "name": "fid",
      "dataType": "integer",
      "primaryKeyIndex": 0,
      "size": 64
    },
    {
      "id": "d7cbf9c0-90d8-49bb-938b-339d793f6add",
      "name": "geom",
      "dataType": "geometry",
      "primaryKeyIndex": null,
      "geometryType": "POINT",
-     "geometryCRS": "EPSG:4326",
+     "geometryCRS": "EPSG:2113",
    },
    {
      "id": "8c14a92e-f0d7-40cb-91ed-aa90e8ffbdfb",
      "name": "crt_date",
      "dataType": "float",
      "primaryKeyIndex": null,
      "size": 64
    },
    {
      "id": "0deab354-ec25-4012-bc4c-5d89de17f51f",
      "name": "mod_date",
      "dataType": "float",
      "primaryKeyIndex": null,
      "size": 64
    },
  ]
```

## Related links:

## Checklist:

No update to CHANGELOG yet.

### DIFF
--- a/sno/working_copy/postgis.py
+++ b/sno/working_copy/postgis.py
@@ -249,52 +249,10 @@ class WorkingCopy_Postgis(WorkingCopy):
             dbcur.execute(SQL("DROP SCHEMA {} CASCADE").format(Identifier(self.schema)))
 
     def write_meta(self, dataset):
-        """
-        Populate the following postgis tables with data from this dataset:
-        geometry_columns, spatial_ref_sys
-        """
-        table = dataset.table_name
-        with self.session() as db:
-            dbcur = db.cursor()
-            dbcur.execute(
-                SQL(
-                    "DELETE FROM {} WHERE f_table_schema=%s AND f_table_name=%s;"
-                ).format(self._table_identifier("geometry_columns")),
-                (self.schema, table),
-            )
-
-            geom_columns = postgis_adapter.generate_postgis_geometry_columns(
-                dataset, self.schema
-            )
-            for row in geom_columns:
-                dbcur.execute(
-                    SQL("INSERT INTO {} ({}) VALUES ({});").format(
-                        self._table_identifier("geometry_columns"),
-                        SQL(",").join([Identifier(k) for k in row]),
-                        SQL(",").join([SQL("%s")] * len(row)),
-                    ),
-                    tuple(row.values()),
-                )
-
-            spatial_refs = postgis_adapter.generate_postgis_spatial_ref_sys(dataset)
-            for row in spatial_refs:
-                dbcur.execute(
-                    SQL(
-                        """
-                        INSERT INTO {} ({}) VALUES ({})
-                        ON CONFLICT (srid) DO UPDATE
-                        SET auth_name=EXCLUDED.auth_name,
-                            auth_srid=EXCLUDED.auth_srid,
-                            srtext=EXCLUDED.srtext,
-                            proj4text=EXCLUDED.proj4text;
-                    """
-                    ).format(
-                        self._table_identifier("spatial_ref_sys"),
-                        SQL(",").join([Identifier(k) for k in row]),
-                        SQL(",").join([SQL("%s")] * len(row)),
-                    ),
-                    tuple(row.values()),
-                )
+        # TODO - we should be writing any custom CRSs to public.spatial_ref_sys
+        # Not entirely straight-forward - what if the supplied CRS is a slightly modified EPSG definition?
+        # Should we overwrite the data in the table?
+        pass
 
     def _create_spatial_index(self, dataset):
         L = logging.getLogger(f"{self.__class__.__qualname__}._create_spatial_index")
@@ -392,30 +350,6 @@ class WorkingCopy_Postgis(WorkingCopy):
 
             dbcur.execute(
                 SQL("CREATE SCHEMA IF NOT EXISTS {};").format(Identifier(self.schema))
-            )
-            dbcur.execute(
-                SQL(
-                    """CREATE TABLE IF NOT EXISTS {} (
-                          f_table_catalog    VARCHAR(256) NOT NULL,
-                          f_table_schema     VARCHAR(256) NOT NULL,
-                          f_table_name       VARCHAR(256) NOT NULL,
-                          f_geometry_column  VARCHAR(256) NOT NULL,
-                          coord_dimension    INTEGER NOT NULL,
-                          srid               INTEGER NOT NULL,
-                          type               VARCHAR(30) NOT NULL
-                    );"""
-                ).format(self._table_identifier("geometry_columns"))
-            )
-            dbcur.execute(
-                SQL(
-                    """CREATE TABLE IF NOT EXISTS {} (
-                        srid       INTEGER NOT NULL PRIMARY KEY,
-                        auth_name  VARCHAR(256),
-                        auth_srid  INTEGER,
-                        srtext     VARCHAR(2048),
-                        proj4text  VARCHAR(2048)
-                    );"""
-                ).format(self._table_identifier("spatial_ref_sys"))
             )
 
             for dataset in datasets:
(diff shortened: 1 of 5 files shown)


rcoup:
```
+++ ice_seed:meta:crs/EPSG:2113.wkt
+ "PROJCS[\"NZGD2000...
--- ice_seed:meta:crs/EPSG:4326.wkt
- "GEOGCS[\"WGS 84\",DATUM...
```

Are these actually double-quoted in the WKT files, or is it a representation issue here?

If so, that's a bug — the wkt file should start with `PROJCS` or `GEOGCS`, not `"`

sno/working_copy/postgis.py
```diff
@@ -336,14 +249,52 @@ def delete(self):
...
-    def save_config(self, **kwargs):
-        self.repo.config["sno.workingcopy.version"] = 1
-        self.repo.config["sno.workingcopy.path"] = self.uri
+            spatial_refs = postgis_adapter.generate_postgis_spatial_ref_sys(dataset)
+            for row in spatial_refs:
+                dbcur.execute(
+                    SQL(
+                        """
+                        INSERT INTO {} ({}) VALUES ({})
+                        ON CONFLICT (srid) DO UPDATE
```
rcoup, line 285: I think we might need to FIXME this... If some repo has a messed up definition of a CRS code, should it really be able to override the one that PostGIS has installed, potentially messing up all transforms happening across the DB?

Maybe for now we warn about what we're overwriting?

sno/working_copy/postgis.py
```diff
@@ -438,38 +389,59 @@ def write_full(self, commit, *datasets, **kwargs):
...
+                    """CREATE TABLE IF NOT EXISTS {} (
+                          f_table_catalog    VARCHAR(256) NOT NULL,
+                          f_table_schema     VARCHAR(256) NOT NULL,
+                          f_table_name       VARCHAR(256) NOT NULL,
+                          f_geometry_column  VARCHAR(256) NOT NULL,
+                          coord_dimension    INTEGER NOT NULL,
+                          srid               INTEGER NOT NULL,
+                          type               VARCHAR(30) NOT NULL
+                    );"""
+                ).format(self._table_identifier("geometry_columns"))
```
rcoup, line 407: these aren't user-created tables. They're created by Postgis when the extension is installed, and populated by PostGIS:
- `geometry_columns`: this is a _view_ over the column/table definitions, so changes whenever a geometry column is created/altered.
- `spatial_ref_sys` is populated by PostGIS with the EPSG database when the extension is installed, but is user-updatable.

sno/working_copy/postgis_adapter.py
```diff
@@ -0,0 +1,280 @@
...
+    pg_col_info, pg_geometry_columns, pg_spatial_ref_sys, id_salt
+):
+    """
+    Given the postgis column info for a particular column, and some extra context in
+    case it is a geometry column, converts it to a ColumnSchema. The extra context will
+    only be used if the given pg_col_info is the geometry column.
+    Parameters:
+    pg_col_info - a single column from pg_table_info.
+    pg_geometry_columns - contents of the "geometry_columns" table.
+    pg_spatial_ref_sys - contents of the "spatial_ref_sys" table.
```
rcoup, line 142: this normally has thousands of rows in it... do we need to extract/iterate through it vs doing a `SELECT...WHERE`?

sno/working_copy/postgis_adapter.py
```diff
@@ -0,0 +1,280 @@
...
+    Returns (geometry-type-without-suffix, suffix) if geometry-type ends with suffix.
+    Otherwise just returns (geometry-type, "")
+    """
+    if geometry_type.endswith(suffix):
+        return geometry_type[:-1], suffix
+    else:
+        return geometry_type, ""
+
+
+def generate_postgis_geometry_columns(v2_obj, postgis_schema):
```
rcoup, line 222: we need to update the column definitions via `ALTER TABLE`, not update `geometry_columns`

sno/working_copy/postgis_adapter.py
rcoup: similarly (though less likely) we have some databases with tens of thousands of rows in `geometry_columns` too...

olsen232: The docstring was actually wrong - we are doing a `SELECT...WHERE` and these only contained the geometry columns or spatial reference systems for the given dataset.

Slightly modified - pg_geometry_columns is no longer passed as an extra table, but is instead joined to the pg_col_info as a couple of extra fields.

sno/working_copy/postgis_adapter.py
olsen232: Done (as much as is required for this change).

The scope of this change only supports diffs and commits - it doesn't yet support resets. So, the ALTER TABLE only happens when the user changes the table, or in a test. So, I haven't exactly "done" this comment, but I will in the reset PR.

On the other hand, I no longer create my own `geometry_columns` table - the code now only ever reads `public.geometry_columns` and never writes to it. So, that is fixed.

sno/working_copy/postgis.py
olsen232: For now, I've just stopped writing CRS's to the `spatial_ref_sys` table. The tests still pass: everything works as expected as long as no dataset contains any custom CRS that the `spatial_ref_sys` doesn't know about, and I don't yet have any postgres tests that try to set a custom CRS. I'll do it in a future PR, I've added a TODO.

We have some options - for instance, we could decide never to overwrite anything in the range 0-9999. If our CRS would overwrite one of those, we could add 1000000 to it (or whatever) - so creating our own range of "CRS that have an authority ID in the range 0-9999 but don't match Postgis definitions", with SRIDs in the range 1000000 to 1009999

Of course, we could still have collisions within that namespace too.... unless we use a larger range based on a hash of the CRS.

And we could do some work on deciding whether or not CRS definitions match and so if they both need a separate ID, or if they can share an ID in spite of minor differences. Eg using the SpatialRef functions for finding matching definitions, or by comparing proj4 text.

So many options...

sno/working_copy/postgis.py
olsen232: Oops! You are absolutely right. Much of this change could be salvaged, but I've made some major alterations. I stopped  creating and writing to either table. Instead I just read from the ones that are already maintained. This works fine unless there is a custom CRS that doesn't match anything in the `spatial_ref_sys` table.

olsen232:
Fixed WKT output in master - this was a holdover from when meta items were more JSON everywhere.

sno/working_copy/postgis.py
rcoup: So the EPSG database is (by design) not mutable — if a coordinate transform  parameter is wrong they issue a new number and deprecate the old one. (names/areas-of-use/etc might change, but not the actual math parameters)

I'd suggest as a reasonably pragmatic first step:

1. INSERT missing CRS entries into `spatial_ref_sys`
2. Don't update existing CRS entries if they exist already

Call it done there, and we can revisit further down the line.

I guess this is the sort of thing a `fsck` command could check in detail?

rcoup: with the above approach, the solution for something weird happening for an end user becomes either:

1. delete the row from `spatial_ref_sys` and Sno will fix it
2. add a row to `spatial_ref_sys` with the correct settings and Sno will respect it

sno/working_copy/postgis.py
rcoup: change not pushed?

olsen232: Remembered to push, forgot to commit...

PR #275, opened 2020-10-15
Postgis: Support for resets involving meta changes

olsen232:
## Description

Support for `sno checkout` between two commits where any of the following could have changed:
- dataset schemas
- CRS definitions
- other metadata not supported by Postgis (eg title, description).

This PR handles inserting CRS definitions if they are missing (but it never updates or deletes existing ones).

This PR has some optimisations so that schema changes don't always cause the entire table to be recreated.
These changes are done using single SQL commands:

ALTER TABLE x DROP COLUMN;
ALTER TABLE x RENAME COLUMN y TO z;
ALTER TABLE x ALTER COLUMN y TYPE z;

But, if columns have been inserted or reordered, the entire table will be rewritten. There is no Postgres command for reordering columns, and if a column has been inserted, we may need to populate it from the dataset anyway - so there is no single SQL command that will take care of everything, and any attempt to optimise may or may not be any quicker.

## Related links:

### DIFF
--- a/sno/working_copy/postgis.py
+++ b/sno/working_copy/postgis.py
@@ -16,6 +16,8 @@ from . import postgis_adapter
 from sno import geometry, crs_util
 from sno.db_util import changes_rowcount
 from sno.filter_util import UNFILTERED
+from sno.schema import Schema
+
 
 """
 * database needs to exist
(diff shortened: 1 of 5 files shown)


sno/working_copy/postgis.py
```diff
@@ -471,6 +494,26 @@ def delete_features(self, dbcur, dataset, pk_iter):
 
         return feat_count
 
+    def drop_table(self, target_tree_or_commit, *datasets):
```
craigds, line 497: I just realised that this is called by `reset()` without using a transaction. So the table drop gets committed before the table is recreated...

`reset()` should wrap all its changes in a `self.session()` to avoid the drop being exposed to other transactions until the table's been recreated/repopulated

sno/working_copy/postgis.py
```diff
@@ -609,3 +652,96 @@ def _update_state_table_tree_impl(self, dbcur, tree_id):
...
+        return sum(dt.values()) == 0
+
+    def _apply_meta_title(self, dataset, src_value, dest_value, dbcur):
+        pass  # This is a no-op for postgis
+
+    def _apply_meta_description(self, dataset, src_value, dest_value, dbcur):
+        pass  # This is a no-op for postgis
+
+    def _apply_meta_metadata_dataset_json(self, dataset, src_value, dest_value, dbcur):
+        pass  # This is a no-op for postgis
```
craigds, line 690: these will be the norm - most other databases won't support these either. Move these to base.py?

craigds (approved):

sno/working_copy/postgis.py
rcoup: Basically _every_ sno command needs to be wrapped in a single transaction. Firstly so changes either fully apply or don't at all; second for a consistent view of the database (dataset tables are consistent between queries when concurrent updates are happening); third to make sure other connections has a consistent view of the database (ie: Sno's changes are all visible or none)

sno/working_copy/postgis.py
rcoup: Still think we should use `COMMENT ON TABLE {mydataset} IS 'the dataset title';` 😄

rcoup (changes requested):
Need to get transaction handling right, after that GTG 😄 

Can we write some tests for the transaction handling so we can be assured it is & stays right?

sno/working_copy/postgis.py
olsen232: Done

sno/working_copy/postgis.py
olsen232: Not done - wrapping the reset() in a single session breaks GPKG support :(. I'll fix this in a separate PR.

PR #279, opened 2020-10-19
Don't rely on SpatialReference for parsing authority

olsen232:
## Description

The osgeo.osr SpatialReference class is very fiddly to use well for simple tasks. For instance, the user supplies some WKT. We want to parse the authority from it and store it at a path based on its authority. SpatialReference can do this - but whether it successfully returns the authority it parsed, or None, or an error, depends on many things:

- is osgeo configured so that it can find the proj4 database 
- does osgeo think the WKT matches the definition from the database with that authority
- are the axes defined in the same order as the definition in the database
- how is the SpatialReference initialised - eg SpatialReference(wkt) or SpatialReference.SetFromUserInput(wkt)
- how is SetAxisMappingStrategy configured (maybe?)

This is a bit too much to worry about when we just want to be told what authority is written inside the WKT, and it makes testing difficult. Finding all these conditions took me some time and I am still not sure what the perfect configuration for everything is, or if there are tradeoffs. So, I have switched to extracting using regexes - also more efficient.

## Related links:

https://github.com/koordinates/sno/issues/278

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
  * `data ls` now accepts an optional ref argument
  * `meta get` now accepts a `--ref=REF` option
  * `clone` now accepts a `--branch` option to clone a specific branch.
+ * Bugfix - don't drop the user-supplied authority from the supplied CRS and generate a new unrelated one. [#278](https://github.com/koordinates/sno/issues/278)
 
 ## 0.5.0
 
--- a/sno/crs_util.py
+++ b/sno/crs_util.py
@@ -1,3 +1,5 @@
+import re
+
 from osgeo.osr import SpatialReference
 
 from .cli_util import StringFromFile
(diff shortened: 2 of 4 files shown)


craigds (approved):

rcoup:
Which of your bullet points actually apply (ie. we can reproduce) and which need to be filed as bugs with GDAL?

(1) if this isn't set right nothing will work
(4) seems like a clear bug if we can reproduce?

(3) & (5) are maybe the same thing?

olsen232:
1 was trickier than you'd expect. If SpatialReference didn't work properly when not configured, I probably would've figured this bit out sooner. However, in this case, when all we are trying to do is read the authority marker from a CRS definition, it works better when not configured - so if I fired up a debugger and failed to configure GDAL properly, everything would work perfectly. As above, I'm now using regular expressions for safety's sake, but a different strategy would be to *deconfigure* GDAL before trying to read the authority marker - a deconfigured instance will report what it reads, but a configured instance is opinionated and might refuse to return the authority in the definition.

4. I'll put together a test case next week.

3 and 5 are at least related concepts, not sure if they are exactly the same thing or not. Maybe setting SetAxisMappingStrategy to something different can help here, but there's only one call to SetAxisMappingStrategy in our codebase right now, and I tried using that same call but didn't achieve anything at all, so I don't know if it helps us here or not.

rcoup:
> 1. was trickier than you'd expect.

`import sno` configures gdal/everything...

olsen232:
Yup, configuring it properly wasn't the tricky part - the tricky part was that if I fired up a debugger and tested out how SpatialReference worked without importing sno, it would work better for this use-case than the properly configured one in sno, which I didn't expect... and this one just one more thing that I could get confused about as I tried to figure out what GDAL does and doesn't do

PR #281, opened 2020-10-21
Open a single db transaction for a single operation

olsen232:
## Description

Went through all the functions in the working copies and made sure they don't open more than one transaction in a sequence, or call more than one function in a sequence that would result in opening more than one transaction in a sequence.

This can mostly be easily fixed by opening a transaction that is reused, and that contains all the functions called. Any transactions that those functions open will reuse the one ongoing transaction.

A couple of things needed changing - 
- we can't switch database engine mid-transaction to create the spatial index using gdal. Luckily, it seems like reusing the apsw engine works? Is there a downside?
- Rearranged some DELETEs so they didn't need `PRAGMA defer_foreign_keys`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 ## 0.7.0 (UNRELEASED)
 
  * Support for checking out a dataset with a string primary key (or other non-integer primary key) as a GPKG working copy [#307](https://github.com/koordinates/sno/issues/307)
+ * Improved error recovery: Sno commands now write to the working copy within a single transaction, which is rolled back if the command fails. [#281](https://github.com/koordinates/sno/pull/281)
  * Bugfix - `sno meta set` didn't allow updates to `schema.json`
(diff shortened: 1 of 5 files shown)


rcoup:
So I think we're effectively swapping from GDALs spatial index to Spatialite's here. 

I want to dig in to figure out what that actually means in practice wrt differences.

olsen232:
... can we use the GDAL engine everywhere??

No idea how this ties in with sqlalchemy as and when we get that working

olsen232:
@rcoup is this something you are still looking into?

rcoup:
AFAICS [GDAL GPKG Spatial Index code](https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagetablelayer.cpp#L2971-L3193) and the [Spatialite equivalent](https://www.gaia-gis.it/fossil/libspatialite/artifact/37d793474882116a) basically seem to do the same thing. Which I guess is because it's defined as an [official GPKG extension](http://www.geopackage.org/spec121/#extension_rtree).

👍

PR #285, opened 2020-10-22
No merge renames

olsen232:
## Description

Two commits - 
first commit upgrades pygit2 to 1.3.0 to have this functionality.
second commit turns off `find_renames` during a merge. Find renames is not useful to us all during a merge for a couple of reasons - 

- the sno code for distinguishing and dealing with all the different types of rename conflicts - eg rename/add, rename/delete, rename/edit rename/rename - has never been written
- it is currently not possible to write this code since pygit2 doesn't properly expose rename conflicts - a rename/rename conflict shows up in the conflicts list, but spread into three pieces, and there is no way to reassemble it.

To make things as sensible as possible given the constraints above, the best thing to do for now is turn off rename detection. That way, a rename/rename conflict is not a conflict it all - it is a delete/delete "conflict" (ie, not a conflict) and then two unrelated adds that also don't conflict. This may not be exactly what the user wanted, but it is better than creating broken partial-conflicts that we can't even reassemble into real conflicts.

Most real world conflicts (ie, not renames which are rare) will continue to work as before.

## Related links:

https://github.com/koordinates/sno/issues/79

### DIFF
--- a/sno/merge.py
+++ b/sno/merge.py
@@ -94,7 +94,7 @@ def do_merge(repo, ff, ff_only, dry_run, commit, commit_message, quiet=False):
         return merge_jdict
 
     tree3 = commit_with_ref3.map(lambda c: c.tree)
-    index = repo.merge_trees(**tree3.as_dict())
+    index = repo.merge_trees(**tree3.as_dict(), flags={"find_renames": False})
 
     if index.conflicts:
         merge_index = MergeIndex.from_pygit2_index(index)
--- a/tests/test_conflicts.py
+++ b/tests/test_conflicts.py
@@ -2,6 +2,7 @@ import json
 import pytest
 
 from sno.merge_util import MergeIndex
+from sno.repo import SnoRepo
 from sno.structs import CommitWithReference
 
 H = pytest.helpers.helpers()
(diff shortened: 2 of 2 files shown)


craigds (commented):
Not sure about this.

If the intent was "Alice and Bob both deleted feature X and added another feature" then this is fine, but in that case hopefully pygit2 wouldn't be detecting renames anyway?

In the other case ("Alice and Bob both changed the PK of feature X to different values") this causes data breakage since now there are two features and no conflicts occured.

I think it'd probably be better to figure out where the bug in pygit/libgit lies and fix it there, and leave this throwing loud errors in sno until then.

@rcoup what's your opinion on this?

olsen232:
This was the workaround I proposed about 6 months ago but never implemented until now, so I forget exactly how the discussion went - and yeah, we don't have to do it - maybe it's better to keep them as conflicts even though it's not clear what the conflicts mean or how they'll be interpreted by the app / the user.

rcoup:
> so I forget exactly how the discussion went

Possible to find it? Would be good to have some more context.

> maybe it's better to keep them as conflicts even though it's not clear what the conflicts mean or how they'll be interpreted by the app / the user.

isn't that desired unless we _know_ what happened and why?

rcoup:
> doesn't properly expose rename conflicts - a rename/rename conflict shows up in the conflicts list, but spread into three pieces, and there is no way to reassemble it.

Can you write this as an example of how it's produced/tagged?

rcoup:
> I think it'd probably be better to figure out where the bug in pygit/libgit lies and fix it there

This should be our approach in general, but if there's a sane workaround in the meantime we should add it, as long as we _do_ file & propose a fix upstream.

olsen232:
An edit/edit conflict looks like this:
```
# Maybe first conflict in a long list of conflicts
{
  "ancestor": {
    "path": "example.txt",
    "contents": "ancestor contents"
  },
  "ours": {
    "path": "example.txt",
    "contents": "ours contents"
  },
  "theirs": {
    "path": "example.txt",
    "contents": "theirs contents"
  },
},
... more conflicts follow
```

Whereas a rename/rename conflict looks like this:
```
# Maybe first conflict in a long list of conflicts
{
  "ancestor": {
    "path": "example.txt",
    "contents": "ancestor contents"
  },
  "ours": null,
  "theirs": null,
},
{
  "ancestor": null,
  "ours": {
    "path": "a.txt",
    "contents": "ours contents"
  },
  "theirs": null,
},
{
  "ancestor": null,
  "ours": null,
  "theirs": {
    "path": "b.txt",
    "contents": "theirs contents"
  },
},
... more conflicts follow
```

Each rename/rename conflict has three parts, and they are not connected together, and they could be interleaved with other disassembled conflicts so there is no way to reassemble them. This is the data that pygit2 exposes.

rcoup:
How does/is this meant to work in libgit2 from a Git perspective?

olsen232:
I think libgit2 has a lower level interface that has the original information, it's just pygit2 doesn't expose it. But, haven't looked into this in enough detail.

We can reassemble some of the conflicts if the contents are an exact match, but not if libgit2 detected an inexact-match rename.

PR #289, opened 2020-10-27
Fix for "unsupported mandatory extension: '.sno'"

olsen232:
## Description

pygit2.Repository.merge_trees for some reason requires access to the git index file - even though it doesn't use it. The file can be missing, this works perfectly.
We don't need the file: we only have a customised index there because it convinces git that it doesn't understand the working-copy part of the repository, so it doesn't try to mess with it.
Just to make it work, the call to merge_trees is now wrapped with a context manager that deletes the index file and then recreates it once the call is finished.

This solution is not ideal since it changes the file-system - multiple simultaneously running sno's could conflict with each other (not that this is a recommended way to use sno anyway), or a crashing sno could leave it in a bad way. Ideally, we would do something more like we do to run `git gc` - we set an environment variable that tells git to use a different index file.

Tests here were broken due to setting repo_version to "1" or "2" but testing if it equals 1 or 2 - the v2 test was never actually run. click.Choice flags only work with strings, which is annoying, and is why repo_version values are sometimes "1" and "2" instead of 1 and 2. This changes all the tests to use integers, which is less surprising, and makes the fixtures more robust to receiving either strings or ints.

### DIFF
--- a/sno/sno_repo.py
+++ b/sno/sno_repo.py
@@ -1,3 +1,4 @@
+import contextlib
 import logging
 import struct
 import subprocess
@@ -287,3 +288,14 @@ class SnoRepo(pygit2.Repository):
             raise InvalidOperation(f'"{dir_path}" isn\'t empty')
         elif not dir_path.exists():
             dir_path.mkdir(parents=True)
+
+    @contextlib.contextmanager
+    def no_locked_index_file(self):
+        try:
+            (self.gitdir_path / SnoRepoFiles.INDEX).unlink()
+        except FileNotFoundError:
+            pass  # Use missing_ok once we have python 3.8
+        try:
+            yield
+        finally:
+            self.lock_git_index()
--- a/sno/merge.py
+++ b/sno/merge.py
@@ -104,7 +104,12 @@ def do_merge(repo, ff, ff_only, dry_run, commit, commit_message, quiet=False):
         return merge_jdict
 
     tree3 = commit_with_ref3.map(lambda c: c.tree)
-    index = repo.merge_trees(**tree3.as_dict())
+
+    with repo.no_locked_index_file():
+        # For no real reason, repo.merge_trees requires access to the index,
+        # and unfortunately, it doesn't respect GIT_INDEX_FILE env variable.
+        # TODO - find a solution that doesn't requrite modifying the file system.
+        index = repo.merge_trees(**tree3.as_dict())
 
     if index.conflicts:
         merge_index = MergeIndex.from_pygit2_index(index)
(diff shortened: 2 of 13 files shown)


sno/sno_repo.py
```diff
@@ -287,3 +288,12 @@ def _ensure_exists_and_empty(dir_path):
             raise InvalidOperation(f'"{dir_path}" isn\'t empty')
         elif not dir_path.exists():
             dir_path.mkdir(parents=True)
+
+    @contextlib.contextmanager
+    def no_locked_index_file(self):
+        if (self.gitdir_path / SnoRepoFiles.INDEX).exists():
+            (self.gitdir_path / SnoRepoFiles.INDEX).unlink()
```
craigds, line 295:
```suggestion
        (self.gitdir_path / SnoRepoFiles.INDEX).unlink(missing_ok=True)
```

craigds (approved):

sno/sno_repo.py
olsen232: Alas - this is python 3.8 only

rcoup:
> pygit2.Repository.merge_trees for some reason requires access to the git index file - even though it doesn't use it.


Why? Can we fix it?

olsen232:
Presumably yes! There are at least a couple of fixes that are better than this workaround

- I dig deeper into the libgit2 C code, become a contributor, stop it from doing this particular bad thing. Then I am a good citizen!
- I modify any created instance of the libgit2::Repository class to be `bare` even though the repo may or may not be `bare` according to the config file. This would mean that it would behave like previously when our repos were always bare, and merge_trees would work. This doesn't fix the underlying libgit bug, but from our point of view, the end result is the same: libgit2 code works and knows that it can't / shouldn't play with the working copy.
- Similar: modify any created instance of the libgit2::Repository class to point at a different index - probably a path that doesn't exist. That's what we do now for git commands eg:

```
$ git gc
Unsupported mandatory extension: '.sno'
$ GIT_INDEX_FILE=.sno/whatever git gc
*git does gc*
```

I'll look into this some more

olsen232:
There is a reason that merge-trees - an operation that seems not to depend on the working copy or index - works for bare repositories or repos with no index, but fails if the index has an unknown extension. The reason is git-attributes: https://git-scm.com/docs/gitattributes. Merging two files at a particular path seems like a stateless pure function, but it actually depends on the merge strategy for that path. The merge strategy for a path is found in git attributes - https://git-scm.com/docs/gitattributes - which are found both in $GIT_DIR/info/attributes, or in various .gitattributes files in the working copy. However, if these files are removed from the working copy, they are still respected if they are still present in the index. Hence the need to check the index to do a merge, or at least any merge containing files where we need to know the merge strategy.

Probably won't mess with libgit2 C code then - I'll look into the other two fixes above.

PR #292, opened 2020-10-29
Allow `sno switch` to switch to new local branch tracking remote branch

olsen232:
## Description

`sno switch x` switches to x if x is a local branch - as before.
`sno switch origin/x` fails as before: you cannot switch to a remote branch
But, `sno switch x` where x is not yet a local branch but is the name of a remote branch, creates a local branch x that tracks `origin/x`. (Just as it does in git)

sno checkout behave similarly, but note that with sno checkout, you CAN `sno checkout origin/x` - this refish is resolved to a particular commit, and that commit is checked out as a "detached HEAD".

## Related links:

Fixes https://github.com/koordinates/sno/issues/259

## Checklist:

### DIFF
--- a/sno/checkout.py
+++ b/sno/checkout.py
@@ -42,7 +42,7 @@ def reset_wc_if_needed(repo, target_tree_or_commit, *, discard_changes=False):
 
 @click.command()
 @click.pass_context
-@click.option("branch", "-b", help="Name for new branch")
+@click.option("new_branch", "-b", help="Name for new branch")
 @click.option(
     "--force",
     "-f",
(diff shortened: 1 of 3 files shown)


sno/checkout.py
```diff
@@ -178,27 +208,65 @@ def switch(ctx, create, force_create, discard_changes, refish):
         # Switch to existing branch
         #
         # refish could be:
-        # - branch name
+        # - local branch name eg 'master'
+        # - remote branch name eg 'master' (if do_guess is True)
```
craigds, line 212:
```suggestion
        # - local branch name (eg 'master') that only exists on remote (if do_guess is True)
```

sno/checkout.py
```diff
@@ -178,27 +208,65 @@ def switch(ctx, create, force_create, discard_changes, refish):
...
+            branch.upstream = existing_branch
+
         head_ref = branch.name
 
     reset_wc_if_needed(repo, commit, discard_changes=discard_changes)
 
     repo.set_head(head_ref)
 
 
+def _find_remote_by_name(repo, name):
```
craigds, line 256: I'd suggest calling this `_find_remote_branch_by_name`, since it finds a branch rather than a remote

craigds (approved):

PR #294, opened 2020-10-31
apply: Add `--ref` to apply onto other branches

craigds:
## Description

Adds a `--ref` option which allows you to apply a patch onto a ref other than HEAD. This is useful for server apps which need to apply patches in bare repos to a variety of branches without moving HEAD.

## Related links:

none

## Checklist:

### DIFF
--- a/sno/apply.py
+++ b/sno/apply.py
@@ -91,13 +91,30 @@ def unjson_feature(geom_column_name, f):
 
 
 def apply_patch(
-    *, repo, commit, patch_file, allow_empty, allow_missing_old_values=False, **kwargs
+    *,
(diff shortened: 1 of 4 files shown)


sno/apply.py
```diff
@@ -237,6 +255,7 @@ def parse_delta(change):
         "certainty about the applicability of the patch. Use with caution."
     ),
 )
+@click.option("--ref", default="HEAD", help="What ref to apply the patch onto.")
```
olsen232, line 258: I would say, "which ref". Is it better?

olsen232 (approved):

PR #298, opened 2020-11-03
Always open pygit2.Repository as bare

olsen232:
## Description

From pygit / libgit's point of view, we don't have a working copy, and the index file is unreadable. (This is by design: it stops the `git` command line tool from messing up the repo). Previously, repositories were "bare" for the same reason, but now they aren't: it was a bit odd to have git-bare sno-bare repos, with no working copy at all, and git-bare sno-non-bare repos, with a sno working copy (eg a GPKG).
 So now the repo's config is appropriately bare or non-bare, but it means pygit / libgit sometimes has trouble reading the repository if it does some operation that might involve the index, even if only peripherally - like merge_trees, which doesn't really need the index, but opens it anyway to see if there are merge directives in there.

We stopped this previously by temporarily deleting the index, but this is a better fix: we always open a pygit2.Repository as bare, ignoring the config.

## Related links:

https://github.com/koordinates/sno/issues/297

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ jsonschema==3.2.0         # via -r requirements/requirements.in
 msgpack==0.6.2            # via -r requirements/requirements.in
 #psycopg2==2.8.5           # via -r requirements/requirements.in
 pycparser==2.20           # via cffi
-#pygit2==1.1.0             # via -r requirements/requirements.in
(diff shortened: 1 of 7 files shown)


rcoup:
`vendor/makefile.vc` needs updating too for Windows?

rcoup (approved):
Nice 👍

PR #302, opened 2020-11-09
Allow custom CRS to be edited in Postgis WC

olsen232:
There are two different cases where a CRS might differ in the dataset and the WC, and we want to treat them differently.

In one case, the CRS might be a standard one like EPSG:2193, and it might be in the WC already because it came installed
with Postgis. The dataset might have this same CRS, perhaps imported from some data source, and these two CRS might
differ slightly. In this case, we don't want to mess with the CRS that came with the Postgis working copy, but we also
don't want to commit the Postgis version to the dataset and add a spurious change to the history - what if another
working copy has another version? The CRS could end up switching backwards and forwards depending on which working copy is used. So in this case, we want to ignore the change.

In the other case, the CRS is custom and comes from the dataset. In this case, we are happy to let the user update the
CRS in the working copy and commit the change to the working copy, just as they can in GPKG.

Also improves CRS code generation logic -
generated CRS code shouldn't change if the authority / name is set,
and remains unchanged.

No change to CHANGELOG - still covered by "adding Postgis working copy".
Still TODO: Add documentation about different working copies.

### DIFF
--- a/sno/crs_util.py
+++ b/sno/crs_util.py
@@ -83,21 +83,53 @@ def parse_authority(crs):
     return spatial_ref.GetAuthorityName(None), spatial_ref.GetAuthorityCode(None)
 
 
-def get_identifier_str(crs, authority=None):
+def get_identifier_str(crs):
     """
-    Given a CRS, generate a stable, unique identifier for it of type 'str'. Eg: "EPSG:2193"
+    Given a CRS, find or generate a stable, unique identifier for it of type 'str'.
+    Eg: "EPSG:2193" or "CUSTOM:201234"
     """
-    if authority is not None:
-        auth_name, auth_code = authority
-    else:
-        auth_name, auth_code = parse_authority(crs)
+    result = _find_identifier_str(crs)
+    if not result:
+        result = f"CUSTOM:{_generate_identifier_int(crs)}"
+    return result.replace("/", "_")
 
+
+def _find_identifier_str(crs):
+    """Given a CRS, find a sensible identifier string for it from within the WKT."""
+    auth_name, auth_code = parse_authority(crs)
+    # Use AUTH_NAME:AUTH_CODE if both are set:
     if auth_name and auth_code:
         return f"{auth_name}:{auth_code}"
+    # Use AUTH_NAME or AUTH_CODE if one of them is set and probably a real identifier:
     code = auth_name or auth_code
-    if code and code.strip() not in ("0", "EPSG"):
+    if code and code.strip() not in ("", "0", "EPSG", "ESRI"):
         return code
-    return f"CUSTOM:{get_identifier_int(crs, (auth_name, auth_code))}"
+    # Otherwise, use the CRS name, if set.
+    name = parse_name(crs)
+    if name is not None:
+        name = name.strip()
+    return name if name else None
+
+
+def get_identifier_int(crs):
+    """
+    Given a CRS, find or generate a stable, unique identifer for it of type 'int'.
+    Eg: 2193 or 201234
+    """
+    # Find the auth code from the WKT if one is set and is an integer:
+    result = _find_identifier_int(crs)
+    # Otherwise, generate a stable ID based on the WKT authority / name / contents.
+    if not result:
+        result = _generate_identifier_int(crs)
+    return result
+
+
+def _find_identifier_int(crs):
+    """Given a CRS, find a sensible identifier int for it from its authority in the WKT."""
+    auth_name, auth_code = parse_authority(crs)
+    if auth_code and auth_code.isdigit() and int(auth_code) > 0:
+        return int(auth_code)
+    return None
 
 
 MIN_CUSTOM_ID = 200000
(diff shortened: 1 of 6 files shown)


sno/crs_util.py
```diff
@@ -157,3 +189,12 @@ def normalise_wkt(wkt):
...
     return "".join(token_value)
+
+
+def has_standard_authority(crs):
+    """
+    Returns True if this CRS appears to be defined by some standard authority - EPSG, ESRI, or Google.
+    Doesn't check if the CRS actually matches (or roughly matches) the definition specified by that authority.
+    """
+    auth_name, auth_code = parse_authority(crs)
+    return auth_name in ("EPSG", "ESRI") or auth_code == 600613  # GOOGLE
```
rcoup, line 200: I think you mean 900913?

```suggestion
    return auth_name in ("EPSG", "ESRI") or auth_code == 900913  # GOOGLE
```

(900913 was later defined as `EPSG:3857`, FWIW)

sno/working_copy/postgis.py
```diff
@@ -335,12 +335,19 @@ def write_meta_crs(self, dataset):
         with self.session() as db:
             dbcur = db.cursor()
             for row in spatial_refs:
-                # We only write a CRS if none is there for the given ID - the CRS of a given ID shouldn't change.
+                # We do not automatically overwrite a CRS if it seems likely to be one
+                # of the Postgis builtin definitions - Postgis has lots of EPSG and ESRI
+                # definitions built-in, plus the 600613 (GOOGLE) definition.
```
rcoup, line 340:
```suggestion
                # definitions built-in, plus the 900913 (GOOGLE) definition.
```

sno/working_copy/postgis.py
```diff
@@ -335,12 +335,19 @@ def write_meta_crs(self, dataset):
...
                 dbcur.execute(
                     SQL(
                         """
-                        INSERT INTO public.spatial_ref_sys ({}) VALUES ({})
-                        ON CONFLICT (srid) DO NOTHING;
+                        INSERT INTO public.spatial_ref_sys AS SRS ({}) VALUES ({})
+                        ON CONFLICT (srid) DO UPDATE
+                            SET (auth_name, auth_srid, srtext, proj4text)
+                            = (EXCLUDED.auth_name, EXCLUDED.auth_srid, EXCLUDED.srtext, EXCLUDED.proj4text)
+                            WHERE SRS.auth_name NOT IN ('EPSG', 'ESRI') AND SRS.srid <> 600613;
```
rcoup, line 350:
```suggestion
                            WHERE SRS.auth_name NOT IN ('EPSG', 'ESRI') AND SRS.srid <> 900913;
```

sno/crs_util.py
```diff
@@ -83,49 +83,81 @@ def parse_authority(crs):
...
+def _find_identifier_str(crs):
+    """Given a CRS, find a sensible identifier string for it from within the WKT."""
+    auth_name, auth_code = parse_authority(crs)
+    # Use AUTH_NAME:AUTH_CODE if both are set:
     if auth_name and auth_code:
         return f"{auth_name}:{auth_code}"
+    # Use AUTH_NAME or AUTH_CODE if one of them is set and probably a real identifier:
     code = auth_name or auth_code
-    if code and code.strip() not in ("0", "EPSG"):
+    if code and code.strip() not in ("", "0", "EPSG", "ESRI"):
```
rcoup, line 105: http://www.geopackage.org/spec/#r11 suggests -1 is a "normal" value for "i don't have a CRS" too, depending on the source. 

But that's unlikely to come out of WKT/GDAL, more likely to see in a GPKG or PostGIS  DB.

rcoup (approved):
One thing that might be relevant — it'd be fairly normal to define a srid number in `spatial_ref_sys` (GPKG or PostGIS) in the `srid` and/or `auth_srid` fields _without_ having it in the associated WKT. 

Does that get picked up?

ie:
```sql
INSERT INTO spatial_ref_sys (srid, auth_name, auth_srid, srtext)
VALUES (
  (61234, 'CUSTOM', 0, 'PROJCS[... no authority stanzas...]'),
  (62345, 'CUSTOM', 62345, 'PROJCS[... no authority stanzas...]')
);
```

sno/crs_util.py
olsen232: Done

sno/working_copy/postgis.py
olsen232: Done

sno/working_copy/postgis.py
olsen232: Done

olsen232:
We currently don't pick up anything that isn't in the WKT itself.

The basic structure right now is that we have a WKT CRS - that come from somewhere - and we know is associated with a particular geometry column.
We store that WKT CRS in the dataset by selecting a filename for it eg `crs/EPSG:2193.wkt` and linking to that filename to schema.json, and we store that CRS in the WC by picking a srid for it eg 2193 and linking to that srid in the geometry_columns table. The filename / srid we pick are functions of the CRS and are only used to keep the CRS properly "JOINED" to the data - we could pick anything, but obviously it makes most sense to pick 2193 for EPSG:2193.

The reason for having the above structure is probably because the place we generally get a CRS - the OGR import source - hides the means by which a CRS is attached: all we know is that this is the data, and this is the associated CRS. So from import onwards, we only know the CRS and it's up to us to keep it attached as we see fit.
And, mostly it doesn't matter very much: mostly we get official definitions with authorities included.

We could revise this - work harder at finding how the CRS is currently attached on import (ie, what srid it has even if it has no authority) and keep that srid everywhere. We would have to be careful to check and change the logic everywhere - if we just start looking in the working copy for srids right now and treating them as prescriptive, we will also find srids that we auto-generated a moment ago, and we may not want to treat those as prescriptive. See example below:

Example of thing to avoid if we start reading srids from working copy:
eg: User attached CRS with authority="PA27". We store this briefly in the dataset at `crs/PA27.wkt`. Then we store in the working copy. We don't have an int srid that fits in the working copy, so we generate one: 201234. Then we find this srid in the working copy and decide it is prescriptive, so we move the CRS in the dataset from `crs/PA27.wkt` to `crs/CUSTOM:201234.wkt`.

PR #304, opened 2020-11-11
Tighten up type roundtrips ready for postgis launch

olsen232:
## Description

Lots of little changes to code dealing with types, now that we have two different working copies that support different types.

Approximated types:
- Made sure that every type supported by sno can be written - somehow - to a working copy, even if just as a string.
- Implemented "approximated types":
eg "time" is approximated as "text" in GPKG, and "int8" is approximated as "int16" in Postgis.
If approximated types aren't changed by the user and remain the closest possible approximation, they don't count as a diff.
- Added tests to ensure that all types are either roundtrippable or known to be "approximated" in both working copies.

- Make sure the more exotic types returned by postgis are strings and not eg numeric.Decimal, which we wouldn't be able serialise to a dataset using MessagePack.

- Changed schema.json - drop attributes that are null eg {"size": null} or {"primaryKeyIndex": null}. Otherwise we have too many ways to represent the same data - we don't need to distinguish between null and missing, and there was already a
lack of consistency as to which was the right way. So, now the standard is that null attributes are dropped, and loaded schemas are normalised to this standard to prevent spurious diffs.

- Standardise how X_TYPE_TO_Y_TYPE mappings - which we have a few of by now - are encoded in python.

## Related links:

https://github.com/koordinates/sno/issues/293

## Checklist:


tests/test_working_copy_postgis.py
```diff
@@ -345,3 +341,24 @@ def test_edit_crs(data_archive, geopackage, cli_runner, new_postgis_db_schema):
                     )
 
                     raise SucceedAndRollback()
+
+
+def test_approximated_types():
+    postgis_adapter.APPROXIMATED_TYPES == compute_approximated_types(
```
craigds, line 347:
```suggestion
    assert postgis_adapter.APPROXIMATED_TYPES == compute_approximated_types(
```

craigds (approved):
🚀

sno/working_copy/postgis.py
```diff
@@ -46,17 +46,38 @@ def adapt_geometry_from_db(g, dbcur):
 
 def adapt_timestamp_from_db(t, dbcur):
     # TODO - revisit timezones.
-    if isinstance(t, str):
-        # This makes postgis timestamps behave more like GPKG ones.
-        return t.replace(" ", "T") + "Z"
-    return t
+    if t is None:
+        return t
+    # This makes postgis timestamps behave more like GPKG ones.
```
rcoup, line 51: Postgis timestamps-with-time-zones strings could be `2020-11-13T10:00:00+13:00` though? Seems like suffixing `Z` is too simplistic?

sno/working_copy/postgis.py
```diff
@@ -46,17 +46,38 @@ def adapt_geometry_from_db(g, dbcur):
...
 
 
 # See https://github.com/psycopg/psycopg2/blob/master/psycopg/typecast_builtins.c
 TIMESTAMP_OID = 1114
 TIMESTAMP = new_type((TIMESTAMP_OID,), "TIMESTAMP", adapt_timestamp_from_db)
 psycopg2.extensions.register_type(TIMESTAMP)
 
+# We mostly want data out of the DB as strings, just as happens in GPKG.
+# Then we can serialise it using MessagePack.
+# TODO - maybe move adapt-to-string logic to MessagePack?
```
rcoup, line 66: feels like it should be a wc-specific function, perhaps with utils/base-classes/mixins/helpers to DRY.

PR #306, opened 2020-11-13
Tighten up working copy states ready for postgis

olsen232:
## Description

Various tweaks to how working copy state is sanity-checked -

More nuance: working copies can be
- created (that is, existing) or not
- initialised (= existing plus having the sno-specific tracking + state tables) or not
- has_data or not - if they seem to contain user data already, ignoring the state tables above.

Most places we try to create a working copy, we expect them to be either non-existent, or both created+initialised. If they are created but not initialised, an error will be raised. If they are not created at all, that might still be okay depending on the command.

Some commands will create a working copy automatically if one doesn't exist - eg `checkout`, `init --import`, `clone`. These move the working copy from one valid state - uncreated - to the other - created + initialised - so they will still fail if the working copy is already created but uninitialised.

#### is_created is slightly different for postgres:
Postgres working copies are treated as "uncreated" if the schema exists but doesn't contain anything. This is so you can create an empty postgres schema ahead of time, and then have postgres work exactly the same as GPKG - it starts off "uncreated" and will get created at a sensible time, probably the first time you import data. An empty postgres schema shouldn't trigger an alert "WC already exists but is missing sno tables" like an empty GPKG should, or like a postgres schema that is non-empty but is missing sno tables.

#### create-workingcopy is special:
`sno create-workingcopy` is more lenient about these things so that you can use it to recreate the working copy if it gets broken somehow. It will delete any existing working copy you have at the old configured location, delete any existing working copy you may have at the newly provided location, and then write a new working copy from scratch at the new provided location. However, it takes care not to delete the old schema if it is the same as the new schema, since sno might lack permissions to recreate the schema.

## Related links:

https://github.com/koordinates/sno/issues/293

### DIFF
--- a/sno/working_copy/base.py
+++ b/sno/working_copy/base.py
@@ -8,14 +8,16 @@ import pygit2
 
 
 from sno.diff_structs import RepoDiff, DatasetDiff, DeltaDiff, Delta
-from sno.exceptions import InvalidOperation, NotYetImplemented
+from sno.exceptions import (
+    InvalidOperation,
+    NotYetImplemented,
+    NotFound,
+    NO_WORKING_COPY,
+)
 from sno.filter_util import UNFILTERED
 from sno.repository_version import get_repo_version
 from sno.schema import Schema
 from sno.structure import RepositoryStructure
-
-
-SNO_WORKINGCOPY_PATH = "sno.workingcopy.path"
 
 
 L = logging.getLogger("sno.working_copy.base")
@@ -39,6 +41,8 @@ class Mismatch(ValueError):
 
 
 class WorkingCopy:
+    SNO_WORKINGCOPY_PATH = "sno.workingcopy.path"
+
     VALID_VERSIONS = (1, 2)
 
     TRACKING_NAME = "track"
@@ -53,25 +57,34 @@ class WorkingCopy:
         return self._sno_table(self.STATE_NAME)
 
     @classmethod
-    def get(cls, repo, init_if_missing=False):
+    def get(cls, repo, *, allow_uncreated=False, allow_invalid_state=False):
         """
-        Get the working copy associaated with this sno repo.
-        If no working copy is present, returns None -
-        unless init_if_missing is returned, in which case it returns a WorkingCopy object that is not "created".
-        The type of this WorkingCopy object depends on the repo config - it represents a particular type of WorkingCopy
-        at a particular location, which nevertheless doesn't exist yet. It is up to the caller to create it by calling
-        `create()` and to populate it by calling eg `write_full()`.
-
+        Get the working copy associated with this sno repo, as specified in the repo config.
+        Note that the working copy specified in the repo config may or may not exist or be in a valid state.
+        An instance of this class can represent a working copy that doesn't exist or is in an invalid state,
+        (similar to how pathlib.Path can point to files that may or may not exist).
+        If allow_uncreated is True, a non-existant working copy may be returned - otherwise, only an existing
+        working copy will be returned, and None will be returned if no working copy is found.
+        If allow_invalid_state is True, an invalid-state working copy may be returned - otherwise, only a valid
+        working copy will be returned, and a NotFound(NO_WORKING_COPY) will be raised if it is in an invalid state.
         """
-        if init_if_missing:
-            cls.ensure_config_exists(repo)
-
         repo_cfg = repo.config
-        path_key = SNO_WORKINGCOPY_PATH
+        path_key = cls.SNO_WORKINGCOPY_PATH
         if path_key not in repo_cfg:
             return None
 
         path = repo_cfg[path_key]
+        return cls.get_at_path(
+            repo,
+            path,
+            allow_uncreated=allow_uncreated,
+            allow_invalid_state=allow_invalid_state,
+        )
+
+    @classmethod
+    def get_at_path(
+        cls, repo, path, *, allow_uncreated=False, allow_invalid_state=False
+    ):
         if cls.is_postgres_uri(path):
             from .postgis import WorkingCopy_Postgis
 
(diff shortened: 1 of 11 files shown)


sno/working_copy/base.py
```diff
@@ -53,25 +57,26 @@ def STATE_TABLE(self):
...
 
         """
-        if init_if_missing:
-            cls.ensure_config_exists(repo)
-
         repo_cfg = repo.config
-        path_key = SNO_WORKINGCOPY_PATH
+        path_key = cls.SNO_WORKINGCOPY_PATH
         if path_key not in repo_cfg:
             return None
```
craigds, line 73: >          If no working copy is present, returns None -
>          unless allow_uncreated is returned, in which case it returns a WorkingCopy object that is not "created".

seems to conflict with this. docstring not quite accurate?

sno/working_copy/base.py
```diff
@@ -53,25 +57,26 @@ def STATE_TABLE(self):
...
 
         path = repo_cfg[path_key]
+        return cls.get_at_path(repo, path, allow_uncreated, allow_invalid_state)
+
+    @classmethod
+    def get_at_path(cls, repo, path, allow_uncreated=False, allow_invalid_state=False):
```
craigds, line 79:
```suggestion
    def get_at_path(cls, repo, path, *, allow_uncreated=False, allow_invalid_state=False):
```
(and change caller(s) to suit)

multiple boolean flags are a good fit for keyword-only arguments since it's too easy to pass them in the wrong order otherwise

sno/working_copy/gpkg.py
```diff
@@ -178,9 +168,48 @@ def delete(self):
...
+        the GPKG file exists and has the necessary gpkg_sno tables - state and tracking.
+        """
+        if not self.is_created():
+            return False
+        with self.session() as db:
+            dbcur = db.cursor()
+            dbcur.execute(
+                f"""
+                SELECT count(*) FROM sqlite_master
+                WHERE type='table' AND name IN ({self.STATE_TABLE}, {self.TRACKING_TABLE});
```
craigds, line 189: `self.STATE_TABLE` and `self.TRACKING_TABLE` are identifier-quoted strings, does this work? It seems like it probably shouldn't

sno/working_copy/gpkg.py
```diff
@@ -178,9 +168,48 @@ def delete(self):
...
+        """
+        if not self.is_created():
+            return False
+        with self.session() as db:
+            dbcur = db.cursor()
+            dbcur.execute(
+                f"""
+                SELECT count(*) FROM sqlite_master
+                WHERE type='table'
+                    AND name NOT IN ({self.STATE_TABLE}, {self.TRACKING_TABLE})
```
craigds, line 206: same question here

sno/working_copy/postgis.py
```diff
@@ -299,7 +331,7 @@ def create(self):
                 )
             )
 
-    def delete(self):
+    def delete(self, keep_container_if_possible=False):
         """ Delete all tables in the schema"""
         # We don't use drop ... cascade since that could also delete things outside the schema.
```
craigds, line 336: Because tables can depend on each other with foreign keys, and we don't drop them in any particular order, it's quite likely this function can fail because of FK constraints, even though all the tables will eventually be dropped.

Probably this could be avoided with `SET CONSTRAINTS ALL DEFERRED`

sno/working_copy/base.py
olsen232: Fixed up the doc.

sno/working_copy/base.py
olsen232: Done

sno/working_copy/postgis.py
olsen232: Done

sno/working_copy/gpkg.py
olsen232: It works because the engine we're using doesn't have identifier specific quoting anyway - gpkg.ident() just adds double quotes and escapes any existing double quotes, which is fine for the name of a column, or for a string. Tests would fail if it didn't work.

Will leave this as-is for now - can revisit when we get to sqlalchemy.

sno/working_copy/gpkg.py
olsen232: As above.

sno/working_copy/gpkg.py
craigds: double-quotes _is_ identifier-specific quoting, at least in postgres. I'm surprised it works in sqlite

rcoup: https://sqlite.org/lang_keywords.html

> If you want to use a keyword as a name, you need to quote it. There are four ways of quoting keywords in SQLite:
>
> * `'keyword'` A keyword in single quotes is a string literal.
> * `"keyword"` A keyword in double-quotes is an identifier.
> * `[keyword]` A keyword enclosed in square brackets is an identifier. This is not standard SQL. This quoting mechanism is used by MS Access and SQL Server and is included in SQLite for compatibility.
> * `` `keyword` `` A keyword enclosed in grave accents (ASCII code 96) is an identifier. This is not standard SQL. This quoting mechanism is used by MySQL and is included in SQLite for compatibility.
> 
> For resilience when confronted with historical SQL statements, SQLite will sometimes bend the quoting rules above:
> * If a keyword in single quotes (ex: `'key'` or `'glob'`) is used in a context where an identifier is allowed but where a string literal is not allowed, then the token is understood to be an identifier instead of a string literal.
> * If a keyword in double quotes (ex: `"key"` or `"glob"`) is used in a context where it cannot be resolved to an identifier but where a string literal is allowed, then the token is understood to be a string literal instead of an identifier.

🤷

PR #312, opened 2020-11-22
Always return features as dicts

olsen232:
## Description

Drop support for returning features from datasets as tuples - always return them as dicts.
There were three possible reasons for getting features as tuples as opposed to dicts -
a) as an optimisation, tuples are slightly quicker if you don't need the keys
b) the get_feature_tuple code let you specify which columns you wanted in which order, which might be useful...
c) a + b together - generating a tuple could be even quicker if you don't even need all the values

However, they were relatively few calls to get_feature_tuple, relatively few places where we only need some subset of a row, and datasets V2 had limited support for these optimisations anyway - if you specified that you only wanted certain columns in a certain order, it would just generate the dict and then convert that to a tuple for you - nothing the client couldn't do themselves, and not something they do very often. In any case unpacking a feature from the git object is what takes the most time currently, and the entire feature is unpacked regardless - whether it is put into a dict or a tuple afterwards doesn't really change the overall speed.

As a result of this change, dataset1 and 2 have more in common than before, so some of the code is able to be moved into base_dataset.py

I did some profiling and it doesn't appear to have changed anything - some tests were up 5% quicker, the some were up to 5% slower, so it looks like mostly noise:
https://docs.google.com/spreadsheets/d/174DhZ4N3cS3pfjleHlPjmtxmk5HZDrmVIflAIcIKSvk/edit#gid=922066028

This is done to make the code simpler ready for alphanumeric PKs - I began work on alphanumeric PKs but ran into a problem where get_feature_tuple not doing what I wanted, so decided to make this change I've had planned for a while

### DIFF
--- a/sno/working_copy/gpkg.py
+++ b/sno/working_copy/gpkg.py
@@ -639,9 +639,10 @@ class WorkingCopy_GPKG(WorkingCopy):
 
                 CHUNK_SIZE = 10000
                 total_features = dataset.feature_count
-                for rows in self._chunk(dataset.feature_tuples(col_names), CHUNK_SIZE):
-                    dbcur.executemany(sql_insert_features, rows)
-                    feat_progress += len(rows)
+                for row_dicts in self._chunk(dataset.features(), CHUNK_SIZE):
+                    row_tuples = (row_dict.values() for row_dict in row_dicts)
+                    dbcur.executemany(sql_insert_features, row_tuples)
+                    feat_progress += len(row_dicts)
 
                     t0a = time.monotonic()
                     L.info(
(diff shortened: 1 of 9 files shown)


sno/working_copy/gpkg.py
```diff
@@ -639,9 +639,10 @@ def write_full(self, target_tree_or_commit, *datasets, safe=True):
 
                 CHUNK_SIZE = 10000
                 total_features = dataset.feature_count
-                for rows in self._chunk(dataset.feature_tuples(col_names), CHUNK_SIZE):
-                    dbcur.executemany(sql_insert_features, rows)
-                    feat_progress += len(rows)
+                for row_dicts in self._chunk(dataset.features(), CHUNK_SIZE):
+                    row_tuples = (row_dict.values() for row_dict in row_dicts)
+                    dbcur.executemany(sql_insert_features, row_tuples)
+                    feat_progress += db.changes()  # DONOTSUBMIT
```
craigds, line 645: ☝️ 

I can't think of a reason why this would be different than `len(row_dicts)`, since `INSERT INTO` always inserts a row

craigds (approved):

sno/working_copy/gpkg.py
olsen232: Done

PR #313, opened 2020-11-24
Add support for string PK datasets in GPKG WC

olsen232:
## Description

GPKG feature tables (=spatial tables, with geometry) only support integer primary keys. Data from non-GPKG sources might have other types of primary key, most commonly strings. This workaround creates GPKG working copies with dummy integer primary keys, and demotes the actual primary key column to just a `UNIQUE NOT NULL` column (but only in the GPKG working copy - in the dataset it keeps its status as a primary key).

## Related links:

https://github.com/koordinates/sno/issues/307

## Checklist:


sno/working_copy/gpkg.py
```diff
@@ -616,20 +649,21 @@ def write_full(self, target_tree_or_commit, *datasets, safe=True):
...
-                    CREATE TABLE {gpkg.ident(table)}
-                    ({', '.join(col_specs)});
-                """
-                )
+                table_spec = gpkg_adapter.v2_schema_to_sqlite_spec(dataset)
+
+                # GPKG requires an integer primary key for spatial tables, so we add it in if needed:
+                if dataset.has_geometry and "PRIMARY KEY" not in table_spec:
+                    table_spec = (
+                        '"integer-pk" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,'
```
craigds, line 657: maybe `".sno-generated-pk"` would be better so it's clearer where it came from

craigds (approved):
LGTM

sno/working_copy/gpkg.py
olsen232: Sort-of-done - changed to `sno-auto-pk`. What do you think?

Nuance: an "auto-pk" sounds to me more like something temporary and disposable. A "generated-pk" sounds to me more like something we expended some effort creating for the user, and maybe like something we could regenerate consistently (but we can't, it's just some random number).

docs/GPKG_WC.md
```diff
@@ -11,11 +11,22 @@ Most geospatial data can be converted to GPKG format without losing any fidelity
 
 #### Primary keys
 
-GPKG requires integer primary keys. If your data has alphanumeric primary keys, no primary keys at all, a composite key made up of more than one column, or any other type of primary key than an integer, it will not be possible to check out your data into a GPKG.
+GPKG requires integer primary keys, for tables that have a geometry column - other tables do not have this constraint [[GPKG spec]](http://www.geopackage.org/spec120/#feature_user_tables). Depending on what kind of primary keys your dataset has, you may or may not be able to check it out using a GPKG working copy.
```
rcoup, line 14: We're saying you can 
```suggestion
GPKG requires integer primary keys for tables that have a geometry column. Other tables do not have this constraint [[GPKG spec]](http://www.geopackage.org/spec120/#feature_user_tables). Sno follows these requirements as described below:
```

docs/GPKG_WC.md
```diff
@@ -11,11 +11,22 @@ Most geospatial data can be converted to GPKG format without losing any fidelity
...
 
-Work in this area is ongoing, expect to see a workaround for alphanumeric primary keys and for data without primary keys as Sno development continues.
+ * **Integer primary keys** \
+     There are no issues checking out datasets with integer primary keys in a GPKG working copy.
```
rcoup, line 17:
```suggestion
     Integer primary keys are preserved fully and function normally in a GPKG working copy.
```

docs/GPKG_WC.md
```diff
@@ -11,11 +11,22 @@ Most geospatial data can be converted to GPKG format without losing any fidelity
...
 
-Tracking bug for data with alphanumeric primary keys: [#307](https://github.com/koordinates/sno/issues/307)
+ * **String primary keys** (or any other non-integer type) \
+     If the dataset **does not** have a geometry column, then these primary keys are checked out without modification. \
+     If the dataset **does** have a geometry column, then the non-integer primary key column is demoted to a regular column (but constrained to be unique and non-null), and another column is added to the table that is an integer primary key column, for the sole purpose of making the table conform to the GPKG specification. The additional integer column is called `.sno-auto-pk`. It is not part of the dataset, its contents are arbitary (except that they conform to GPKG specification), and it is not tracked by sno - any edits specifically to this column will not be committed.
```
rcoup, line 21:
```suggestion
     If the dataset **does** have a geometry column, then the primary key column is demoted to a regular column (but constrained to be unique and not-null), and another column is added to the table that is an integer primary key column, for the sole purpose of making the table conform to the GPKG specification. The additional integer column is called `.sno-auto-pk`. It is not part of the dataset, its contents are arbitrary (except that they conform to the GPKG requirements), and it is not tracked by Sno - any edits specifically to this column will not be committed.
```

docs/GPKG_WC.md
```diff
@@ -11,11 +11,22 @@ Most geospatial data can be converted to GPKG format without losing any fidelity
...
+
+ * **No primary key column** \
+   This is not yet supported.
+
+ * **Composite primary keys** (multiple primary key columns) \
+   This is not yet supported.
+
+Work in this area is ongoing, expect to see a workaround for data without primary keys as Sno development continues.
```
rcoup, line 29:
```suggestion
Work in this area is ongoing, expect to see support for data without primary keys as Sno development continues.
```

sno/gpkg_adapter.py
```diff
@@ -229,6 +228,33 @@ def generate_sqlite_table_info(v2_obj):
...
+    col_name = gpkg.ident(column_schema.name)
+    result = f"{col_name} {gpkg_type}"
+
+    is_pk = column_schema.pk_index is not None
+    if is_pk:
+        if gpkg_type == "INTEGER":
+            result += " PRIMARY KEY AUTOINCREMENT NOT NULL"
+        elif has_geometry:
+            # GPKG feature-tables only allow integer PKs, so we demote this PK to a regular UNIQUE field.
+            result += f" UNIQUE NOT NULL CHECK({col_name}<> '')"
```
rcoup, line 250:
```suggestion
            result += f" UNIQUE NOT NULL CHECK({col_name}<>'')"
```

docs/GPKG_WC.md
```diff
@@ -11,11 +11,22 @@ Most geospatial data can be converted to GPKG format without losing any fidelity
...
-Tracking bug for data with alphanumeric primary keys: [#307](https://github.com/koordinates/sno/issues/307)
+ * **String primary keys** (or any other non-integer type) \
+     If the dataset **does not** have a geometry column, then these primary keys are checked out without modification. \
+     If the dataset **does** have a geometry column, then the non-integer primary key column is demoted to a regular column (but constrained to be unique and non-null), and another column is added to the table that is an integer primary key column, for the sole purpose of making the table conform to the GPKG specification. The additional integer column is called `.sno-auto-pk`. It is not part of the dataset, its contents are arbitary (except that they conform to GPKG specification), and it is not tracked by sno - any edits specifically to this column will not be committed.
+
+ * **No primary key column** \
+   This is not yet supported.
+
+ * **Composite primary keys** (multiple primary key columns) \
+   This is not yet supported.
```
rcoup, line 27: For non-spatial tables too?

rcoup (approved):
Nice work 🌟 

I presume using [ROWID](https://www.sqlite.org/rowidtable.html) didn't work out?

Are there any edge cases wrt schema changes we need to be careful with? I'm wondering what happens when you drop/add a geometry column to a table with a non-integer PK.

craigds:
anyone have any idea _why_ geopackage spec doesn't support non-integer PKs for spatial tables? I googled a fair bit but couldn't find any reason anywhere.

rcoup:
You can probably find the working group minutes if you search hard enough, but my understanding is that they wanted to limit/simplify the scope of the spec as far as they could. To make it as simple as possible to quickly and correctly implement (into GIS packages, libraries, etc).

olsen232 (commented):
So one possible reason - a not very convincing reason in my opinion - you can link XML metadata to individual features by their integer primary key, using the gpkg_metadata and gpkg_metadata_reference. But not sure if they added the requirement to have integer primary keys just so that certain users could easily do this if they wanted - doesn't make sense to me.

@rcoup I didn't understand the ROWID thing last time - I thought it was just for my own education, and I thought "if QGIS supports INTEGER PRIMARY KEY, but now TEXT PRIMARY KEY, how on earth will it support TEXT PRIMARY KEY WITHOUT ROWID which is even more dissimilar?"
But just now I realised you probably expected me to try INTEGER PRIMARY KEY WITHOUT ROWID and then try inserting text into it. That is possible: I've just inserted some text into INTEGER PRIMARY KEY WITHOUT ROWID - but QGIS still doesn't like it. Which I'm fine with since otherwise we'd have to decide which design to go with. And it would be a bit ugly anyway - even if QGIS did like, who can say if ArcGIS would too, or if they would continue to support it in the future.

rcoup:
> @rcoup I didn't understand the ROWID thing last time - I thought it was just for my own education, and I thought "if QGIS supports INTEGER PRIMARY KEY, but now TEXT PRIMARY KEY, how on earth will it support TEXT PRIMARY KEY WITHOUT ROWID which is even more dissimilar?"

I more meant whether, since ROWID is present & hidden in any SQLite non-integer-pk table, whether we could just tell GPKG that `ROWID` is the PK for the table (in `gpkg_contents`) and not have to deal with another column.

olsen232:
The `gpkg_contents` table doesn't have any metadata about what is or isn't the primary key. Right?

I think the solution you describe is very similar in structure to what we is now submitted - SQLite creating a ROWID automatically, and then us somehow getting it to be treated as the primary key column by QGIS, is very similar to SQLite creating an INTEGER PRIMARY KEY AUTOINCREMENT column that maps directly onto the ROWID anyway. The difference - and the selling point for your suggested solution - would be that the ROWID would appear to be the primary key to QGIS, but a different column would appear to the primary key to Sno, so we wouldn't have to manually adapt between the two different schemas on roundtrip. But, the problem with that solution is that I don't know the mechanism by which we would make QGIS and Sno automatically have different opinions about what the primary key column is. Right now that mechanism is manual in that Sno manually rewrites the schema if it detects that this is what has happened. But, it might be the simplest we can get.

PR #326, opened 2020-12-08
Generate PKs while importing, for sources that lack them.

olsen232:
## Description

Assigns PKs during import, for data that has no primary key, simply from the sequence 1, 2, 3...
If the data is reimported with `--replace-existing`, attempts to give each feature the same primary key it had last time. This is achieved by storing metadata in `$DATASET_PATH/meta/generated-pks.json`

Includes one end-to-end test.

Still TODO:
Add more unit tests
Add more configuration options, eg, let the user choose the name or size of the generated primary key.

## Related links:

https://github.com/koordinates/sno/issues/212

## Checklist:

### DIFF
--- a/sno/pk_generation.py
+++ b/sno/pk_generation.py
@@ -1,0 +1,212 @@
+import functools
+from .serialise_util import (
+    json_pack,
+    json_unpack,
+)
+from collections.abc import Iterable
+from .schema import ColumnSchema, Schema
+
+
+class PkGeneratingImportSource:
+    """
+    Wrapper of ImportSource that makes it appear to have a primary key, even though the delegate ImportSource does not.
+    In the simplest case, every feature encountered is just assigned a primary key from the sequence 1, 2, 3...
+    However, this mapping from each feature to its primary key is stored as metadata the imported dataset, so that if
+    the same (or similar) data is reimported, then the same primary keys can be assigned to each feature. The
+    reimport will reuse primary keys for any features that are unchanged, and any new primary keys that must be
+    assigned are appended in the same manner to this mapping.
+
+    Note that reimporting depends only on the data to be imported and the stored primary key metadata - local edits to
+    imported features have no effect on how the data is reimported.
+
+    For the sake of efficiency, the entire feature is not stored in the mapping, but only a hash of its contents.
+    Since multiple features with the same contents may be imported, the mapping to be stored has the structure:
+
+    >>> {feature hash -> [list of primary keys]}
+
+    In fact the inverse mapping is stored, since it has a simpler structure, primary keys are unique:
+
+    >>> {primary key -> feature hash}
+
+    This is stored in $DATASET_PATH/meta/generated-pks.json, along with the column-schema of the new primary key::
+
+      {
+        "primaryKeySchema": {
+          "id": "ad068414-3a04-45ab-851d-bfa5104c60d6",
+          "name": "generated-pk",
+          "dataType": "integer",
+          "primaryKeyIndex": 0,
+          "size": 64
+        },
+        "generatedPrimaryKeys": {
+          "1": "181e23cf3a3c5e74254707687c4be2b5b02dbf63",
+          "2": "021ac25fcf4dafc72053f84d2b87ec5662adcb83",
+          "3": "8e775122edbdd367c8d383fffeabf6580de485fd",
+          ...
+        }
+      }
+
+    """
+
+    GENERATED_PKS_PATH = ".sno-dataset/meta/generated-pks.json"
+
+    DEFAULT_PK_COL = {
+        "id": ColumnSchema.new_id(),
+        "name": "generated-pk",
+        "dataType": "integer",
+        "primaryKeyIndex": 0,
+        "size": 64,
+    }
+
+    @classmethod
+    def wrap_if_needed(cls, source_or_sources, repo):
+        """
+        Wraps an ImportSource in a PkGeneratingImportSource if the original data lacks a primary key.
+        If multiple ImportSources are supplied, wraps those that lack a primary key and returns a new list.
+        """
+        if isinstance(source_or_sources, Iterable):
+            sources = source_or_sources
+            return [cls.wrap_if_needed(s, repo) for s in sources]
+
+        source = source_or_sources
+        if not source.schema.pk_columns:
+            return PkGeneratingImportSource(source, repo)
+        else:
+            return source
+
+    def __init__(self, delegate, repo):
+        self.delegate = delegate
+        self.load_data_from_repo(repo)
+
+    def load_data_from_repo(self, repo):
+        tree = repo.head_tree
+        generated_pks_blob = None
+
+        if tree is not None:
+            try:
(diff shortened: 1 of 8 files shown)


sno/schema.py
```diff
@@ -303,6 +305,13 @@ def encode_feature_blob(self, feature):
         data = msg_pack([self.legend.hexhash(), non_pk_values])
         return data
 
+    def hash_feature(self, feature):
+        """
+        Given a feature, git-hash it using this schema.
+        Doesn't include its path, so primary key values are not hashed.
```
rcoup, line 311: > primary key values are not hashed.

Feels like in general this behaviour is a footgun, right? In the pk-generating scenario it's obviously useful, but it'd be easy to miss it when someone sees there's a `hash_feature()` method. Maybe make it an opt-in behaviour, or rename the method?

sno/pk_generation.py
```diff
@@ -0,0 +1,176 @@
...
+)
+from collections.abc import Iterable
+from .schema import ColumnSchema, Schema
+
+
+class PkGeneratingImportSource:
+    """
+    Wrapper of ImportSource that makes it appear to have a primary key,
+    even though the delegate ImportSource does not.
+    """
```
rcoup, line 14: can we add a more detailed explanation of how the hashes/keys are stored and updated over time?

rcoup: Seems to be memory-limited, right — O(N) with number of features?

And based on all the features ever loaded in the repo, not just the current/previous revision?

sno/pk_generation.py
```diff
@@ -0,0 +1,176 @@
...
+    def features(self):
+        schema = self.schema
+        for feature in self.delegate.features():
+            feature[self.primary_key] = None
+            feature_hash = schema.hash_feature(feature)
+            feature[self.primary_key] = self.generate_pk(feature_hash)
+            yield feature
+
+    def generate_pk(self, feature_hash):
+        unused_pks = self.hash_to_unused_pks.get(feature_hash)
```
rcoup, line 117: Can you explain what/how `unused_pks` work? IMO we shouldn't be re-using numbers at all.

* Feature A gets imported
* and later deleted.
* then if an identical feature later gets imported again, it should get a new PK from the sequence.

sno/pk_generation.py
rcoup: these only apply to imports, right? not later edits, assuming they're not a reimport?

And what's the path to dealing with (or blocking) the merging multiple imports/edits?

sno/schema.py
olsen232: We use this in two places already - assign PKs, and detect renames - and we don't even need the hash-full-feature version. I think this one is more useful since it does exactly what git does and hashes a file but not the file's name.

However, I see your point that it's not clear from the name or the signature - I'll just add a parameter to hash or not-hash the pk values.

sno/pk_generation.py
olsen232: > can we add a more detailed explanation of how the hashes/keys are stored and updated over time?

Done

olsen232: > Seems to be memory-limited, right — O(N) with number of features?

Yup that's right - 40 bytes plus change for every feature ever imported / reimported into this repo. So, should be on the order of 40MB for every million features. Features edited not as part of an import don't count towards this total.

olsen232: > these only apply to imports, right? not later edits, assuming they're not a reimport?
Correct

> And what's the path to dealing with (or blocking) the merging multiple imports/edits?

This isn't handled specifically by this PR - and its not the intended flow - but it actually works as well as any other metadata conflict: let's say you import from the same source twice, but in two different branches, independently - each import is made without reference to the other (but perhaps both reference an earlier import). Then you can merge the two imports together, but you can get (up to) 3 types of conflicts

- the schema file can conflict if either import added new columns, since they will have new UUID.
- the generated-pks metadata file can conflict, if the generated mapping from feature to primary-key wasn't exactly the same each time
- every time two different features were imported with the same primary key, this will conflict

The good news is: you can resolve any of these by existing mechanisms. `sno resolve dataset:meta:generated-pks.json: --with=theirs` will accept their version of this metadata file.
The bad news is: resolving metadata conflicts is generally incomplete, and not just for generated primary keys. Although you can pick a winner, you can't do anything more complicated (ie merge two files into one), and picking a certain schema as the winner has implications for every feature, but there is no code for showing those implications yet. This is because conflict resolution hasn't been revisited since schema changes were added.

sno/pk_generation.py
olsen232: > Can you explain what/how unused_pks work?

Unused PKs are PKs that haven't yet been assigned during this import. I'll rename them to `unassigned_pks`.

Say you import 6 features, but some are identical: [A, A, A, B, B, C]
Say these are assigned primary keys: {A: [1,2,3], B: [4, 5], C: [6]}
This dict is called `hash_to_pks` and it exists at the end of the import.
Next time you import, you have a similar but different list of features: [A, A, B, B, D]
The same `hash_to_pks` dict is loaded again, but also duplicated as `hash_to_unused_pks`.

As you import the first A, you check the hash_to_unused_pks to see if you can find a primary key for it. `1` is the correct answer. But now, the 1 is removed from the list of unused_pks so that the next A encountered will be assigned `2`.

> IMO we shouldn't be re-using numbers at all.
The purpose of this dict - and of pk-less imports in general - is so that we DO want to reuse numbers. PK-less imports would be simpler if we could assign random PKs to every feature we encounter, but we want to assign a certain feature the same PK every time we reimport it. But then assign different PK if we see an identical feature a moment later. That's what this dict is for - keeping track of which PKs are the correct choice for feature A, but not using any of them more than once.

But you have another concern also:

> Feature A gets imported
> and later deleted.
> then if an identical feature later gets imported again, it should get a new PK from the sequence.

This is a non-goal of PK-less imports. Any local edits made should have no effect on how we re-import from the source next time we do an import (excepting that the user may find a way to edit the stored metadata). If local edits to features did have an effect on how we reimported from the same source, this would be inconsistent with current imports. Currently imports are stateless and depend only on the source, not on the local repo.

This PR makes imports stateful if they are PK-less imports, but only in-so-far as it tries to give the same features the same primary keys as they were given last time they were imported. It still doesn't let local feature edits affect the import.

Note that for regular imports using existing code - if you import the feature with PK=1, then delete it, then reimport from the same source, it reappears with the same PK as last time. So, PK-less imports is consistent with this.

The main use-case for this is mirroring, where there should generally be no local edits anyway.

sno/schema.py
```diff
@@ -293,15 +295,28 @@ def feature_to_raw_dict(self, feature):
                 raw_dict[column.id] = value
         return raw_dict
 
-    def encode_feature_blob(self, feature):
+    def encode_feature(self, feature, without_pk=False):
```
rcoup, line 298: nit, but can we make this `with_pk` so we don't end up with double-negatives?

sno/pk_generation.py
rcoup: Cool, thanks for clearing that up :)

rcoup (approved):
One minor naming tweak, but looks good.

PR #336, opened 2020-12-16
Optionally detect similar features during PK-less reimport.

olsen232:
## Description

Detects features which have changed slightly during a re-import from a data source without a primary key, and reimporting them with the same primary key as last time so they show as edits as opposed to inserts.
Features are currently considered similar if they differ by only one field. This could be changed / configurable in the future if better strategies are discovered.

Usage:
`sno import SOURCE --replace-existing --similarity-detection-limit=X`

where SOURCE is a datasource without primary keys,
and X is a number large enough that it should generally be larger than the number of edits (or inserts + deletes), but not too large so as to take forever. (Setting it to a thousand in the case that there are a thousand edits will result in a million feature-to-feature comparisons).

In the case where there are too many inserts + deletes to check, none of them will be checked, and they will all remain inserts and deletes - the new features will be assigned new primary keys and treated as separate from the old features.

`--similarity-detection-limit` is currently set to zero  by default - zero means "don't do similarity detection". Our mirroring pipeline may want to set it to some higher number. A non-zero but low default might also benefit end-users, once we have end-users doing reimports of PK-less datasets (we don't yet).

## Related links:

https://github.com/koordinates/sno/issues/212

## Checklist:

### DIFF
--- a/sno/init.py
+++ b/sno/init.py
@@ -133,6 +133,17 @@ def _add_datasets_to_working_copy(repo, *datasets, replace_existing=False):
     help="Replace existing dataset(s) of the same name.",
 )
 @click.option(
+    "--similarity-detection-limit",
+    hidden=True,
+    type=click.INT,
+    default=50,
+    help=(
+        "When replacing an existing dataset where primary keys are auto-generated: the maximum number of unmatched "
+        "features to search through for similar features, so that primary keys can be reassigned for features that "
+        "are similar but have had minor edits. Zero means that no similarity detection is performed. (Advanced users only)"
+    ),
+)
+@click.option(
     "--allow-empty",
     is_flag=True,
     default=False,
@@ -167,6 +178,7 @@ def import_table(
     tables,
     table_info,
     replace_existing,
+    similarity_detection_limit,
     allow_empty,
     max_delta_depth,
     do_checkout,
@@ -247,8 +259,11 @@ def import_table(
                 # will result in a new schema object, and thus a new blob for every feature.
                 # Note that alignment works better if we add the generated-pk-column first (when needed),
                 # if one schema has this and the other lacks it they will be harder to align.
-                import_source = PkGeneratingImportSource.wrap_if_needed(
-                    import_source, repo
+                import_source = PkGeneratingImportSource.wrap_source_if_needed(
+                    import_source,
+                    repo,
+                    dest_path=dest_path,
+                    similarity_detection_limit=similarity_detection_limit,
                 )
                 import_source.schema = existing_ds.schema.align_to_self(
                     import_source.schema
--- a/sno/pk_generation.py
+++ b/sno/pk_generation.py
@@ -1,9 +1,12 @@
-from collections.abc import Iterable
+from itertools import zip_longest
+import pygit2
+import subprocess
 
+from .dataset2 import Dataset2
 from .import_source import ImportSource
+from .exceptions import NotYetImplemented, SubprocessError
 from .serialise_util import (
     json_pack,
-    json_unpack,
 )
 from .schema import ColumnSchema, Schema
 
@@ -47,9 +50,22 @@ class PkGeneratingImportSource(ImportSource):
         }
       }
 
+    During a reimport, if similarity_detection_limit is set to some X > 0, and the import results in a number
+    of inserts + deletes[1] that is less than X, then these inserts and deletes will be searched to see if we can
+    find some inserts that are similar (not identical) to the deletes. These matching new features will be
+    reassigned a primary key from the delete, connecting the two features together so that that insert + delete
+    will instead show up in the log as an edit.
+    Currently, two features are considered "similar" if they differ by only one field, and they have only one such
+    counterpart.
+    Searching these inserts and deletes is computationally costly - every insert must be checked against every delete
+    so each doubling of the limit can result in four times the processing time.
+
+    [1] An insert is a feature which is not present in the previous import and so could not be reassigned an existing
+    primary key, and a delete is a feature that was present in the previous import but not in the current one.
     """
 
-    GENERATED_PKS_PATH = ".sno-dataset/meta/generated-pks.json"
+    GENERATED_PKS_ITEM = "generated-pks.json"
+    GENERATED_PKS_PATH = ".sno-dataset/meta/" + GENERATED_PKS_ITEM
 
     DEFAULT_PK_COL = {
         "id": ColumnSchema.new_id(),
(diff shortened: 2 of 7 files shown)


sno/pk_generation.py
```diff
@@ -109,16 +128,37 @@ def load_data_from_repo(self, repo):
...
+
+        # The number of inserts, deletes, previous- and current-feature-count, are related by the given formula:
+        # prev-FC + inserts - deletes = curr-FC
+        # Since we know prev-FC and curr-FC already, we can already calculate the number of inserts we can encounter
+        # before we know that (inserts + deletes) definitely exceeds the similarity_detection_limit, and once that many
+        # inserts are encountered, we give up on similarity detection.
+        feature_count_delta = self.feature_count - self.prev_dest_dataset.feature_count
+        self.similarity_detection_insert_limit = (
+            max(self.similarity_detection_limit - feature_count_delta, 0) // 2
+        )
```
craigds, line 142: this formula doesn't seem quite right to me

suppose `similarity_detection_limit=100`, which means if `(deletes+inserts)` exceeds 100 we stop trying to do similarity detection

if feature_count is constant, ie `feature_count_delta=0`, it makes sense: `similarity_detection_insert_limit = 50`, ie if we see 51 inserts that are non-exact matches for previously-deleted things, then we know there are at least 51 deletes too (because we know the feature_count is constant so there must be the same number of deletes as inserts) and so we know we'll exceed our limit of 100 while trying to match them.

if feature_count increases (suppose `feature_count_delta=20`) then this formula calculates `similarity_detection_insert_limit=40`. When we see 41 inserts that are non-exact matches for previously-deleted things, we know there are at least 31 deletes too. However now we _haven't_ exceeded our limit of 100 (41+31=72)

sno/pk_generation.py
```diff
@@ -144,25 +184,138 @@ def _init_schema(self):
...
+
+        return match if match_count == 1 else None
+
+    def _is_similar(self, lhs, rhs):
+        # NOTE: This is one of several possible similarity metrics.
+        # TODO: Add more and make them configurable, if this proves useful.
+
+        dissimilar_count = 0
+
+        for l, r in zip_longest(lhs.values(), rhs.values()):
```
craigds, line 301: are these dicts _guaranteed_ to be in the same order? what guarantees that?

sno/pk_generation.py
```diff
@@ -144,25 +184,138 @@ def _init_schema(self):
...
+    def _assign_pk(self, feature, pk, feature_hash=None):
+        if feature_hash is None:
+            feature_hash = self.schema.hash_feature(feature, without_pk=True)
+
+        feature[self.primary_key] = pk
         self.pk_to_hash[pk] = feature_hash
-        return pk
+        return feature
+
+    def _match_similar_features(self, old_features, new_features):
```
craigds, line 257: I'd rename this to `_match_and_remove_similar_features` to make clear that it mutates both lists

sno/pk_generation.py
```diff
@@ -144,25 +184,138 @@ def _init_schema(self):
...
+            old_features, new_features
+        ):
+            pk = old_feature[self.primary_key]
+            yield self._assign_pk(new_feature, pk)
+            similar_count += 1
+
+        assert len(old_features) == orig_old_features_len - similar_count
+        assert len(new_features) == orig_new_features_len - similar_count
+
+    def _extract_similar_pairs(self, old_features, new_features):
```
craigds, line 272: same here (`_extract_and_remove_similar_pairs`)

sno/pk_generation.py
```diff
@@ -144,25 +184,138 @@ def _init_schema(self):
...
+
+        # Different primary key + one other different field (or fewer) -> similar.
+        return True
+
+    def _find_deleted_features(self, hash_to_unassigned_pks):
+        unassigned_pks = set()
+        for pks in hash_to_unassigned_pks.values():
+            unassigned_pks.update(pks)
+
+        pk_filter = lambda pk: pk in unassigned_pks
```
craigds, line 316: my linter (sublimelinter / pyflakes) would dislike this - if you're assigning a name to a function just use `def`

craigds (changes requested):

sno/init.py
```diff
@@ -132,6 +132,17 @@ def _add_datasets_to_working_copy(repo, *datasets, replace_existing=False):
     is_flag=True,
     help="Replace existing dataset(s) of the same name.",
 )
+@click.option(
+    "--similarity-detection-limit",
+    hidden=True,
+    type=click.INT,
+    default=0,
```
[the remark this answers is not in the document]
olsen232: Changed it to 50.

sno/pk_generation.py
```diff
@@ -60,45 +75,49 @@ class PkGeneratingImportSource(ImportSource):
...
+            if source.schema.pk_columns
+            else PkGeneratingImportSource(source, repo, **kwargs)
+        )
+
+    @classmethod
+    def wrap_sources_if_needed(cls, sources, repo, **kwargs):
+        """Wraps any of the given ImportSources that lack a primary key, returns the result as a new list."""
+        return [cls.wrap_source_if_needed(source, repo, **kwargs) for source in sources]
+
+    def __init__(self, delegate, repo, dest_path=None, similarity_detection_limit=0):
```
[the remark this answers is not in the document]
olsen232: Done

sno/pk_generation.py
```diff
@@ -109,16 +128,37 @@ def load_data_from_repo(self, repo):
...
+            max(self.similarity_detection_limit - feature_count_delta, 0) // 2
+        )
+
+    def _prev_import_dest_tree(self, repo):
+        """Returns the dataset tree that was created the last time this datasource was imported."""
+        if repo.is_empty:
+            return None
+
+        try:
+            # The required dataset tree is in the commit where this file was last modified:
```
rcoup, line 150: No way to do this in libgit2/pygit2?

sno/pk_generation.py
olsen232: Oops - feature_count_delta should have been added, not subtracted - the formula then gives 60. When we see 61 inserts, we know there are at least 41 deletes too - 61 + 41 = 102.
I also had missed checking if abs(feature_delta) exceeds the similarity_detection_limit - the rearranged formula doesn't say anything about whether there can be a negative number of inserts or deletes - of course there can't be, but when given large feature_deltas the formula assumes there can. Math is hard. Anyway, added the check.

sno/pk_generation.py
olsen232: I personally guarantee it. You have my word.
I just added this guarantee to the two main interfaces that define features in Sno:
`base_dataset.py` and `import_source.py`

sno/pk_generation.py
olsen232: Chose _match_similar_features_and_remove

sno/pk_generation.py
olsen232: Chose _pop_similar_pairs

sno/pk_generation.py
olsen232: Done

sno/pk_generation.py
olsen232: There is... it's now done using pygit2. It's about the same amount of code, but I guess it's better not to use git directly if avoidable.

craigds (approved):
nice work! 🎉

PR #341, opened 2021-01-10
Fix build -  start php install from fresh, pyca -> ghchr.io/pyca, brew install --force

olsen232:
Adding a newline to CHANGELOG.md breaks the build for MacOS and Linux. 

So the build is not broken - it is green - but it is also not working: any trivial change breaks the build


craigds:
upstreamed this as pyca/infra#326

PR #342, opened 2021-01-12
Drop support for datasets V1 (except for upgrade)

olsen232:
Note that this PR comes in two commits -

- the first part simply removes or upgrades all tests that rely on V1 datasets working properly: 
    - for tests which come in pairs, with a V1 and a V2 variant, the V1 variant is simply removed
    - for a test which currently only tests against a V1 repo, it is modified to test against a V2 repo
    
    This part took a lot of small changes, but isn't very exciting.

- the second part actually drops support for V1 - trying to run any command except for 
  `sno upgrade` or `sno data version` on a V1 repo, will simply cause an error message to be displayed:

  ```
  Sno repo uses Datasets v1, which is no longer supported.
  The latest version is v2. Use `sno upgrade SOURCE DEST` to upgrade this repo.
  ```

  This part also has lots of small changes, and is also not very exciting, but probably more worth your time.

## Checklist:!


sno/repo.py
```diff
@@ -265,23 +266,27 @@ def _create_with_git_command(cls, cmd, gitdir_path, temp_workdir_path=None):
...
         # Write sno repo version to config:
-        self.config[SnoConfigKeys.SNO_REPOSITORY_VERSION] = str(repo_version)
+        self.config[SnoConfigKeys.SNO_REPOSITORY_VERSION] = str(LATEST_REPO_VERSION)
         # Write working copy config:
         WorkingCopy.write_config(self, wc_path, bare)
 
+    def ensure_latest_version(self):
+        if self.version != LATEST_REPO_VERSION:
+            raise InvalidOperation(
+                f"Sno repo uses Datasets v{self.version}, which is no longer supported.\n"
```
craigds, line 285:
```suggestion
                f"This repo uses Datasets v{self.version}, which is no longer supported.\n"
```

tests/test_data.py
```diff
@@ -10,7 +11,7 @@
             "points",
             id="1",
         ),
-        pytest.param("points2", id="2"),
+        pytest.param("points", id="2"),
```
craigds, line 14: remove this altogether - the test no longer needs to be parametrized

tests/test_diff.py
```diff
@@ -36,7 +36,7 @@ def test_diff_points(
     repo_version, output_format, data_working_copy, geopackage, cli_runner
 ):
     """ diff the working copy against HEAD """
-    data_archive = "points2" if repo_version == 2 else "points"
+    data_archive = "points" if repo_version == 2 else "points"
```
craigds, line 39: should remove the V1_OR_V2 parametrization on this test

tests/test_diff.py
```diff
@@ -269,7 +269,7 @@ def test_diff_reprojection(
...
```
craigds, line 272: and here. all the tests in this file?

craigds (changes requested):

tests/test_data.py
olsen232: Done

tests/test_diff.py
olsen232: Done

tests/test_diff.py
olsen232: Done - this was a bit lax, I must've accidentally "fixed" this file using a search/replace instead of doing it the proper way.

sno/repo.py
olsen232: Done: Sno repo -> This Sno repo.

Just like the output from `sno data version` this output is explicit that even though repos might be "Datasets V1" or "Datasets V2", they are still Sno repos. Seems good to me to keep the word "Sno" in the output.

craigds (approved):
🚀 nice work

Issue #60, opened 2020-04-23
Add a `git reset --hard {commit}` equivalent


craigds:
At present `sno reset` exists, but it actually does something totally unrelated to `git reset`: it throws away uncommited changes in your working tree. That's the equivalent of `git checkout .`

git reset actually doesn't touch your working tree; it modifies the repository head and possibly the index. So it's a totally different command.

In testing sno, I find myself needing a more gittish reset command so I can undo a commit easily. At present I'm using `BRANCH=$(git symbolic-ref --short HEAD) ; sno checkout HEAD^ ; git branch -f $BRANCH HEAD ; sno checkout $BRANCH` , but obviously that's cumbersome.

I assume a reset command would be useful generally, but at least for testing sno it's sorely missed.

UberMouse:
Will this let me peel a commit off so I can reset it? `git reset HEAD~1` equivalent. Or only peel the commit off and discard the changes?

olsen232:
FYI a slightly less magic / undocumented way - of resetting your branch head to a commit of your choice such as HEAD^ is as follows:
```
sno checkout HEAD^    # go back one commit
sno branch -D [branch_name]    # delete branch 
sno checkout -b [branch_name    # recreate branch at this point 
```

Of course we can make a reset command one day - seems like it has never quite been a priority - but at least this workaround is something I'd be okay with putting in public documentation.

olsen232:
@rcoup I want Rob to okay this plan before I (or anyone) does any work on it - The way that `sno reset` does one simple thing at the moment might be an improvement on git's weird mixture of `reset` - hard, soft, and mixed - and `checkout`s for undoing things, and as such might be part of Rob's plan for sno to be simpler than git.

However, seems like a `git reset --hard {commit}` would still be useful, if we can work out where to put it.
Would it make sense if we just added an optional ref arguement to `sno reset`?
So:
`sno reset` is equivalent to `sno checkout HEAD --discard-changes` (as it is now)
`sno reset REF` would be equivalent to `sno checkout REF --discard-changes`, except that it also updates the current branch head to point to REF too. Or to put it another way, it would be the same as `git reset --hard REF`.

We could also add support for --hard and --soft, maybe, except - maybe that's adding too many confusing things again.

craigds:
I find sno's reset command a bit coonfusing because it isn't the same as git's reset command at all:

- git's reset command changes the repo head. _optionally_ the index & working copy can be updated too (`--hard`)
[...]

PR #344, opened 2021-01-13
Add a proper sno reset command.

olsen232:
sno reset COMMIT -> stays on the same branch but sets the branch tip to be COMMIT.

Removes existing sno reset alias which did the same thing as a no-args sno restore.

## Related links:

https://github.com/koordinates/sno/issues/60

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.8.0 (UNRELEASED)
 
-### End of support for Datasets V1
+### Breaking changes
 
  * Backwards compatibility with Datasets V1 ends at Sno 0.8.0 - all Sno commands except `sno upgrade` will no longer work in a V1 repository. Since Datasets V2 has been the default since Sno 0.5.0, most users will be unaffected. Remaining V1 repositories can be upgraded to V2 using `sno upgrade EXISTING_REPO NEW_REPO`, and the ability to upgrade from V1 to V2 continues to be supported indefinitely. [#342](https://github.com/koordinates/sno/pull/342)
+ * `reset` now behaves more like `git reset` - specifically, `sno reset COMMIT` stays on the same branch but sets the branch tip to be `COMMIT`. [#60](https://github.com/koordinates/sno/issues/60)
 
 ### Other changes
 
--- a/sno/checkout.py
+++ b/sno/checkout.py
@@ -276,16 +276,16 @@ def _find_remote_branch_by_name(repo, name):
     "--source",
     "-s",
     help=(
-        "Restore the working tree files with the content from the given tree. "
+        "Restore the working copy with the content from the given tree. "
         "It is common to specify the source tree by naming a commit, branch or "
-        "tag associated with it."
+        "tag associated with it. "
     ),
     default="HEAD",
 )
 @click.argument("pathspec", nargs=-1)
 def restore(ctx, source, pathspec):
     """
-    Restore specified paths in the working tree with some contents from a restore source.
+    Restore specified paths in the working copy with some contents from a restore source.
     """
     repo = ctx.obj.repo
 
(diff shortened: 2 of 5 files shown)


sno/checkout.py
```diff
@@ -307,6 +307,41 @@ def restore(ctx, source, pathspec):
...
+@click.pass_context
+@click.option(
+    "--discard-changes",
+    "--force",
+    "-f",
+    is_flag=True,
+    help="Discard local changes in working copy if necessary",
+)
+@click.argument("refish")
+def reset(ctx, discard_changes, refish):
```
craigds, line 320: So this command is similar to `git reset` but has some differences:

* `sno reset <commit>` - is like `git reset --keep <commit>` - it resets the working copy as well as the branch tip, but refuses to discard local changes 👍 

* `sno reset --discard-changes <commit>` is like `git reset --hard <commit>` - it resets the working copy as well as the branch tip, _and_ discards local changes. This option might not actually be necessary since you can always `sno restore` to discard the local changes before running `sno reset`. I'd tend toward leaving this option out.

I feel like sooner or later we're going to want a soft-reset option to allow the user to leave the working copy alone `--mixed` is the default in `git reset` and comes in handy for some workflows. Probably a more descriptive name wouldn't hurt, e.g. `--ignore-working-copy`. We don't need to distinguish between 'soft' and 'mixed' since we don't have an index area.

@UberMouse mentioned he would find a soft-reset option useful ( https://github.com/koordinates/sno/issues/60#issuecomment-618246032 )

sno/checkout.py
```diff
@@ -307,6 +307,41 @@ def restore(ctx, source, pathspec):
...
+    repo = ctx.obj.repo
+
+    try:
+        commit_or_tree, ref = repo.resolve_refish(refish)
+        commit = commit_or_tree.peel(pygit2.Commit)
+    except (KeyError, pygit2.InvalidSpecError):
+        raise NotFound(f"{refish} is not a commit", exit_code=NO_COMMIT)
+
+    same_commit = repo.head_commit == commit
+    if not discard_changes and not same_commit:
```
craigds, line 333: `and not same_commit` - doesn't that mean that we're not properly checking the repo is not dirty if we happen to already be on the target commit? I think this clause could be removed

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 ### Other changes
 
  * Support for detecting features which have changed slightly during a re-import from a data source without a primary key, and reimporting them with the same primary key as last time so they show as edits as opposed to inserts. [#212](https://github.com/koordinates/sno/issues/212)
+ * `reset` now behaves more like `git reset` - specifically, `sno reset COMMIT` stays on the same branch but sets the branch tip to be `COMMIT`. [#60](https://github.com/koordinates/sno/issues/60)
```
craigds, line 17: I'd put this at the top in a `Breaking changes` section. Maybe along with the datasets-v1 stuff

craigds (commented):
Looks like a good change to me. Could use a comment from @rcoup on the concept, in case he was dead-set on keeping the current `reset` command for some reason.

craigds (commented):
Ah I actually just read the last few comments on #344 which indicate @rcoup  is happy for this change 👍

sno/checkout.py
olsen232: I don't necessarily think of it as a shortcut, but I think `--discard-changes` should stay because
a) it is consistent with existing sno commands, lots of which won't run if there's changes unless you --discard-changes.
b) It's useful to put in error messages. You just told sno to do something but it didn't due to uncommitted changes. But, it suggests that if you run the command again with --discard-changes, then it will do it. Anything more than that - even if it's just "run sno restore first" - seems to me like we are sending the user on a slight detour (even though the total characters typed is no worse) - Sno won't run THIS command unless you first run THAT command, and it's less clear from the name what THAT command does.

I talked to @UberMouse today and he was pretty vague about what he once needed this for, which is unsurprising for a FR that is unimplemented for 6 months - he must have found a workaround. He told me that he uses these things only for dev - his requiring them or not isn't a good indicator if they are generally useful - and right now he was pretty ¯\_(ツ)_/¯ about if a --hard version would be okay for him instead of --soft.

There's a bunch more context at the bottom of #60 - 
- Rob has already okayed this and the removal of the current `sno reset`
- I explain how we don't support mixed in general - that is, any command that updates your HEAD currently doesn't let you do so while keeping uncommitted working copy changes. This could be fixed but it's a more in depth change, and a bit different for us than it is in git (git can't get schema violations by doing a mixed reset, but we can)

sno/checkout.py
olsen232: Your understanding of the logic seems to be correct. This behaviour is consistent with other commands, eg:

Sno checkout:

$ sno checkout HEAD^
Error: You have uncommitted changes in your working copy.
Commit these changes first (`sno commit`) or just discard them by adding the option `--discard-changes`.
$ sno checkout HEAD
(no problem)
$ sno checkout -b new_branch   # kind of like checkout HEAD, but also actually does something
(no problem)

Sno reset:

$ sno reset HEAD^ 
Error: You have uncommitted changes in your working copy.
Commit these changes first (`sno commit`) or just discard them by adding the option `--discard-changes`.
$ sno reset HEAD
(no problem)

Of course, `sno reset HEAD` is a no-op. But that's no reason to make it not work. But in the case of sno reset, there's only two options - either we need to discard-changes, or the command is a no-op (unlike sno checkout above, which can still have useful side effects). So, maybe we shouldn't do this check. But the check is consistent with other commands... what do you think?

CHANGELOG.md
olsen232: Done

sno/checkout.py
craigds: okay. I expect someone'll request a soft-reset eventually, but until then we can ignore it

sno/checkout.py
craigds: I think it's weird, but okay. It's strange because git actually _does_ allow you to checkout and reset all you like while you have local changes, as long as those changes don't conflict with the tip change. Is the reason that we don't allow that, something to do with what you alluded to above about schema changes? Or was it just difficult to reason about so we never bothered to implement it so far?

craigds (approved):

sno/checkout.py
olsen232: Moved discussion offline

PR #347, opened 2021-01-20
Switch GPKG to use sqlalchemy library

olsen232:
## Description

First PR to switch out all DB libraries in favour of sqlalchemy. So far, GPKG is switched.

Lots of little changes - this is the simplest one, easy to understand why it's required everywhere to make it work with the sqlalchemy API
```
dbcur = db.cursor(); dbcur.execute(); dbcur.fetchone(); dbcur.rowcount
-> 
r = db.execute(); r.fetchone(); r.rowcount
```

Here is another very common one:
```
execute("""UPDATE a SET b = ? WHERE c = ?;""", (b, c))
->
execute("""UPDATE a SET b = :b WHERE c = :c;""", {"b": b, "c": c}
```
Since this sqlalchemy's preferred cross-platform format for placeholders - sqlalchemy transforms these to the placeholder's expected by the underlying DB API.

Unfortunately, this means no more """DELETE FROM x WHERE pk IN (?,?,?,?,?)""", [1,2,3,4,5] - I couldn't get IN to work except by using multiple named placeholders, which seemed a bit crazy to me. Now it's """DELETE FROM x WHERE pk = :pk""", [{"pk": 1}, {"pk": 2}, ...]

Another thing which is no longer working - sqlite specific command "INSERT OR REPLACE". Approximated this with a DELETE followed by an INSERT, which will be pretty similar when inside a transaction, but this could also be worth looking into more.

Also lots of changes to tests where this used to work:
```
db = geopackage(path)
with db: # db is a context manager, it opens and closes a transaction
  db.cursor().execute(...)
# But db also works outside of context manager - does autocommits, and probably closes at some stage.
db.cursor().execute(...)
```

The code I have currently have only works from within a context manager, so I fixed the tests so that all the SQL is run from within a context manager, but to make it the test as robust as possible I close it again before running any sno commands. I switched to "with working_copy.session()" a lot since it's the simplest code I have for opening and closing a session.

Still TODO (or to do more of):
- Switch PostGIS to also use sqlalchemy
- Factor out common code from PostGIS / GPKG working copies, move into working_copy/base.py
- Get rid of _portable_insert weirdness in conftest.py - won't need it when we have sqlalchemy
- avoid manually opening connections to GPKG files in tests when sno_repo.working_copy.session() would work instead - it's more portable (eg, it also works for a PostGIS repo, makes it easier to write multi-purpose tests) (
- move simple working_copy stat-getters out of test code and into working_copy.base, to make tests simpler and avoid repetition. Eg instead of: `with working_copy.session() as db: assert HELPER.row_count(db, table_name) == expected` it could simply be: `assert working_copy.row_count(table_name) == expected`
- Look into sqlalchemy support for INSERT OR REPLACE - doing a DELETE followed by an INSERT right now.


sno/working_copy/base.py
```diff
@@ -221,6 +221,17 @@ def assert_db_tree_match(self, tree, *, table_name="*"):
...
+    def tracking_changes_count(self, dataset=None):
+        with self.session() as db:
+            if dataset is not None:
+                r = db.execute(
+                    f"SELECT COUNT(*) FROM {self.TRACKING_TABLE} WHERE table_name=?;",
+                    (dataset.table_name),
+                )
+            else:
+                r = db.execute(f"SELECT COUNT(*) FROM {self.TRACKING_TABLE}")
+            return r.fetchone()[0]
```
craigds, line 233:
```suggestion
            return r.scalar()
```

sno/working_copy/gpkg.py
```diff
@@ -9,10 +9,11 @@
 import click
 import pygit2
 from osgeo import gdal
+from sqlalchemy import table, column
```
craigds, line 12: `table`/`column` names seem very likely to be shadowed in local functions, leading to confusion.

A common pattern is to `import sqlalchemy as sa` and then use `sa.table` / `sa.column` everywhere, might be worth doing?

sno/working_copy/gpkg.py
```diff
@@ -164,14 +138,13 @@ def is_initialised(self):
         if not self.is_created():
             return False
         with self.session() as db:
-            dbcur = db.cursor()
-            dbcur.execute(
+            r = db.execute(
```
craigds, line 141: I always get confused between sessions and connections and engines..

Check this `db` thing - does it have a `scalar` method?

```python
return db.scalar(sql, params)
```
rather tthan .execute()/.fetchone


If not, at a minimum you can call `.scalar()` on the resultset, which just looks a bit tidier than `.fetchone()[0]`

sno/working_copy/gpkg.py
```diff
@@ -690,64 +644,65 @@ def write_full(self, target_tree_or_commit, *datasets, safe=True):
...
 
-                dbcur.execute(
+                db.execute(
                     f"""DELETE FROM {self.TRACKING_TABLE} WHERE table_name=?;""",
                     (table,),
                 )
 
-    def _execute_diff_query(self, dbcur, dataset, feature_filter=None, meta_diff=None):
+    def _execute_dirty_pks_query(self, db, dataset):
+        sql_changed = f"SELECT pk FROM {self.TRACKING_TABLE} " "WHERE table_name=?;"
```
craigds, line 700:
```suggestion
        sql_changed = f"SELECT pk FROM {self.TRACKING_TABLE} WHERE table_name=?;"
```

craigds (commented):
This is a great start, nice job 👍 

Some minor changes noted. I think in general we should be trying to avoid custom SQL where possible (except in backend-specific code) since we're going to be adding extra backends in future.

But that can come in followup steps, and maybe just as part of adding those backends. Just something to keep in mind when adding new / refactoring existing queries

sno/working_copy/base.py
olsen232: Changed a bunch of these in files in this PR... will do a follow up PR where I fix this everywhere.

sno/working_copy/gpkg.py
olsen232: Done

sno/working_copy/gpkg.py
olsen232: Did a bit of a search/replace for fetchone()[0] in this PR but will tidy it up properly in a follow up PR.

rcoup:
> Look into sqlalchemy support for INSERT OR REPLACE - doing a DELETE followed by an INSERT right now.

Seems supported by the [PG](https://docs.sqlalchemy.org/en/14/dialects/postgresql.html?highlight=on_conflict_do_update#sqlalchemy.dialects.postgresql.Insert.on_conflict_do_update) & [SQLite](https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#insert-on-conflict-upsert) dialects at least.

sno/working_copy/gpkg.py
rcoup: Maybe we should use `(sa_)?{engine,conn,session}` names rather than `db`?

sno/working_copy/gpkg.py
```diff
@@ -73,7 +53,7 @@ def _sno_table(self, name, suffix=""):
         n = f"{self.SNO_TABLE_PREFIX}{name}"
         if suffix:
             n += "_" + suffix
-        return gpkg.ident(n)
+        return n
 
     @contextlib.contextmanager
     def session(self, bulk=0):
```
rcoup, line 59: We should be able to ditch this whole context manager and use [SQLAlchemy's transaction handling](https://docs.sqlalchemy.org/en/14/core/connections.html#using-transactions)?

sno/working_copy/gpkg.py
```diff
@@ -89,57 +69,38 @@ def session(self, bulk=0):
...
-                orig_journal = dbcur.execute("PRAGMA journal_mode;").fetchone()[0]
-                orig_locking = dbcur.execute("PRAGMA locking_mode;").fetchone()[0]
-
-                dbcur.execute("PRAGMA synchronous = OFF;")
-                dbcur.execute("PRAGMA cache_size = -1048576;")  # -KiB => 1GiB
+            try:
+                self._session = self.sessionmaker()
 
+                if bulk:
+                    self._session.execute("PRAGMA synchronous = OFF;")
```
rcoup, line 87: see comments earlier, but we should avoid this stuff except for initial WC creation.

rcoup (commented):
Haven't got all the way through yet, but I'm keen to see basically all the SQL text disappear.

Happy to do that in separate PRs, but I think it needs to be done.

sno/geometry.py
```diff
@@ -300,6 +300,11 @@ def wkb_to_gpkg_geom(wkb, **kwargs):
     return ogr_to_gpkg_geom(ogr_geom, **kwargs)
 
 
+def wkt_to_gpkg_geom(wkt, **kwargs):
```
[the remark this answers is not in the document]
olsen232: Should be gone now - it was originally part of this change but I separated it out into a different commit - but left some changes somehow:
https://github.com/koordinates/sno/commit/6a53fc1d01647446848012d458658fe01936a9f2

tests/conftest.py
```diff
@@ -18,12 +18,13 @@
 
 import click
 from click.testing import CliRunner
-from sno.db_util import execute_insert_dict, changes_rowcount
+from sno.db_util import execute_insert_dict
 from sno.geometry import Geometry
+from sno.gpkg import gpkg_engine
+from sno.geometry import wkt_to_gpkg_geom
```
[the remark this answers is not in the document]
olsen232: Should be gone now - it was originally part of this change but I separated it out into a different commit - but left some changes somehow:
https://github.com/koordinates/sno/commit/6a53fc1d01647446848012d458658fe01936a9f2

sno/gpkg.py
```diff
@@ -23,47 +23,17 @@ def param_str(value):
...
-                raise KeyError(key)
-        else:
-            return super().__getitem__(key)
-
-
-def db(path, **kwargs):
-    db = apsw.Connection(str(path), **kwargs)
-    db.setrowtrace(Row)
-    dbcur = db.cursor()
-    dbcur.execute("PRAGMA foreign_keys = ON;")
```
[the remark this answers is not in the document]
olsen232: "PRAGMA foreign_keys = ON" is back

sno/working_copy/gpkg.py
olsen232: Will rename db in a follow up PR - maybe once I have postGIS using sqlalchemy too.

Right now, some of these db's are  "either a sqlalchemy-session, or a psypopg2 cursor" which actually have similar-enough functionality that some things work on both. So, cleaning up naming seems easiest once we have sqlalchemy everywhere and it's clear at least what everything is

sno/working_copy/gpkg.py
olsen232: Done

olsen232:
@rcoup How do you use sqlalchemy to do an "INSERT OR REPLACE"?
I can probably do it right now just by writing: 
```
db.execute("""INSERT OR REPLACE INTO t (a, b, c) VALUES(:a, :b, :c)"""), list_of_rows_as_dicts)
```

Except - we are trying to move away from strings of custom SQL - and - this stops working if column "a" is actually something tricky like "a-a" which doesn't have an unambiguous way of writing as a placeholder - :a-a is not good. So, that is why I am now using this pattern instead, at least for user-defined tables:
```
db.execute(sa.table("t", sa.column("a"), sa.column("b"), sa.column("c")).insert(), list_of_rows_as_dicts)
```

I don't know how to modify the new pattern to be an insert-or-replace, or how to fix the old pattern to work with tricky placeholders (if we should - the new one seems like better sqlalchemy, pity about the missing insert-or-replace).

sno/ogr_import_source.py
```diff
@@ -624,16 +629,15 @@ def features(self):
         Overrides the super implementation for performance reasons
         (it turns out that OGR feature iterators for GPKG are quite slow!)
         """
-        db = gpkg.db(self.ogr_source)
-        dbcur = db.cursor()
-        dbcur.execute(f"SELECT * FROM {self.quote_ident(self.table)};")
-        for gpkg_feature in dbcur:
-            yield self._gpkg_feature_to_sno_feature(gpkg_feature)
+        with self.engine.connect() as db:
+            r = db.execute(f"SELECT * FROM {self.quote_ident(self.table)};")
```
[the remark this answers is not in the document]
olsen232: I won't switch all the statements to be the sqlalchemy-domain-specific-language yet. Except I did the ones for inserting to the user-defined-tables already, but that was actually because it didn't work otherwise due to how sqlalchemy placeholders work.
There'll be plenty of them to do.

sno/gpkg.py
```diff
@@ -207,13 +176,13 @@ def pk(db, table):
 
 
 def geom_cols(db, table):
-    q = db.cursor().execute(
+    q = db.execute(
```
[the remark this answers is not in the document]
olsen232: Yup - happy to switch everything to use sqlalchemy functions. But, it's not done in this PR. I might do it after porting PostGIS to sqlalchemy, since then I can port common WC code to use sqlalchemy properly.

sno/working_copy/base.py
```diff
@@ -596,27 +605,27 @@ def _filter_by_paths(self, datasets, paths):
...
+        self, base_ds, target_ds, db, commit=None, track_changes_as_dirty=False
     ):
         """
         Update the given table in working copy from its current state to target_ds.
         The table must exist in the working copy in the source and continue to exist in the destination,
         and not have any unsupported meta changes - see _is_meta_update_supported.
         base_ds - the dataset that this working copy table is currently based on.
         target_ds - the target desired state for this working copy table.
-        dbcur - database cursor.
+        db - database.
```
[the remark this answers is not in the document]
olsen232: Will update names for all the "db"s once what they are has settled.

sno/working_copy/gpkg.py
olsen232: Had a quick try at this but gave up for right now - took me a while of experimenting before to get it working so that tests pass.
The code will in the end be very similar to this anyway right - all I am hoping is to switch these few lines of code:
```
  db.execute("BEGIN TRANSACTION")
  yield db
  db.commit()
except Exception:
  db.rollback()
```
goes to something like:
```
with connection.begin():
  yield connection
```

which would be nice, but doesn't change the code very much, and I tried it right now and it made some tests fail - possibly because i need to add the workarounds suggested here:
https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#pysqlite-serializable

For now I'll add a TODO and try and close off this PR.

sno/working_copy/gpkg.py
olsen232: @rcoup What counts as initial WC creation - if we need to drop and rewrite an entire table at some point, does that count?

olsen232:
I've been preferring to punt most things in this PR - preferring to get something checked in with passing tests (not necessarily checked into master) and then work from there. It's been a week already I've been working on this PR: next step probably a first pass at postGIS, then begin cleaning up and switching to more portable SQL from there.

Unfortunately the tests are failing like mad on github actions - they pass locally for me. One cause is obvious - macos seems to have a sqlite3 Connection that doesn't have extensions enabled. Not sure about the other platforms.

sno/working_copy/gpkg.py
rcoup: no, because the user might have other data in the DB that isn't committed or isn't Sno-managed at all. I think solely "we're creating a new GPKG file and populating it".

[SQLite docs](https://www.sqlite.org/howtocorrupt.html#_disabling_sync_using_pragmas):
> Setting `PRAGMA synchronous=OFF` can cause the database to go corrupt if there is an operating-system crash or power failure, though this setting is safe from damage due to application crashes.

rcoup:
> Unfortunately the tests are failing like mad on github actions - they pass locally for me. One cause is obvious - macos seems to have a sqlite3 Connection that doesn't have extensions enabled. Not sure about the other platforms.

Yes, the default sqlite3 binary on macOS doesn't have extensions enabled. But we should be using the sqlite library we're building & bunding in `venv/lib/`. I think we probably either need to build Python on macOS, edit the RPATH in the python binaries so it picks up our libsqlite, or build our own pysqlite (linking to our own libsqlite) and get SA to use that instead.

rcoup:
`INSERT OR REPLACE` is a slightly less verbose form of `INSERT... ON CONFLICT DO UPDATE` which SQLAlchemy [supports](https://github.com/koordinates/sno/pull/347#issuecomment-764507308).

I think for _bulk_ queries you need to use the `excluded` approach: https://stackoverflow.com/a/55369524/2662

olsen232:
Some things are fixed:
Failing tests are fixed, mostly thanks to a consistent sqlite version
INSERT OR UPDATE is fixed - now doing this properly instead of DELETE WHERE + INSERT INTO

Something not yet fixed:
PRAGMAs still need some work

Some more things not fixed but can wait until Postgres is also running on sqlalchemy - 
Tidy up how SQL is constructed.
Rename "db" everywhere.
Change "fetchone" to scalar.

rcoup (approved):
LGTM as soon as you're happy connections/PRAGMAs have the same behaviour as before.

rcoup:
> Tidy up how SQL is constructed.
> Rename "db" everywhere.
> Change "fetchone" to scalar.

+ clean up transaction/session handling (and reduce dangerous optimisations)

I think the first priority for SQL has got to be f-strings where we're doing identifier substitution.

PR #348, opened 2021-01-25
Make sqlalchemy + sqlite + spatialite work together

olsen232:
## Description

First PR for sqlalchemy (started over). Proves that sqlalchemy, sqlite+extensions, and spatialite works.
Requires python from brew on macos so that sqlite extensions are enabled, instead of python-3.7.6-macosx10.9.pkg

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,8 +51,6 @@ jobs:
     env:
       CCACHE_DIR: ${{ github.workspace }}/.ccache
       CCACHE_COMPRESS: "1"
-      PY3_PKG: python-3.7.6-macosx10.9.pkg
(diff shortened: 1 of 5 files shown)


craigds (commented):
I _think_ this looks fine but I'm a bit confused about how the homebrew runtime dependencies work. We don't have any dependencies specified in https://github.com/koordinates/homebrew-sno/blob/master/Casks/sno.rb

Probably mostly a question for @rcoup : Will this change cause the PyInstaller package to vendor in the python from homebrew?

craigds (commented):
solid giffing 💯

olsen232:
This PR actually abandoned - started a new sqlalchemy branch that doesn't yet contain this commit.

PR #350, opened 2021-02-02
Port postgis WC to sqlalchemy

olsen232:
## Description

PostGIS now works using sqlalchemy. Identifiers are now properly quoted. This is now far enough along that it could be merged to master once through review.

Still TODO in further PR(s):
- Completely remove apsw dependency.
- Rename "db" to "session" or similar everywhere (and standardise on it being either the first or last function arg).
- Maybe switch some more code from "SELECT" to "select()" - right now such functions are only being used for inserts or updates. Unclear to me if there is much benefit, the end result will be that SQL that is currently hard-coded will be generated at runtime, not sure if it will be more readable...


sno/working_copy/postgis.py
```diff
@@ -166,61 +129,53 @@ def _sno_table(self, sno_table):
...
     def _table_identifier(self, table):
-        return Identifier(self.schema, table)
+        # TODO - use sqlalchemy not strings
+        return f'"{self.schema}"."{table}"'
 
     @contextlib.contextmanager
     def session(self, bulk=0):
         """
-        Context manager for PostgreSQL DB sessions, yields a connection object inside a transaction
+        Context manager for GeoPackage DB sessions, yields a connection object inside a transaction
```
rcoup, line 138: ?

sno/working_copy/postgis.py
```diff
@@ -166,61 +129,53 @@ def _sno_table(self, sno_table):
...
-                del self._db
-                L.debug("session: new/done")
+                self._session.close()
+                del self._session
+                L.debug(f"session(bulk={bulk}): new/done")
 
     def _configure_output(self, db):
         """Output timestamptz in UTC, and output interval as ISO 8601 duration."""
-        dbcur = db.cursor()
-        dbcur.execute("SET timezone='UTC';")
```
rcoup, line 209: I think this one still needs setting? AFAICS:
- libpq will use `$PGTZ` by default
- otherwise, the server will use whatever the _database_ is set to
- we want UTC...

(should be in `sno.sqlalchemy.postgis_engine()` though)

sno/working_copy/postgis.py
```diff
@@ -166,61 +129,53 @@ def _sno_table(self, sno_table):
...
-        dbcur.execute("SET intervalstyle='iso_8601';")
 
     def _register_geometry_type(self, db):
+        # TODO
```
rcoup, line 174: isn't this done in `sno.sqlalchemy.postgis_engine()`?

sno/working_copy/postgis.py
```diff
@@ -166,61 +129,53 @@ def _sno_table(self, sno_table):
...
+            L.debug(f"session: new...")
 
             try:
-                yield self._db
-            except:  # noqa
-                self._db.rollback()
+                # TODO - use "with" for sqlalchemy transaction.
+                self._session = self.sessionmaker()
+                self._configure_output(self._session)
+                # self._register_geometry_type(self._session)
```
rcoup, line 157:
```suggestion
```

Feels like these are/should be done in `sno.sqlalchemy.postgis_engine()`?

sno/working_copy/postgis.py
```diff
@@ -231,86 +186,73 @@ def is_created(self):
...
-                    key TEXT NOT NULL,
-                    value TEXT NULL,
-                    PRIMARY KEY (table_name, key)
-                );
-                """
-                ).format(self.STATE_TABLE)
+            db.execute(f"""CREATE SCHEMA IF NOT EXISTS "{self.schema}";""")
+
+            metadata = sqlalchemy.MetaData(schema=self.schema)
+            sqlalchemy.schema.Table(
```
rcoup, line 232: save these into the base class? They can be used for SQLite as well...

sno/working_copy/postgis.py
```diff
@@ -409,27 +339,21 @@ def write_meta_crs(self, dataset):
...
-                    ),
-                    tuple(row.values()),
-                )
+            # We do not automatically overwrite a CRS if it seems likely to be one
+            # of the Postgis builtin definitions - Postgis has lots of EPSG and ESRI
+            # definitions built-in, plus the 900913 (GOOGLE) definition.
+            # See POSTGIS_WC.md for help on working with CRS definitions in a Postgis WC.
+            db.execute(
+                """
+                INSERT INTO public.spatial_ref_sys AS SRS (srid, auth_name, auth_srid, srtext, proj4text)
```
rcoup, line 348: `spatial_ref_sys` isn't necessarily in `public`... it can be installed into any schema.

That said, if someone does move it into a `postgis` schema or wherever, it's IMO up to them to make sure it's in the db/user `search_path` — so we can always access it _without_ a schema qualifier. Same applies to `ST_Something()` functions/etc.

sno/working_copy/base.py
```diff
@@ -381,22 +381,92 @@ def _execute_dirty_pks_query(self, db, dataset):
...
+    def _execute_dirty_rows_query(
+        self, db, dataset, feature_filter=None, meta_diff=None
+    ):
         """
         Does a join on the tracking table and the table for the given dataset, and returns a result
         containing all the rows that have been inserted / updated / deleted.
         """
-        raise NotImplementedError()
+        # TODO - use sqlalchemy, move this if-block.
+        if hasattr(self, "schema"):
```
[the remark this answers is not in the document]
rcoup: Or... `main` is the [default DB schema name](https://sqlite.org/lang_naming.html) in SQLite, would using that make query construction simpler?

rcoup (changes requested):
Looks good! Some minor comments.

Main thing I think is that `Identifier`/`quote_ident()`/etc also escape eg: `"`, so either need to fix f-strings as you describe in the intro; or at least use something slightly more robust in the meantime if you want this merged first?

Would something similar to [sqlalchemy-utils `quote()`](https://sqlalchemy-utils.readthedocs.io/en/latest/orm_helpers.html#quote) work in the interim? ([implementation](https://sqlalchemy-utils.readthedocs.io/en/latest/_modules/sqlalchemy_utils/functions/orm.html#quote))

sno/working_copy/base.py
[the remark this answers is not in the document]
olsen232: Yeah this was a hack while I was developing - sorry for leaving it in the PR, even in an "FYI" PR like this one. The proper solution is to let subclasses override how they do the two things here that need to be different here, eg:
```
self.table_identifier(dataset)  # Quote dataset.table_name, with or without a schema 
self.CAST_TO_TEXT  # Add a "::text" to something if needed - neither needed nor allowed in SQLite.
```

sno/working_copy/base.py
```diff
@@ -381,22 +381,92 @@ def _execute_dirty_pks_query(self, db, dataset):
...
+            schema = dataset.schema
+
+        pk_field = schema.pk_columns[0].name
+        col_names = ",".join([f'TAB."{col.name}"' for col in schema])
+
+        diff_sql = f"""
+            SELECT
+                TRA.pk AS ".__track_pk",
+                {col_names}
+            FROM {self.TRACKING_TABLE} TRA LEFT OUTER JOIN {table_identifier} TAB
```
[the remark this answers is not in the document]
olsen232: I would imagine the local variable would have precedence over a global table. Will have to check, since we are using these elsewhere...

This is a good candidate for using a sqlalchemy select() function and let it build the query itself. Will probably do that, then we don't need aliases anyway.

sno/working_copy/base.py
```diff
@@ -381,22 +381,92 @@ def _execute_dirty_pks_query(self, db, dataset):
...
+                ):
+                    db.execute(
+                        f"DELETE FROM {self.TRACKING_TABLE} WHERE table_name=:table_name;",
+                        {"table_name": table},
+                    )
+                    continue
+
+                pks = dataset_filter.get("feature", ())
+                db.execute(
+                    f"DELETE FROM {self.TRACKING_TABLE} WHERE table_name=:table_name AND pk=:pk;",
```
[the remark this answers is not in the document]
olsen232: Oh good. It's pretty ugly like this. Will change.

sno/working_copy/base.py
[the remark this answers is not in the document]
olsen232: I'll rename schema to db_schema, in a separate PR

sno/working_copy/postgis.py
olsen232: Yeah this is now back in sqlalchemy.postgis_engine

sno/working_copy/postgis.py
olsen232: Yes - deleted this dead code.

sno/working_copy/postgis.py
olsen232: Yes - dead code again.

sno/working_copy/postgis.py
olsen232: I've moved these into another file:

sno/working_copy/table_defs.py

I think you'll like it (or at least the general idea of it)

sno/working_copy/postgis.py
olsen232: Ok, I'll remove the public from all PostGIS metadata-table identifiers.

sno/working_copy/postgis.py
```diff
@@ -751,96 +629,6 @@ def _db_geom_to_gpkg_geom(self, g):
...
-        )
-        params = [table]
-
-        if feature_filter is not UNFILTERED:
-            diff_sql += SQL("\nAND {}.pk IN %s").format(self.TRACKING_TABLE)
-            params.append(tuple([str(pk) for pk in feature_filter]))
-
-        dbcur = db.cursor()
-        dbcur.execute(diff_sql, params)
-        return dbcur.fetchall()
```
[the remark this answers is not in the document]
olsen232: It mostly doesn't make that much difference, but I'll standardise on returning the resultset - there could be efficiency gains (depending on what the client wants to do) and it has more information (eg r.rowcount, r.lastrowid).

sno/working_copy/postgis.py
```diff
@@ -33,54 +32,16 @@
...
-    1266: "TIME",
-    704: "INTERVAL",
-    1186: "INTERVAL",
-    1700: "DECIMAL",
-}
-
-for oid in ADAPT_TO_STR_TYPES:
-    t = new_type((oid,), ADAPT_TO_STR_TYPES[oid], adapt_to_string)
-    psycopg2.extensions.register_type(t)
+def insert_or_replace_command(table_name, col_names, schema=None, pk_col_names=None):
```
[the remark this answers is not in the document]
olsen232: Maybe - I'll try it out. The "on_conflict()" functionality that I'm using is part of the sqlalchemy code for the Postgres dialact, but it will probably work regardless.

This code has moved around a bit as I've started to create table definitions ahead of time, instead of creating them on the fly when they are needed.

olsen232:
I have started using sqlalchemy's quote - in fact I use sqlalchemy's IdentifierPreparer directly, which is what sqlalchemy.quote uses. This means I can ask for the fully-qualified name of a table-object directly, or of a column-object, which might be more robust than just asking sqlalchemy to quote identifiers and hoping it knows what I mean.

sno/working_copy/base.py
```diff
@@ -39,18 +44,104 @@ def __str__(self):
...
+
+    @property
+    @functools.lru_cache(maxsize=1)
+    def SNO_STATE_NAME(self):
+        """The table name of sno_track table, not including the schema."""
+        return self.sno_tables.sno_state.name
+
+    @property
+    def CAST_TO_TEXT(self):
+        """Returns a SQL suffix that casts an expression to text."""
```
craigds, line 97: postgres' `::TEXT` is a bit non-standard, the standard SQL way is `CAST(fieldname AS TEXT)`, which afaik works in both sqlite and postgres. But sqlalchemy also has a way of producing it: https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.cast

So... the existence of this property seems odd/unnecessary

sno/working_copy/base.py
```diff
@@ -452,15 +642,157 @@ def update_state_table_tree(self, tree):
...
+        if not pk_list:
+            return 0
+
+        pk_column = self.preparer.quote(dataset.primary_key)
+        sql = f"""DELETE FROM {self.table_identifier(dataset)} WHERE {pk_column} IN :pks;"""
+        feat_count = 0
+        CHUNK_SIZE = 100
+        for pks in self._chunk(pk_list, CHUNK_SIZE):
+            t = sqlalchemy.text(sql)
+            t = t.bindparams(sqlalchemy.bindparam("pks", expanding=True))
```
craigds, line 775: I think you can move these lines outside the loop, saves a bit of computation

sno/working_copy/gpkg.py
```diff
@@ -438,119 +398,60 @@ def _drop_spatial_index(self, dbcur, dataset):
...
+        self._create_triggers(dbcur, dataset)
 
     def _create_triggers(self, db, dataset):
-        table = dataset.table_name
-        pkf = gpkg.ident(dataset.primary_key)
-        ts = gpkg.param_str(table)
+        table_identifier = self.table_identifier(dataset)
+        pk_column = self.quote(dataset.primary_key)
+
+        # SQLite doesn't let you do param substitutions in CREATE TRIGGER:
```
craigds, line 415: Wow: https://stackoverflow.com/questions/29032750/using-sqlite-trigger-with-argument

that's madness

sno/working_copy/postgis_adapter.py
```diff
@@ -58,16 +57,13 @@ def v2_schema_to_postgis_spec(schema, v2_obj):
     Generate the SQL CREATE TABLE spec from a V2 object eg:
     'fid INTEGER, geom GEOMETRY(POINT,2136), desc VARCHAR(128), PRIMARY KEY(fid)'
     """
-    result = [
-        SQL("{} {}").format(Identifier(col.name), SQL(v2_type_to_pg_type(col, v2_obj)))
-        for col in schema
-    ]
+    result = [f'"{col.name}" {v2_type_to_pg_type(col, v2_obj)}' for col in schema]
```
craigds, line 60: this change seems wrong - column identifiers aren't escaped any more (`Identifier`)

sno/working_copy/postgis_adapter.py
```diff
@@ -58,16 +57,13 @@ def v2_schema_to_postgis_spec(schema, v2_obj):
...
 
     if schema.pk_columns:
-        pk_col_names = (Identifier(col.name) for col in schema.pk_columns)
-        result.append(SQL("PRIMARY KEY({})").format(SQL(", ").join(pk_col_names)))
+        pk_col_names = ", ".join((f'"{col.name}"' for col in schema.pk_columns))
```
craigds, line 63: and this one

craigds (changes requested):
Looks great - having all that stuff moved to generic implementations in the base class is awesome.

I think we need to go further, replacing more custom SQL with SA constructs wherever possible, to safeguard against SQL injection. That can come in a later PR though

sno/working_copy/base.py
olsen232: Removed, added `CAST(fieldname AS TEXT)`

sno/working_copy/base.py
olsen232: Done

sno/working_copy/gpkg.py
olsen232: Ack

sno/working_copy/postgis_adapter.py
olsen232: Fixed

sno/working_copy/postgis_adapter.py
olsen232: Fixed

craigds (approved):

PR #352, opened 2021-02-05
init: Add `--initial-branch` switch

craigds:
## Description

* adds a `--initial-branch` switch to the `init` command.
* Respect git's `init.defaultBranch` config if that switch isn't supplied
* update all the test data to use `main` for the default branch

This doesn't actually change the _default_ branch (it's still set to `master`), but makes it easier to mess with at init time.

## Related links:

none

## Checklist:


olsen232 (approved):
Looks good, as long as you can fix the end-to-end test

PR #361, opened 2021-02-23
Support two different locations for XML metadata

olsen232:
## Description

As of this commit, we support two different locations for XML metadata:
`<dataset>/.sno-dataset/meta/metadata/dataset.json`
`<dataset>/metadata.xml`

A dataset can have one or the other, not both.
Committing changes and patches continues to update the XML in its current location.
Imports now import XML to `metadata.xml`.
`metadata.xml` only supports one XML attachment per dataset - trying to add another causes the commit to fail with an appropriate error.

This is part of a change to eventually treat metadata.xml as an attachment, which can be checked out and edited like in git.

## Related links:

Internally, see The Friday Compromise 
## Checklist:


sno/meta_items.py
```diff
@@ -5,9 +5,15 @@
     "title",
     "description",
     "schema.json",
-    "metadata/dataset.json",
+    "metadata.xml",
+    "metadata/dataset.json",  # deprecated - imported as metadata.json going forward.
```
craigds, line 9:
```suggestion
    "metadata/dataset.json",  # deprecated - imported as metadata.xml going forward.
```

craigds (approved):

tests/data/patches/polygons.snopatch
```diff
@@ -8,12 +8,8 @@
...
...ode&gt;&lt;/gmd:classification&gt;&lt;/gmd:MD_SecurityConstraints&gt;&lt;/gmd:metadataConstraints&gt;&lt;gmd:metadataConstraints&gt;&lt;gmd:MD_LegalConstraints&gt;&lt;gmd:useLimitation&gt;&lt;gco:CharacterString&gt;Crown copyright reserved&lt;/gco:CharacterString&gt;&lt;/gmd:useLimitation&gt;&lt;gmd:useConstraints&gt;&lt;gmd:MD_RestrictionCode codeList=\"http://asdd.ga.gov.au/asdd/profileinfo/gmxCodelists.xml#MD_RestrictionCode\" codeListValue=\"copyright\"&gt;copyright&lt;/gmd:MD_RestrictionCode&gt;&lt;/gmd:useConstraints&gt;&lt;/gmd:MD_LegalConstraints&gt;&lt;/gmd:metadataConstraints&gt;&lt;gmd:metadataConstraints&gt;&lt;gmd:MD_LegalConstraints&gt;&lt;gmd:useLimitation&gt;&lt;gco:CharacterString&gt;Released under Creative Commons Attribution 4.0 International&lt;/gco:CharacterString&gt;&lt;/gmd:useLimitation&gt;&lt;gmd:useConstraints&gt;&lt;gmd:MD_RestrictionCode codeList=\"http://asdd.ga.gov.au/asdd/profileinfo/gmxCodelists.xml#MD_RestrictionCode\" codeListValue=\"license\"&gt;license&lt;/gmd:MD_RestrictionCode&gt;&lt;/gmd:useConstraints&gt;&lt;/gmd:MD_LegalConstraints&gt;&lt;/gmd:metadataConstraints&gt;&lt;/gmd:MD_Metadata&gt;</MDI>\n  </Metadata>\n</GDALMultiDomainMetadata>\n"
```
rcoup, line 12: This isn't right...

GDAL's "metadata" is a wrapper/container - in this case the metadata root should be the `<gmd:MD_Metadata>` element, and the `md_standard_uri` field in `gpkg_metadata` should be `http://www.isotc211.org/2005/gmd`

sno/working_copy/gpkg_adapter.py
```diff
@@ -474,29 +485,111 @@ def json_to_gpkg_metadata(v2_metadata_json, table_name, reference=False):
...
+
+
+def _get_uri_from_xml(xml_metadata):
+    from xml.dom.minidom import parseString
+
+    try:
+        doc = parseString(xml_metadata)
+        element = doc.documentElement
+        return (
+            _KNOWN_METADATA_URIS.get(element.tagName)
```
rcoup, line 500: I don't think we should be using GDALs metadata — is this an artifact of the OGR importer? It's basically a wrapper of actual metadata standards.

Need tests for this.

sno/working_copy/gpkg_adapter.py
```diff
@@ -474,29 +485,111 @@ def json_to_gpkg_metadata(v2_metadata_json, table_name, reference=False):
...
+            or element.namespaceURI
+            or "(unknown)"
+        )
+    except Exception:
+        return "(unknown)"
+
+
+def xml_to_gpkg_metadata(metadata_xml, table_name, reference=False):
+    if reference:
+        timestamp = datetime_to_iso8601_utc(datetime.now())
```
rcoup, line 511: Feels like this should probably be the commit timestamp of the last change to the metadata blob?

rcoup (changes requested):
Looks good

1. Something important I spotted with respect to "GDAL metadata", and metadata standard-detection in general.
2. What's the interim method to actually commit/update `metadata.xml`?

sno/working_copy/gpkg_adapter.py
olsen232: We don't "use" GDAL metadata as such. We accept any XML metadata that is attached and store it unchanged. The OGR importer doesn't provide any metadata (as far as I know, in any case we don't use it) - it's up to us to extract the metadata. We do this using the same code we use for extracting metadata from the GPKG working copy.

The code for guessing the URI is now moved from the importer to here. I can add some tests.

tests/data/patches/polygons.snopatch
olsen232: This is just the metadata that is attached to the test data. As far as I know, it's been like this since you created the test data over a year ago. This particular test actually recreates the polygons test repo by applying a patch - so the patch contains all the same data as the polygons test repo.

I can try to unwrap the XML in the test dta manually if you want. I don't know why it's how it is, but I don't have any evidence that it hasn't been accurately preserved by Sno - kept exactly as it was supplied to Sno long ago.

sno/working_copy/gpkg_adapter.py
rcoup: In principle yes, we accept whatever, but GDAL is a wrapper around _actual_ metadata :)

If there's no path to adding it at import time I think it's just a case of fixing the test data. I guess I just overlooked it.

sno/meta_items.py
olsen232: Done

tests/data/patches/polygons.snopatch
olsen232: Fixed XML for polygons test repo in particular, in a separate commit.

Weird test data is unique to polygons test repo and doesn't seem to point to any particular Sno issue.

sno/working_copy/gpkg_adapter.py
olsen232: As above:
"Fixed XML for polygons test repo in particular, in a separate commit.
Weird test data is unique to polygons test repo and doesn't seem to point to any particular Sno issue."

Code for guessing URI still has no tests - I will add tests in a separate PR.

sno/working_copy/gpkg_adapter.py
olsen232: Not yet done - this is a bigger change than you would think since it means the working-copy generated is no longer a function of a particular git tree, but of the entire repo history - need to wire through some more data to the right places.

Will do this in a separate PR.

rcoup: Is it worth doing?

PR #362, opened 2021-02-25
Add support for SQL Server working copy

olsen232:
## Description

A working implementation of a SQL server working copy.
Tests are passing - checkouts, diffs, and commits are working.

TODO:
Documentation is not yet updated.
Dependencies are not yet updated - so, this works for me locally, and all new sqlserver tests are passing, but this won't work for anyone else because they don't have the appropriate SQL server driver etc. I installed these manually to my local Sno environment. However, this won't be immediately obvious in github actions, since all of the relevant tests will be skipped anyway since no SQL server database is configured.

### DIFF
--- a/sno/geometry.py
+++ b/sno/geometry.py
@@ -57,6 +57,11 @@ class Geometry(bytes):
         crs_id_bytes = struct.pack("<i", crs_id)
         return Geometry.of(self[:4] + crs_id_bytes + self[8:])
 
+    @property
+    def crs_id(self):
+        wkb_offset, is_le, crs_id = parse_gpkg_geom(self)
+        return crs_id
+
     @classmethod
     def from_wkt(cls, wkt):
         return wkt_to_gpkg_geom(wkt)
@@ -290,12 +295,18 @@ def gpkg_geom_to_ogr(gpkg_geom, parse_crs=False):
     return geom
 
 
-def wkt_to_gpkg_geom(wkb, **kwargs):
-    ogr_geom = ogr.CreateGeometryFromWkt(wkb)
+def wkt_to_gpkg_geom(wkt, **kwargs):
+    if wkt is None:
+        return None
+
+    ogr_geom = ogr.CreateGeometryFromWkt(wkt)
     return ogr_to_gpkg_geom(ogr_geom, **kwargs)
 
 
 def wkb_to_gpkg_geom(wkb, **kwargs):
+    if wkb is None:
+        return None
+
     ogr_geom = ogr.CreateGeometryFromWkb(wkb)
     return ogr_to_gpkg_geom(ogr_geom, **kwargs)
 
--- a/sno/geometry.py
+++ b/sno/geometry.py
@@ -59,6 +59,11 @@ class Geometry(bytes):
 
     @property
     def crs_id(self):
+        """
+        Returns the CRS ID as it is embedded in the GPKG header - before the WKB.
+        Note that datasets V2 zeroes this field before committing,
+        so will return zero when called on Geometry where it has been zeroed.
+        """
         wkb_offset, is_le, crs_id = parse_gpkg_geom(self)
         return crs_id
 
@@ -296,6 +301,7 @@ def gpkg_geom_to_ogr(gpkg_geom, parse_crs=False):
 
 
 def wkt_to_gpkg_geom(wkt, **kwargs):
+    """Given a well-known-text string, returns a GPKG Geometry object."""
     if wkt is None:
         return None
 
@@ -304,6 +310,7 @@ def wkt_to_gpkg_geom(wkt, **kwargs):
 
 
 def wkb_to_gpkg_geom(wkb, **kwargs):
+    """Given a well-known-binary bytestring, returns a GPKG Geometry object."""
     if wkb is None:
         return None
 
@@ -312,6 +319,7 @@ def wkb_to_gpkg_geom(wkb, **kwargs):
 
 
 def hex_wkb_to_gpkg_geom(hex_wkb, **kwargs):
+    """Given a hex-encoded well-known-binary bytestring, returns a GPKG Geometry object."""
     if hex_wkb is None:
         return None
 
--- a/sno/sqlalchemy.py
+++ b/sno/sqlalchemy.py
@@ -1,3 +1,8 @@
+import re
+import socket
+from urllib.parse import urlsplit, urlunsplit
+
+
 import sqlalchemy
 from pysqlite3 import dbapi2 as sqlite
 import psycopg2
@@ -95,7 +100,51 @@ def postgis_engine(pgurl):
             geometry_type = new_type((r[0],), "GEOMETRY", _adapt_geometry_from_pg)
             register_type(geometry_type, psycopg2_conn)
 
+    pgurl = _append_query_to_url(pgurl, {"fallback_application_name": "sno"})
+
     engine = sqlalchemy.create_engine(pgurl, module=psycopg2)
     sqlalchemy.event.listen(engine, "connect", _on_connect)
 
     return engine
+
+
+CANONICAL_SQL_SERVER_SCHEME = "mssql"
+INTERNAL_SQL_SERVER_SCHEME = "mssql+pyodbc"
+SQL_SERVER_DRIVER_LIB = "ODBC+Driver+17+for+SQL+Server"
+
+
+def sqlserver_engine(msurl):
+    url = urlsplit(msurl)
+    if url.scheme != CANONICAL_SQL_SERVER_SCHEME:
+        raise ValueError("Expecting mssql://")
+
+    # SQL server driver is fussy - doesn't like localhost, prefers 127.0.0.1
+    url_netloc = re.sub(r"\blocalhost\b", _replace_with_localhost, url.netloc)
+
+    url_query = _append_to_query(
+        url.query, {"driver": SQL_SERVER_DRIVER_LIB, "Application+Name": "sno"}
+    )
+
+    msurl = urlunsplit(
+        [INTERNAL_SQL_SERVER_SCHEME, url_netloc, url.path, url_query, ""]
+    )
+
+    engine = sqlalchemy.create_engine(msurl)
+    return engine
+
+
+def _replace_with_localhost(*args, **kwargs):
+    return socket.gethostbyname("localhost")
+
+
+def _append_query_to_url(uri, query_dict):
+    url = urlsplit(uri)
+    url_query = _append_to_query(url.query, query_dict)
+    return urlunsplit([url.scheme, url.netloc, url.path, url_query, ""])
+
+
+def _append_to_query(url_query, query_dict):
+    for key, value in query_dict.items():
+        if key not in url_query:
+            url_query = "&".join(filter(None, [url_query, f"{key}={value}"]))
+    return url_query
--- a/sno/sqlalchemy.py
+++ b/sno/sqlalchemy.py
@@ -1,6 +1,6 @@
 import re
 import socket
-from urllib.parse import urlsplit, urlunsplit
+from urllib.parse import urlsplit, urlunsplit, urlencode, parse_qs
 
 
 import sqlalchemy
@@ -11,6 +11,7 @@ from psycopg2.extensions import Binary, new_type, register_adapter, register_typ
 
 from sno import spatialite_path
 from sno.geometry import Geometry
+from sno.exceptions import NotFound
 
 
 def gpkg_engine(path):
@@ -110,7 +111,6 @@ def postgis_engine(pgurl):
 
 CANONICAL_SQL_SERVER_SCHEME = "mssql"
 INTERNAL_SQL_SERVER_SCHEME = "mssql+pyodbc"
-SQL_SERVER_DRIVER_LIB = "ODBC+Driver+17+for+SQL+Server"
 
 
 def sqlserver_engine(msurl):
@@ -122,7 +122,7 @@ def sqlserver_engine(msurl):
     url_netloc = re.sub(r"\blocalhost\b", _replace_with_localhost, url.netloc)
 
     url_query = _append_to_query(
-        url.query, {"driver": SQL_SERVER_DRIVER_LIB, "Application+Name": "sno"}
+        url.query, {"driver": get_sqlserver_driver(), "Application Name": "sno"}
     )
 
     msurl = urlunsplit(
@@ -133,18 +133,30 @@ def sqlserver_engine(msurl):
     return engine
 
 
+def get_sqlserver_driver():
+    import pyodbc
+
+    drivers = [
+        d for d in pyodbc.drivers() if re.search("SQL Server", d, flags=re.IGNORECASE)
+    ]
+    if not drivers:
+        drivers = pyodbc.drivers()
+    if not drivers:
+        raise NotFound("SQL Server driver was not found")
+    return sorted(drivers)[-1]  # Latest driver
+
+
 def _replace_with_localhost(*args, **kwargs):
     return socket.gethostbyname("localhost")
 
 
-def _append_query_to_url(uri, query_dict):
+def _append_query_to_url(uri, new_query_dict):
     url = urlsplit(uri)
-    url_query = _append_to_query(url.query, query_dict)
+    url_query = _append_to_query(url.query, new_query_dict)
     return urlunsplit([url.scheme, url.netloc, url.path, url_query, ""])
 
 
-def _append_to_query(url_query, query_dict):
-    for key, value in query_dict.items():
-        if key not in url_query:
-            url_query = "&".join(filter(None, [url_query, f"{key}={value}"]))
-    return url_query
+def _append_to_query(existing_query, new_query_dict):
+    query_dict = parse_qs(existing_query)
+    # ignore new keys if they're already set in the querystring
+    return urlencode({**new_query_dict, **query_dict})
--- a/sno/working_copy/base.py
+++ b/sno/working_copy/base.py
@@ -9,8 +9,6 @@ from enum import Enum, auto
 import click
 import pygit2
 import sqlalchemy
-from sqlalchemy.dialects.postgresql import insert as postgresql_insert
-
 
 from sno.base_dataset import BaseDataset
 from sno.diff_structs import RepoDiff, DatasetDiff, DeltaDiff, Delta
@@ -39,6 +37,8 @@ class WorkingCopyType(Enum):
         path = str(path)
         if path.startswith("postgresql:"):
             return WorkingCopyType.POSTGIS
+        elif path.startswith("mssql:"):
+            return WorkingCopyType.SQL_SERVER
         elif path.lower().endswith(".gpkg"):
             return WorkingCopyType.GPKG
         elif allow_invalid:
@@ -61,6 +61,10 @@ class WorkingCopyType(Enum):
             from .postgis import WorkingCopy_Postgis
 
             return WorkingCopy_Postgis
+        elif self is WorkingCopyType.SQL_SERVER:
+            from .sqlserver import WorkingCopy_SqlServer
+
+            return WorkingCopy_SqlServer
         raise RuntimeError("Invalid WorkingCopyType")
 
 
@@ -149,32 +153,36 @@ class WorkingCopy:
         )
         return self.preparer.format_table(sqlalchemy_table)
 
+    def _table_def_for_column_schema(self, col, dataset):
+        # This is just used for selects/inserts/updates - we don't need the full type information.
+        # We only need type information if some automatic conversion needs to happen on read or write.
+        # TODO: return the full type information so we can also use it for CREATE TABLE.
+        return sqlalchemy.column(col.name)
+
     @functools.lru_cache()
-    def table_def_for_dataset(self, dataset):
+    def _table_def_for_dataset(self, dataset, schema=None):
         """
         A minimal table definition for a dataset.
         Specifies table and column names only - the minimum required such that an insert() or update() will work.
         """
+        schema = schema or dataset.schema
         return sqlalchemy.table(
             dataset.table_name,
-            *[sqlalchemy.column(c.name) for c in dataset.schema.columns],
+            *[self._table_def_for_column_schema(c, dataset) for c in schema],
             schema=self.db_schema,
         )
 
-    def insert_into_dataset(self, dataset):
+    def _insert_into_dataset(self, dataset):
         """Returns a SQL command for inserting features into the table for that dataset."""
-        return self.table_def_for_dataset(dataset).insert()
+        return self._table_def_for_dataset(dataset).insert()
 
-    def insert_or_replace_into_dataset(self, dataset):
+    def _insert_or_replace_into_dataset(self, dataset):
         """
         Returns a SQL command for inserting/replacing features that may or may not already exist in the table
         for that dataset.
         """
-        # Even though this uses the postgresql_insert sqlalchemy feature, it results in generic SQL.
-        pk_col_names = [c.name for c in dataset.schema.pk_columns]
-        stmt = postgresql_insert(self.table_def_for_dataset(dataset))
-        update_dict = {c.name: c for c in stmt.excluded if c.name not in pk_col_names}
-        return stmt.on_conflict_do_update(index_elements=pk_col_names, set_=update_dict)
+        # Its not possible to do this in a DB agnostic way.
+        raise NotImplementedError()
 
     @classmethod
     def get(cls, repo, *, allow_uncreated=False, allow_invalid_state=False):
@@ -342,17 +350,16 @@ class WorkingCopy:
         """
         raise NotImplementedError()
 
-    def get_db_tree(self, table_name="*"):
+    def get_db_tree(self):
         """Returns the hex tree ID from the state table."""
         with self.session() as sess:
             return sess.scalar(
-                f"""SELECT value FROM {self.SNO_STATE} WHERE table_name=:table_name AND key='tree';""",
-                {"table_name": table_name},
+                f"""SELECT value FROM {self.SNO_STATE} WHERE "table_name"='*' AND "key"='tree';""",
             )
 
-    def assert_db_tree_match(self, tree, *, table_name="*"):
+    def assert_db_tree_match(self, tree):
         """Raises a Mismatch if sno_state refers to a different tree and not the given tree."""
-        wc_tree_id = self.get_db_tree(table_name)
+        wc_tree_id = self.get_db_tree()
         expected_tree_id = tree.id.hex if isinstance(tree, pygit2.Tree) else tree
 
         if wc_tree_id != expected_tree_id:
@@ -499,17 +506,12 @@ class WorkingCopy:
             feature_diff = DeltaDiff()
             insert_count = delete_count = 0
 
-            geom_col = dataset.geom_column_name
-
             for row in r:
                 track_pk = row[0]  # This is always a str
                 db_obj = {k: row[k] for k in row.keys() if k != ".__track_pk"}
 
                 if db_obj[pk_field] is None:
                     db_obj = None
-
-                if db_obj is not None and geom_col is not None:
-                    db_obj[geom_col] = self._db_geom_to_gpkg_geom(db_obj[geom_col])
 
                 try:
                     repo_obj = dataset.get_feature(track_pk)
@@ -560,29 +562,32 @@ class WorkingCopy:
         else:
             schema = dataset.schema
 
-        table_identifier = self.table_identifier(dataset)
-        pk_column = self.quote(schema.pk_columns[0].name)
-        col_names = ",".join(
-            [f"{table_identifier}.{self.quote(col.name)}" for col in schema]
+        sno_track = self.sno_tables.sno_track
+        table = self._table_def_for_dataset(dataset, schema=schema)
+
+        cols_to_select = [sno_track.c.pk.label(".__track_pk"), *table.columns]
+        pk_column = table.columns[schema.pk_columns[0].name]
+        tracking_col_type = sno_track.c.pk.type
+
+        base_query = sqlalchemy.select(columns=cols_to_select).select_from(
+            sno_track.outerjoin(
+                table,
+                sno_track.c.pk == sqlalchemy.cast(pk_column, tracking_col_type),
+            )
         )
 
-        sql = f"""
-            SELECT
-                {self.SNO_TRACK}.pk AS ".__track_pk",
-                {col_names}
-            FROM {self.SNO_TRACK} LEFT OUTER JOIN {table_identifier}
-            ON ({self.SNO_TRACK}.pk = CAST({table_identifier}.{pk_column} AS TEXT))
-            WHERE ({self.SNO_TRACK}.table_name = :table_name)
-        """
-
         if feature_filter is UNFILTERED:
-            return sess.execute(sql, {"table_name": dataset.table_name})
+            query = base_query.where(sno_track.c.table_name == dataset.table_name)
         else:
             pks = list(feature_filter)
-            sql += f" AND {self.SNO_TRACK}.pk IN :pks"
-            t = sqlalchemy.text(sql)
-            t = t.bindparams(sqlalchemy.bindparam("pks", expanding=True))
-            return sess.execute(t, {"table_name": dataset.table_name, "pks": pks})
+            query = base_query.where(
+                sqlalchemy.and_(
+                    sno_track.c.table_name == dataset.table_name,
+                    sno_track.c.pk.in_(pks),
+                )
+            )
+
+        return sess.execute(query)
 
     def reset_tracking_table(self, reset_filter=UNFILTERED):
         """Delete the rows from the tracking table that match the given filter."""
@@ -611,10 +616,6 @@ class WorkingCopy:
                 )
                 t = t.bindparams(sqlalchemy.bindparam("pks", expanding=True))
                 sess.execute(t, {"table_name": table_name, "pks": pks})
-
-    def _db_geom_to_gpkg_geom(self, g):
-        """Convert a geometry as returned by the database to a sno geometry.Geometry object."""
-        raise NotImplementedError()
 
     def can_find_renames(self, meta_diff):
         """Can we find a renamed (aka moved) feature? There's no point looking for renames if the schema has changed."""
@@ -665,10 +666,10 @@ class WorkingCopy:
         tree_id = tree.id.hex if isinstance(tree, pygit2.Tree) else tree
         L.info(f"Tree sha: {tree_id}")
         with self.session() as sess:
-            changes = self._update_state_table_tree_impl(sess, tree_id)
+            changes = self._insert_or_replace_state_table_tree(sess, tree_id)
         assert changes == 1, f"{self.SNO_STATE} update: expected 1Δ, got {changes}"
 
-    def _update_state_table_tree_impl(self, sess, tree_id):
+    def _insert_or_replace_state_table_tree(self, sess, tree_id):
         """
         Write the given tree ID to the state table.
 
@@ -676,7 +677,10 @@ class WorkingCopy:
         tree_id - str, the hex SHA of the tree at HEAD.
         """
         r = sess.execute(
-            f"UPDATE {self.SNO_STATE} SET value=:value WHERE table_name='*' AND key='tree';",
+            f"""
+            INSERT INTO {self.SNO_STATE} ("table_name", "key", "value") VALUES ('*', 'tree', :value)
+            ON CONFLICT ("table_name", "key") DO UPDATE SET value=EXCLUDED.value;
+            """,
             {"value": tree_id},
         )
         return r.rowcount
@@ -701,7 +705,7 @@ class WorkingCopy:
                     self._create_spatial_index(sess, dataset)
 
                 L.info("Creating features...")
-                sql = self.insert_into_dataset(dataset)
+                sql = self._insert_into_dataset(dataset)
                 feat_progress = 0
                 t0 = time.monotonic()
                 t0p = t0
@@ -738,12 +742,8 @@ class WorkingCopy:
                 self._create_triggers(sess, dataset)
                 self._update_last_write_time(sess, dataset, commit)
 
-            sess.execute(
-                f"""
-                INSERT INTO {self.SNO_STATE} (table_name, key, value) VALUES ('*', 'tree', :value)
-                ON CONFLICT (table_name, key) DO UPDATE SET value=EXCLUDED.value;
-                """,
-                {"value": commit.peel(pygit2.Tree).hex},
+            self._insert_or_replace_state_table_tree(
+                sess, commit.peel(pygit2.Tree).id.hex
             )
 
     def _create_table_for_dataset(self, sess, dataset):
@@ -776,15 +776,15 @@ class WorkingCopy:
         if not pk_list:
             return 0
 
-        sql = self.insert_or_replace_into_dataset(dataset)
+        sql = self._insert_or_replace_into_dataset(dataset)
         feat_count = 0
         CHUNK_SIZE = 10000
         for row_dicts in self._chunk(
             dataset.get_features_with_crs_ids(pk_list, ignore_missing=ignore_missing),
             CHUNK_SIZE,
         ):
-            r = sess.execute(sql, row_dicts)
-            feat_count += r.rowcount
+            sess.execute(sql, row_dicts)
+            feat_count += len(row_dicts)
 
         return feat_count
 
@@ -953,7 +953,7 @@ class WorkingCopy:
 
             if not track_changes_as_dirty:
                 # update the tree id
-                self._update_state_table_tree_impl(sess, target_tree_id)
+                self._insert_or_replace_state_table_tree(sess, target_tree_id)
 
     def _filter_by_paths(self, datasets, paths):
         """Filters the datasets so that only those matching the paths are returned."""
--- a/sno/working_copy/base.py
+++ b/sno/working_copy/base.py
@@ -1,14 +1,13 @@
+from enum import Enum, auto
 import contextlib
 import functools
 import itertools
 import logging
 import time
-from enum import Enum, auto
-
 
 import click
 import pygit2
-import sqlalchemy
+import sqlalchemy as sa
 
 from sno.base_dataset import BaseDataset
 from sno.diff_structs import RepoDiff, DatasetDiff, DeltaDiff, Delta
@@ -48,7 +47,8 @@ class WorkingCopyType(Enum):
                 f"Unrecognised working copy type: {path}\n"
                 "Try one of:\n"
                 "  PATH.gpkg\n"
-                "  postgresql://[HOST]/DBNAME/SCHEMA\n"
+                "  postgresql://[HOST]/DBNAME/DBSCHEMA\n"
+                "  mssql://[HOST]/DBNAME/DBSCHEMA"
             )
 
     @property
@@ -103,12 +103,9 @@ class WorkingCopy:
     SNO_WORKINGCOPY_PATH = "sno.workingcopy.path"
 
     @property
-    @functools.lru_cache(maxsize=1)
-    def DB_SCHEMA(self):
-        """Escaped, dialect-specific name of the database-schema owned by this working copy (if any)."""
-        if self.db_schema is None:
-            raise RuntimeError("No schema to escape.")
-        return self.preparer.format_schema(self.db_schema)
+    def WORKING_COPY_TYPE_NAME(self):
+        """Human readable name of this type of working copy, eg "PostGIS"."""
+        raise NotImplementedError()
 
     @property
     @functools.lru_cache(maxsize=1)
@@ -147,9 +144,7 @@ class WorkingCopy:
             else dataset_or_table
         )
         sqlalchemy_table = (
-            sqlalchemy.table(table, schema=self.db_schema)
-            if isinstance(table, str)
-            else table
+            sa.table(table, schema=self.db_schema) if isinstance(table, str) else table
         )
         return self.preparer.format_table(sqlalchemy_table)
 
@@ -157,7 +152,7 @@ class WorkingCopy:
         # This is just used for selects/inserts/updates - we don't need the full type information.
         # We only need type information if some automatic conversion needs to happen on read or write.
         # TODO: return the full type information so we can also use it for CREATE TABLE.
-        return sqlalchemy.column(col.name)
+        return sa.column(col.name)
 
     @functools.lru_cache()
     def _table_def_for_dataset(self, dataset, schema=None):
@@ -166,7 +161,7 @@ class WorkingCopy:
         Specifies table and column names only - the minimum required such that an insert() or update() will work.
         """
         schema = schema or dataset.schema
-        return sqlalchemy.table(
+        return sa.table(
             dataset.table_name,
             *[self._table_def_for_column_schema(c, dataset) for c in schema],
             schema=self.db_schema,
@@ -261,7 +256,16 @@ class WorkingCopy:
             repo_cfg[path_key] = str(path)
 
     @classmethod
-    def check_valid_creation_path(cls, workdir_path, wc_path):
+    def subclass_from_path(cls, wc_path):
+        wct = WorkingCopyType.from_path(wc_path)
+        if wct.class_ is cls:
+            raise RuntimeError(
+                f"No subclass found - don't call subclass_from_path on concrete implementation {cls}."
+            )
+        return wct.class_
+
+    @classmethod
+    def check_valid_creation_path(cls, wc_path, workdir_path=None):
         """
         Given a user-supplied string describing where to put the working copy, ensures it is a valid location,
         and nothing already exists there that prevents us from creating it. Raises InvalidOperation if it is not.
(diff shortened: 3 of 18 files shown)


olsen232:
@craigds @rcoup This working copy is mostly done now

tests/conftest.py
```diff
@@ -912,3 +924,65 @@ def ctx(create=False):
...
+        docker run -it --rm -d -p 11433:1433 -e ACCEPT_EULA=Y -e 'SA_PASSWORD=Sql(server)' mcr.microsoft.com/mssql/server
+        SNO_SQLSERVER_URL='mssql://sa:Sql(server)@127.0.0.1:11433/master' spytest -k sqlserver --pdb -vvs
+    """
+    if "SNO_SQLSERVER_URL" not in os.environ:
+        raise pytest.skip(
+            "Requires SQL Server - read docstring at conftest.sqlserver_db"
+        )
+    engine = sqlserver_engine(os.environ["SNO_SQLSERVER_URL"])
+    with engine.connect() as conn:
+        # test connection and postgis support
```
craigds, line 942:
```suggestion
        # test connection
```

craigds (changes requested):
😅 phew

nice work! I had some questions and suggestions but on the whole it looks great 😄

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+from . import sqlserver_adapter
+from .table_defs import SqlServerSnoTables
+from sno import crs_util
+from sno.geometry import Geometry
+from sno.exceptions import InvalidOperation
+from sno.sqlalchemy import sqlserver_engine
+
+
+"""
+* database needs to exist
```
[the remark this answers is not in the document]
olsen232: Moved

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+* database needs to have postgis enabled
+* database user needs to be able to:
+    1. create 'sno' schema & tables
+    2. create & alter tables in the default (or specified) schema
+    3. create triggers
+"""
+
+L = logging.getLogger("sno.working_copy.postgis")
```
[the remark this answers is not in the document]
olsen232: Removed. Also removed equivalent in postgis.py

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+
+        url = urlsplit(uri)
+
+        if url.scheme != "mssql":
+            raise ValueError("Expecting mssql://")
+
+        url_path = url.path
+        path_parts = url_path[1:].split("/", 3) if url_path else []
+        if len(path_parts) != 2:
+            raise ValueError("Expecting mssql://[HOST]/DBNAME/SCHEMA")
```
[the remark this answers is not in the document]
olsen232: URL standard:
This URL isn't any kind of formal standard - it seems to be a sqlalchemy thing - some DB's have sensible URLs, like postgres://, some don't, like SQL server, but sqlalchemy accepts sensible looking URLs even for SQL server and converts them to whatever crazy thing SQL server expects.

Us adding the schema onto the end is also not a standard, but I like it better than a query-string - logically, the next path segment after the database is the database-schema, and so on if there are smaller subcontainers inside that. Just my preference.

Requiring a schema:
I'll keep this as-is for this PR - I can change it in a separate PR if we need to, both for this and for PostGIS. I think I convinced Rob last year that we should take ownership of a schema, although the schema-parsing code is originally his. Giving Sno complete ownership of a db schema makes good sense conceptually and avoids some difficulties:

- git owns a worktree folder, sno owns a GPKG - or a db schema. These are well-defined containers that are wholly owned by a single repo, everything inside them is managed.
- It's clear which tables are owned by which sno repo, instead of them being in a big pile with every other table
- Sno needs to create and delete tables - this avoids collisions with user tables of the same name, etc.
- although its not yet supported, we might one day allow tables to be created by the user inside the working copy and committed directly, instead of being created elsewhere and then imported. Having a clear boundary between the inside and outside of the working copy will make this possible.
- Avoids any mixups where tables end up being managed by two repos at once

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+                # TODO - use tidier syntax for opening transactions from sqlalchemy.
+                yield self._session
+                self._session.commit()
+            except Exception:
+                self._session.rollback()
+                raise
+            finally:
+                self._session.close()
+                del self._session
+                L.debug("session: new/done")
```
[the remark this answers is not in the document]
rcoup: Can we get rid of this altogether yet? Seems like [SAs session/savepoint handling](https://docs.sqlalchemy.org/en/13/orm/session_transaction.html) achieves the same thing in a more cohesive way?

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+        sess.execute(
+            f"""CREATE TABLE {self.table_identifier(dataset)} ({table_spec});"""
+        )
+
+    def _table_def_for_column_schema(self, col, dataset):
+        if col.data_type == "geometry":
+            # This user-defined Geography type adapts WKB to SQL Server's native geography type.
+            crs_name = col.extra_type_info.get("geometryCRS", None)
+            crs_id = crs_util.get_identifier_int_from_dataset(dataset, crs_name) or 0
+            return sqlalchemy.column(col.name, GeographyType(crs_id))
```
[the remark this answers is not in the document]
rcoup: I expect we'll need to support Geography types in MSSQL/PG at some point, but I think we just stick to Geometry everywhere at the moment.

>  two quite different types

Fundamentally, geometry is Euclidian and vertices are connected via lines. Geography is ellipsoidal, and vertices are connected via great elliptical arcs.

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+                USING
+                    (SELECT '{escaped_table_name}', {self.quote(pk_name)} FROM inserted
+                    UNION SELECT '{escaped_table_name}', {self.quote(pk_name)} FROM deleted)
+                    AS SRC (table_name, pk)
+                ON SRC.table_name = TRA.table_name AND SRC.pk = TRA.pk
+                WHEN NOT MATCHED THEN INSERT (table_name, pk) VALUES (SRC.table_name, SRC.pk);
+            END;
+            """,
+            {"table_name": dataset.table_name},
+        )
```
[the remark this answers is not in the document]
rcoup: Can we use the SA statement builder to safely assemble the content of the trigger, then serialise it as a string into the outer SA statement?

sno/working_copy/gpkg.py
```diff
@@ -75,9 +76,28 @@ def _quoted_trigger_name(self, dataset, trigger_type):
...
 
-    def insert_or_replace_into_dataset(self, dataset):
+    def _insert_or_replace_into_dataset(self, dataset):
         # SQLite optimisation.
-        return self.table_def_for_dataset(dataset).insert().prefix_with("OR REPLACE")
+        return self._table_def_for_dataset(dataset).insert().prefix_with("OR REPLACE")
+
+    def _table_def_for_column_schema(self, col, dataset):
+        if col.data_type == "geometry":
+            # This user-defined Geography type adapts WKB to SQL Server's native geography type.
```
rcoup, line 85:
```suggestion
            # This user-defined geometry type adapts columns to our V2 (GPKG-style) format.
```

sno/working_copy/gpkg.py
```diff
@@ -646,3 +666,21 @@ def _update_gpkg_contents(self, sess, dataset, commit=None):
                 {"last_change": gpkg_change_time, "table_name": dataset.table_name},
             ).rowcount
         assert rc == 1, f"gpkg_contents update: expected 1Δ, got {rc}"
+
+
+class GeometryType(UserDefinedType):
+    """UserDefinedType so that GPKG geometry is normalised to V2 format."""
+
+    def result_processor(self, dialect, coltype):
+        def process(gpkg_bytes):
```
rcoup, line 675: is this called lazily on column access? Could it be? Parsing and validating geometries is expensive.

sno/working_copy/postgis.py
```diff
@@ -53,14 +54,8 @@ def __init__(self, repo, uri):
         url_path = f"/{path_parts[0]}"
         self.db_schema = path_parts[1]
 
-        url_query = url.query
-        if "fallback_application_name" not in url_query:
-            url_query = "&".join(
-                filter(None, [url_query, "fallback_application_name=sno"])
-            )
```
rcoup, line 60: we still want this... though should use `urlencode()`

sno/working_copy/postgis.py
```diff
@@ -287,6 +282,12 @@ def _create_table_for_dataset(self, sess, dataset):
             f"""CREATE TABLE IF NOT EXISTS {self.table_identifier(dataset)} ({table_spec});"""
         )
 
+    def _insert_or_replace_into_dataset(self, dataset):
```
rcoup, line 285: this could do with a link to the relevant docs.

sno/working_copy/postgis.py
```diff
@@ -333,9 +334,10 @@ def _create_spatial_index(self, sess, dataset):
         # Create the PostGIS Spatial Index
         L.debug("Creating spatial index for %s.%s", dataset.table_name, geom_col)
         t0 = time.monotonic()
+        index_name = f"{dataset.table_name}_idx_{geom_col}"
         sess.execute(
             f"""
-            CREATE INDEX "{dataset.table_name}_idx_{geom_col}"
+            CREATE INDEX {self.quote(index_name)}
```
rcoup, line 340: Can we do this using [`Index()`](https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#index-types)?

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+    "int": ("integer", 32),
+    "bigint": ("integer", 64),
+    "real": ("float", 32),
+    "float": ("float", 64),
+    "date": "date",
+    "datetime": "timestamp",
+    "datetime2": "timestamp",
+    "datetimeoffset": "timestamp",
+    "decimal": "numeric",
+    "geography": "geometry",
```
rcoup, line 48: I think we need to explicitly not support geography in the meantime.

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+
+    pass
+
+
+@compiles(InstanceFunction)
+def compile_instance_function(element, compiler, **kw):
+    return "(%s).%s()" % (element.clauses, element.name)
+
+
+class GeographyType(UserDefinedType):
```
rcoup, line 458:
```suggestion
class GeometryType(UserDefinedType):
```

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+            type_=self,
+        )
+
+    def column_expression(self, col):
+        # 3. Reading - SQL layer - append with call to .STAsBinary() to convert MS binary to WKB.
+        return InstanceFunction("STAsBinary", col, type_=self)
+
+    def result_processor(self, dialect, coltype):
+        # 4. Reading - Python layer - convert WKB to sno geometry.
+        return lambda wkb: Geometry.from_wkb(wkb)
```
rcoup, line 483: This needs to be lazy too.

rcoup: Does MSSQL have a concept of a per-geometry CRS/SRID like PostGIS does? If so, should `Geometry` be initialised with the feature or WC SRID?

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+
+
+class BaseDateOrTimeType(UserDefinedType):
```
rcoup, line 486: Neither ODBC/MSSQL/SpatialAlchemy have any handling for dates? Why do we want them as strings?

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+        """
+        Returns True if the given meta-diff is supported *without* dropping and rewriting the table.
+        (Any meta change is supported if we drop and rewrite the table, but of course it is less efficient).
+        meta_diff - DeltaDiff object containing the meta changes.
+        """
+        # For now, just always drop and rewrite.
+        return not meta_diff
+
+
+class InstanceFunction(Function):
```
rcoup, line 442: This is generic to MSSQL but not our working copy? Would it be better homed in `sqlserver_adapter.py`?

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+        return Function(
+            "CONVERT",
+            literal_column("NVARCHAR"),
+            col,
+            literal_column("127"),
+            type_=self,
+        )
+
+
+def sqlserver_upsert(*args, **kwargs):
```
rcoup, line 500: This is generic to MSSQL but not our working copy? Would it be better homed in `sqlserver_adapter.py`?

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+def quote(ident):
+    return _PREPARER.quote(ident)
+
+
+V2_TYPE_TO_MS_TYPE = {
+    "boolean": "bit",
+    "blob": "varbinary",
+    "date": "date",
+    "float": {0: "real", 32: "real", 64: "float"},
+    "geometry": "geography",
```
rcoup, line 20:
```suggestion
    "geometry": "geometry",
```

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+    "numeric": "numeric",
+    "nvarchar": "text",
+    "text": "text",
+    "time": "time",
+    "varchar": "text",
+    "varbinary": "blob",
+}
+
+# Types that can't be roundtripped perfectly in SQL Server, and what they end up as.
+APPROXIMATED_TYPES = {"interval": "text"}
```
rcoup, line 60: Seems to support interval types? https://docs.microsoft.com/en-us/sql/odbc/reference/appendixes/interval-data-types?view=sql-server-ver15

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+                ON (KCU.table_schema = C.table_schema)
+                AND (KCU.table_name = C.table_name)
+                AND (KCU.column_name = C.column_name)
+                LEFT OUTER JOIN pg_attribute A
+                ON (A.attname = C.column_name)
+                AND (A.attrelid = (C.table_schema || '.' || C.table_name)::regclass::oid)
+                WHERE C.table_schema=:table_schema AND C.table_name=:table_name
+                ORDER BY C.ordinal_position;
+            """
+            table_info_sql = """
```
rcoup, line 373: redefining the above variable...

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+
+from sqlalchemy.sql.compiler import IdentifierPreparer
+from sqlalchemy.dialects.mssql.base import MSDialect
+
+
+_PREPARER = IdentifierPreparer(MSDialect())
+
+
+def quote(ident):
+    return _PREPARER.quote(ident)
```
rcoup, line 12: We need to drop these for SQLAlchemy's rather than writing more :)

rcoup (changes requested):
Wow, is impressive how little code is getting changed here 🌟

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+        self.dburl = urlunsplit([url.scheme, url.netloc, url_path, url.query, ""])
+        self.engine = sqlserver_engine(self.dburl)
+        self.sessionmaker = sessionmaker(bind=self.engine)
+        self.preparer = IdentifierPreparer(self.engine.dialect)
+
+        self.sno_tables = SqlServerSnoTables(self.db_schema)
+
+    @classmethod
+    def check_valid_creation_path(cls, workdir_path, path):
+        # TODO - promote to superclass
```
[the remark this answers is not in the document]
olsen232: Done - added a superclass with code common to SQL Server and PostGIS - 
DatabaseServer_WorkingCopy.
All the DB URI code is moved to there since its common to both.

sno/working_copy/sqlserver.py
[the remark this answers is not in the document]
olsen232: Tidied this up a little bit and moved it into the base class (except the GPKG implementation, which could also use some love).

I tried make further changes in a hopeful way, but it just caused some tests to fail.
Won't change in this PR - changing transaction management isn't straight forward since the nuances are invisible from the code and only show up as a few failing tests. If a few test fails showing that, eg, GPKG locking or transaction rollbacks have stopped working as expected, it's not clear what I should do except put the code back how it was - will require lots of experimentation.

Open questions:
- is `self.session.begin()` the same as `self.session.execute("BEGIN TRANSACTION")` (Answer: no, but not sure of details)
- will the context-manager `with self.session.begin()` automatically rollback transactions if an exception is raised?
- what's the difference between `with self.session.begin()` and `with self.session.transaction`.

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _sqlserver_to_column_schema(ms_col_info, id_salt):
+    """
+    Given the postgis column info for a particular column, and some extra context in
+    case it is a geometry column, converts it to a ColumnSchema. The extra context will
+    only be used if the given ms_col_info is the geometry column.
+    Parameters:
+    ms_col_info - info about a single column from ms_table_info.
+    ms_spatial_ref_sys - rows of the "spatial_ref_sys" table that are referenced by this dataset.
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+def sqlserver_to_v2_schema(ms_table_info, id_salt):
+    """Generate a V2 schema from the given postgis metadata tables."""
+    return Schema([_sqlserver_to_column_schema(col, id_salt) for col in ms_table_info])
+
+
+def _sqlserver_to_column_schema(ms_col_info, id_salt):
+    """
+    Given the postgis column info for a particular column, and some extra context in
+    case it is a geometry column, converts it to a ColumnSchema. The extra context will
+    only be used if the given ms_col_info is the geometry column.
```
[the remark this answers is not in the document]
olsen232: Fixed

sno/working_copy/table_defs.py
```diff
@@ -100,6 +102,36 @@ def create_all(self, session):
         return self._SQLALCHEMY_METADATA.create_all(session.connection())
 
 
+class SqlServerSnoTables(TableSet):
+    """
+    Tables for sno-specific metadata - PostGIS variant.
```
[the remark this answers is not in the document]
olsen232: Done

tests/conftest.py
```diff
@@ -912,3 +924,65 @@ def ctx(create=False):
...
 
     return ctx
+
+
+@pytest.fixture()
+def sqlserver_db():
+    """
+    Using docker, you can run a SQL Server test - such as those in test_working_copy_sqlserver - as follows:
+        docker run -it --rm -d -p 11433:1433 -e ACCEPT_EULA=Y -e 'SA_PASSWORD=Sql(server)' mcr.microsoft.com/mssql/server
+        SNO_SQLSERVER_URL='mssql://sa:Sql(server)@127.0.0.1:11433/master' spytest -k sqlserver --pdb -vvs
```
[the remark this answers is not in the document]
olsen232: Done

tests/conftest.py
olsen232: Done

sno/geometry.py
```diff
@@ -57,6 +57,11 @@ def with_crs_id(self, crs_id):
         crs_id_bytes = struct.pack("<i", crs_id)
         return Geometry.of(self[:4] + crs_id_bytes + self[8:])
 
+    @property
+    def crs_id(self):
```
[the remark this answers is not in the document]
olsen232: Done

sno/geometry.py
```diff
@@ -290,12 +295,18 @@ def gpkg_geom_to_ogr(gpkg_geom, parse_crs=False):
     return geom
 
 
-def wkt_to_gpkg_geom(wkb, **kwargs):
-    ogr_geom = ogr.CreateGeometryFromWkt(wkb)
+def wkt_to_gpkg_geom(wkt, **kwargs):
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/base.py
```diff
@@ -149,32 +153,36 @@ def table_identifier(self, dataset_or_table):
...
 
+    def _table_def_for_column_schema(self, col, dataset):
+        # This is just used for selects/inserts/updates - we don't need the full type information.
+        # We only need type information if some automatic conversion needs to happen on read or write.
+        # TODO: return the full type information so we can also use it for CREATE TABLE.
+        return sqlalchemy.column(col.name)
+
     @functools.lru_cache()
-    def table_def_for_dataset(self, dataset):
+    def _table_def_for_dataset(self, dataset, schema=None):
```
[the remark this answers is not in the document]
olsen232: Happily they are both actually immutable -
- a Schema is immutable because I made it immutable
- a Dataset is immutable because you made it immutable! and because it is kind of a view of a particular git tree, which is immutable. Like a git tree, it can't be edited directly - instead you just have to use information from it to help build a new commit.

sno/working_copy/sqlserver.py
olsen232: I'm open to moving everything around (maybe not in this PR).

Right now sqlserver_adapter (and other adapter classes) deal specifically with how the Sno model maps onto the information we can store inside the working copy of that type. And then sqlserver.py has basically everything else needed for this WC. So, current location makes sense from that logic, but open to other ways of dividing things up.

I might move it to an upsert file that specifies how to do upserts for all 3 dialects - that will get it out of the way.

sno/working_copy/sqlserver_adapter.py
olsen232: Not done - not sure if we should be using the SQL server geometry type, geography type, or if it should be configurable. (Same with VARCHAR vs NVARCHAR). Have you looked into this?

I tried switching geography for geometry right now and the tests failed since SQL server can't create a spatial index for a geometry type without me specifying the bounding box of the spatial index, apparently - I guess for a Geography type, Earth itself is used as the bounding box.

sno/working_copy/sqlserver_adapter.py
olsen232: Good catch - added "interval" -> "interval", removed this from approximated types.

sno/working_copy/sqlserver.py
olsen232: Oops. Removed first unneeded definition.

sno/working_copy/sqlserver_adapter.py
olsen232: This is sqlalchemy's quote. It's just easier than writing 
IdentifierPreparer(MSDialect()).quote(...) everytime.

However, you are right that I shouldn't be quoting as many things manually. The quote() in this file could go away if i generated a proper sqlalchemy Table definition for a dataset and create that, instead of generating appropriately quoted SQL. The amount of work is about the same - just haven't switched it yet. The other obstacle is it might make the code less symmetrical in terms of creating tables and reading table structures - not sure until I try it.

sno/working_copy/sqlserver_adapter.py
olsen232: Not done, for two reasons:
- from the linked documentation, intervals come in two flavours - year+month intervals, and day+hour+minute+second intervals. These two flavours can't be mixed in a single column, so using this type will probably lead to more problems than approximating as text.
- not sure, but I think the documentation is saying the Microsoft's ODBC driver supports intervals. SQL Server doesn't actually seem to have an interval type, or at least, not one I have found.
See https://docs.microsoft.com/en-us/sql/t-sql/data-types/data-types-transact-sql?view=sql-server-ver15

sno/geometry.py
```diff
@@ -290,12 +295,18 @@ def gpkg_geom_to_ogr(gpkg_geom, parse_crs=False):
...
+    if wkt is None:
+        return None
+
+    ogr_geom = ogr.CreateGeometryFromWkt(wkt)
     return ogr_to_gpkg_geom(ogr_geom, **kwargs)
 
 
 def wkb_to_gpkg_geom(wkb, **kwargs):
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver_adapter.py
olsen232: Not done. 

I'm not sure which SQL server type is appropriate - geometry or geography, or if it should be user-configurable. (Similarly: VARCHAR vs NVARCHAR).

I tried switching it to geometry, but it caused tests to fail since it could no longer create a spatial index - it says something about how creating spatial index for the geometry type column would require me to specify which bounding box the index should be created over. (I assume this is not required for geograpy types, since it just uses the whole world? Not sure.)

Anyway, I left it as is for now. Have you looked into the geometry and geography types?

sno/working_copy/sqlserver.py
olsen232: Oops! Removed useless first definition

sno/working_copy/sqlserver_adapter.py
olsen232: This is sqlalchemy's quote - I just wrapped it since writing `quote()` is easier than writing `IdentifierPreparer(MSDialect()).quote()`

But you are right - I should be trying to avoid manually quoting things. This quote here is only called when generating a table spec eg `("FID" bigint, "GEOM" geography, "CRT_DATE" datetime)` - it could go away if, instead of generating a table specification in SQL, I created one using sqlalchemy types and created that. It's about the same amount of work, but I just haven't switched to it yet. (It also might not end up being as symmetrical as the current code with regards to writing table specs and reading table specs - I don't know exactly unless I try to write it).

sno/sqlalchemy.py
```diff
@@ -95,7 +100,51 @@ def _on_connect(psycopg2_conn, connection_record):
...
+    return socket.gethostbyname("localhost")
+
+
+def _append_query_to_url(uri, query_dict):
+    url = urlsplit(uri)
+    url_query = _append_to_query(url.query, query_dict)
+    return urlunsplit([url.scheme, url.netloc, url.path, url_query, ""])
+
+
+def _append_to_query(url_query, query_dict):
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver.py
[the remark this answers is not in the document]
craigds: okay.

yeah, I don't understand sqlalchemy's approach to transaction management. There's huge warnings all over `session.begin()`. They seem to be saying "the session handles transactions for you! just don't even worry about it!" which is a recipe for disaster. The warnings stop _just_ short of saying that `session.begin()` will actually be deprecated or removed at some point, but it seems concerning

> will the context-manager with self.session.begin() automatically rollback transactions if an exception is raised?

pretty sure yes

sno/working_copy/sqlserver.py
rcoup: > Right now sqlserver_adapter (and other adapter classes) deal specifically with how the Sno model maps onto the information we can store inside the working copy of that type. 

Ah, that was my misunderstanding.

My though process was that:
- upsert/function/type-handling code is stuff that might appear in SQLAlchemy one day (or we could contribute)
- if we want native SQL _importers_ in future we'd likely need it, separately to use in our working copies.

I don't feel strongly about it though 😄

sno/working_copy/sqlserver_adapter.py
olsen232: Switched to geometry and added a heuristic for selecting the spatial index bounding box, as discussed offline.

sno/working_copy/base.py
```diff
@@ -560,29 +562,32 @@ def _execute_dirty_rows_query(
...
-            return sess.execute(sql, {"table_name": dataset.table_name})
+            query = base_query.where(sno_track.c.table_name == dataset.table_name)
         else:
             pks = list(feature_filter)
-            sql += f" AND {self.SNO_TRACK}.pk IN :pks"
-            t = sqlalchemy.text(sql)
-            t = t.bindparams(sqlalchemy.bindparam("pks", expanding=True))
-            return sess.execute(t, {"table_name": dataset.table_name, "pks": pks})
+            query = base_query.where(
+                sqlalchemy.and_(
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/gpkg.py
olsen232: Done

sno/working_copy/gpkg.py
olsen232: I don't know - but, there is no real benefit to doing so at present:
- The only reason we need to see what is actually in the user tables is to generate a diff
- When generating git-tree <> working-copy diffs, we need to check the entirety of each row (that is being tracked), to see what the diffs are
- We don't generate any of the diff lazily - we generate the whole diff before we output any of it. (For git-tree <> working-copy diffs). This could be rewritten so that it does work lazily, but with some difficulty - for instance, hard to see how we would keep rename-detection. The lazy-diff-output work I did before was entirely for git-tree <> git-tree diffs.

So, whether or not we make sure this is lazily-evaluated, the next step is for it to end up in basically a dict of all diffs that are then searched for renames. So not worth fixing except as part of a broader change to optimise git-tree <> working copy diffs.

sno/working_copy/postgis.py
olsen232: Done

sno/working_copy/postgis.py
olsen232: Done

sno/working_copy/sqlserver.py
olsen232: SQL server does have a concept of per-geometry CRS - in fact that is the only concept it has, unlike PostGIS, there is no column- or table-level way to specify the CRS.

In Sno on the other hand we *only* support column-level CRS (roughly equivalent to table-level, since there is normally only one geometry column). We specifically zero-out the CRS's in the individual GPKG geometries on commit so that you can't have different ones per geometry, and we treat them all as having the column-level one.

So, when reading geometries from the working copy, we always remove any embedded CRS ID - regardless of working copy type. Geometries can then be diffed against the stored Sno geometries, independently of CRS, and the CRS diff (if there is one) will show up only in the meta diff - since, Sno the only CRS change that Sno lets you diff and commit is the one where you change the entire column.

This is a piece of missing functionality in SQL Server right now - there is no way to change the CRS of a column except using `sno meta set`, since there is no other easy way for the user to signal to us that they want to change a columns CRS. Perhaps we should do a heuristic where we (for instance) sample a few geometries and see if they all have the same CRS - if so, that is probably the user-selected column CRS, and we can generate a diff.

Apologies, this is all undocumented since I haven't written SQL_SERVER.md or whatever it will be called

sno/working_copy/sqlserver.py
olsen232: Dates are put into the database as dates - that's fine, we declare the column type as "date" or similar and then any string we try to insert in there turns into a date - but we read them out as strings, using this UserDefinedType. This is because in Datasets V2, we store dates, times, etc as strings, so that's what we need to read when we want to do a diff / commit. The strings we store in Datasets V2 are ISO8601 formatted, which happily defines a format for dates, times, datetimes, and intervals.

This conversion could have happened at a different layer - eg we could have read dates from the database as Python datetimes, and then only convert them at a different Datasets V2 layer - but it doesn't, for a few reasons. 

- One is that Datasets V2 is our lingua franca, and I didn't see the point in defining another format for the layer between the database and datasets V2 - so things are read out of the database into V2 format immediately. This means they go straight into a format that can be easily serialised in message pack (strings are good) and skip more pythonic representations (datetimes).

- Another is that different working copies support different things and so doing the conversion immediately at the database layer straight back into V2 makes sense, since at that point you know the entire conversion you need to do. I could have decided to read PostGIS intervals as python timedeltas, and then converted them to something serialisable at another layer. But then SQL server doesn't have an interval type. So in that case, I would be reading a string from the database, converting it into a timedelta, then converting that back into a string, then serialising using message pack.

- Finally, having everything be immediately converted into our message-pack serializable V2 format, also means everything is JSON serialisable, and the diffs can easily be output to the user as text. It would be bad if we were comparing timedeltas and we decided they were different, but when we tried to show them to the user, the textual representation of them somehow ended up the same.

Maybe that was too much info

sno/working_copy/sqlserver.py
olsen232: Addressed below (for some reason github collapsed your upper 12 comments, which is why I missed them and addressed the lower ones first).

Seems happy where it is for right now but happy to move more SQL helpers out of working-copies as and when we need them for importers or testing.

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+                UPDATE SET "value" = SRC."value"
+            WHEN NOT MATCHED THEN
+                INSERT ("table_name", "key", "value") VALUES (SRC."table_name", SRC."key", SRC."value");
+            """,
+            {"value": tree_id},
+        )
+        return r.rowcount
+
+    def _write_meta(self, sess, dataset):
+        """Write the title (as a comment) and the CRS. Other metadata is not stored in a PostGIS WC."""
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+        self._write_meta_title(sess, dataset)
+
+    def _write_meta_title(self, sess, dataset):
+        """Write the dataset title as a comment on the table."""
+        # TODO - probably need to use sp_addextendedproperty @name=N'MS_Description'
```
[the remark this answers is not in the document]
olsen232: Not right now - it just means there's currently no place where the dataset `title` is stored in SQL Server working copy, so it can't be changed except using `sno meta set`.

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+        if new_type == "geometry":
+            new_col_dict["geometryType"] = old_col_dict.get("geometryType")
+            new_col_dict["geometryCRS"] = old_col_dict.get("geometryCRS")
+
+        return new_type == old_type
+
+    def _remove_hidden_meta_diffs(self, dataset, ds_meta_items, wc_meta_items):
+        super()._remove_hidden_meta_diffs(dataset, ds_meta_items, wc_meta_items)
+
+        # Nowhere to put these in postgis WC
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+    def result_processor(self, dialect, coltype):
+        # 4. Reading - Python layer - convert WKB to sno geometry.
+        return lambda wkb: Geometry.from_wkb(wkb)
+
+
+class BaseDateOrTimeType(UserDefinedType):
+    """UserDefinedType so we read dates, times, and datetimes as text."""
+
+    def column_expression(self, col):
+        # When reading, convert dates and times to strings using style 127: ISO8601 with time zone Z.
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,553 @@
...
+
+class InstanceFunction(Function):
+    """
+    An instance function that compiles like this when applied to an element:
+    >>> element.function()
+    Unlike a normal sqlalchemy function which would compile as follows:
+    >>> function(element)
+    """
+
+    pass
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver.py
olsen232: More usefully, added a comment explaining that its only during reading that we read them as text

sno/working_copy/sqlserver.py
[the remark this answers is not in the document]
olsen232: The fstring interpolation was because the database driver won't necessarily accept placeholders everywhere that a string literal would be valid - not sure if this is because of nested statements, or because of triggers in this case. Either way, the eventual SQL that is sent to the database driver must have the string literal inlined, hence the use of f-strings.
(`table_name` being there and unused was an artifact of me trying it one way and then realising I had to do it another, of course).

I've managed to get the sqlalchemy to inline and escape of the string literal itself using some magic.
I think this is about as good as its going to get - sqlalchemy doesn't have built in functions for CREATE TRIGGER or MERGE USING, so its always going to be custom SQL. I can write compiler extensions that generate this custom SQL, but then it probably becomes more likely that I will mess up escaping etc, not less.

sno/working_copy/sqlserver.py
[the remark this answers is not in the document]
olsen232: Rewrote the comment to be better.

olsen232:
Missed all the hidden comments last time - didn't realise github likes to hide comments. Think I've addressed everything now.
@rcoup @craigds

sno/working_copy/sqlserver_adapter.py
rcoup: in general it would be more useful to represent intervals as a double in seconds than anything text-based.

sno/working_copy/gpkg.py
rcoup: yup, that makes sense.

sno/working_copy/sqlserver.py
rcoup: > unlike PostGIS, there is no column- or table-level way to specify the CRS.

PostGIS supports both... per-geometry CRS is normal, with a table-level constraint to check it. But without the constraint it'll let anything be stored.

> Perhaps we should do a heuristic where we (for instance) sample a few geometries and see if they all have the same CRS - if so, that is probably the user-selected column CRS, and we can generate a diff.

kinda-sometimes-working magic is worse IME than no magic. It's ok we only support homogenous CRS in a single dataset — that is by far the most normal scenario. And as long as we have some steps to follow to change the CRS then that's good.

sno/working_copy/sqlserver.py
rcoup: Makes sense 👍

sno/working_copy/sqlserver_adapter.py
rcoup: > (Similarly: VARCHAR vs NVARCHAR).

From a bit of reading, seems like `NVARCHAR` is normal for unicode pre- MSSQL 2019, and [from then `VARCHAR` explicitly supports UTF8](https://docs.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver15). So I think using `NVARCHAR`/`NCHAR`/`NTEXT` makes sense initially, and we can revisit in the future.

sno/working_copy/sqlserver_adapter.py
rcoup: One option is to use use `Table()`/etc, but you can also use the SA SQL expression builder which has escaping functions/templating/etc for SQL strings.

sno/sqlalchemy.py
```diff
@@ -95,7 +101,64 @@ def _on_connect(psycopg2_conn, connection_record):
...
+    url_query = _append_to_query(url.query, new_query_dict)
+    return urlunsplit([url.scheme, url.netloc, url.path, url_query, ""])
+
+
+def _append_to_query(existing_query, new_query_dict):
+    query_dict = parse_qs(existing_query)
+    for key, value in new_query_dict.items():
+        if key not in query_dict:
+            query_dict[key] = value
+    return urlencode(query_dict)
```
rcoup, line 164:
```suggestion
    # ignore new keys if they're already set in the querystring
    return urlencode({**new_query_dict, **query_dict})
```

sno/working_copy/sqlserver.py
```diff
@@ -0,0 +1,481 @@
...
+        geom_col = dataset.geom_column_name
+        index_name = f"{dataset.table_name}_idx_{geom_col}"
+
+        # Create the SQL Server Spatial Index
+        L.debug("Creating spatial index for %s.%s", dataset.table_name, geom_col)
+        t0 = time.monotonic()
+        # Create an index over the extent of existing geometries + 20% room to grow.
+        GROW_FACTOR = 1.2
+        min_x, min_y, max_x, max_y = self._grow_rectangle(
+            self._get_geom_extent(sess, dataset, default=(-180, -90, +180, +90)),
```
rcoup, line 249: Unless the CRS is geographic the default here won't be useful... would be better _not_ to create a spatial index for an empty table with a geometry column.

rcoup (commented):
Nearly there 😄

sno/working_copy/sqlserver.py
olsen232: Marking this as resolved

sno/working_copy/sqlserver_adapter.py
olsen232: Marking this as resolved

sno/working_copy/sqlserver_adapter.py
olsen232: This is not quite true of intervals - unfortunately there is no formula for converting a month into seconds. The interval type, in the working copies we have the support it (so far only PostGIS) can't be collapsed into any single number.

I also don't know if the user wants to do date-math, or to write conversion code - if I gave the user something machine-readable like number of seconds, then an advanced user could easily write a function that converted that into the datatype of their choice for display and manipulation. But all I know about this user is that they are checking out some data that they presumably didn't author, into a working copy type that the data is not native to, so perhaps they are outside the main use-case for this data. They might not have any extra tooling, or any developers to help them write it, but they might instead be glad of the ability to edit or at least view the intervals in a more-or-less human readable way.

sno/working_copy/sqlserver_adapter.py
olsen232: Yes - will revisit this soon to try and get rid of lots of f-strings in all 3 working copies.

sno/sqlalchemy.py
olsen232: Better code 👍

sno/working_copy/base.py
```diff
@@ -96,15 +100,10 @@ class WorkingCopy:
     self.sno_tables - sqlalchemy Table definitions for sno_track and sno_state tables.
     """
 
-    SNO_WORKINGCOPY_PATH = "sno.workingcopy.path"
+    WORKING_COPY_TYPE_NAME = "AbstractBaseClass"
```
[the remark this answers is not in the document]
olsen232: Yeah this doesn't actually need to be here except I like to put things in the superclass that subclasses should override as a sort of documentation, even though in python you don't need to.

I'll set it to [self.__class__].__name__

sno/working_copy/base.py
```diff
@@ -754,18 +774,29 @@ def _write_meta(self, sess, dataset):
...
+
+        # Note that the simplest implementation is to add a trigger here so that any further writes update
+        # the index. Then _create_spatial_index_post needn't be implemented.
+        pass
+
+    def _create_spatial_index_post(self, sess, dataset):
+        """Like _create_spatial_index_pre, but runs AFTER the bulk of features have been written."""
+
+        # Being able to create the index after the bulk of features have been written could be useful for two reasons:
+        # 1. It might be more efficient to write the features first, then index afterwards.
```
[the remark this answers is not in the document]
olsen232: Marking as resolved, although if you know of a way to do it in bulk afterwards for GPKG then let me know

sno/working_copy/postgis.py
```diff
@@ -1,36 +1,38 @@
...
+    """
+    PosttGIS working copy implementation.
 
-L = logging.getLogger("sno.working_copy.postgis")
+    Requirements:
+    1. Database needs to exist
+    2. PostGIS needs to be enabled (unless no geometry is used).
+    3. Database user needs to be able to:
+        - Create the specified schema (unless it already exists).
+        - Create, delete and alter tables and triggers in the speficied schema.
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/postgis.py
```diff
@@ -1,36 +1,38 @@
...
-    3. create triggers
-"""
+class WorkingCopy_Postgis(DatabaseServer_WorkingCopy):
+    """
+    PosttGIS working copy implementation.
 
-L = logging.getLogger("sno.working_copy.postgis")
+    Requirements:
+    1. Database needs to exist
+    2. PostGIS needs to be enabled (unless no geometry is used).
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+    "datetime2": "timestamp",
+    "datetimeoffset": "timestamp",
+    "decimal": "numeric",
+    "geography": "geometry",
+    "geometry": "geometry",
+    "numeric": "numeric",
+    "nvarchar": "text",
+    "text": "text",
+    "time": "time",
+    "varchar": "text",
```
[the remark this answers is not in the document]
olsen232: Done (these mapping won't be used unless the user starts messing with column types, but that is something we allow)

tests/conftest.py
```diff
@@ -912,3 +924,65 @@ def ctx(create=False):
...
 
     return ctx
+
+
+@pytest.fixture()
+def sqlserver_db():
+    """
+    Using docker, you can run a SQL Server test - such as those in test_working_copy_sqlserver - as follows:
+        docker run -it --rm -d -p 11433:1433 -e ACCEPT_EULA=Y -e 'SA_PASSWORD=Sql(server)' mcr.microsoft.com/mssql/server
+        SNO_SQLSERVER_URL='mssql://sa:Sql(server)@127.0.0.1:11433/master' pytest -k sqlserver --pdb -vvs
```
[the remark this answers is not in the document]
olsen232: looks strong to me 👍

tests/test_working_copy_sqlserver.py
```diff
@@ -0,0 +1,313 @@
...
+    with data_archive("types") as repo_path:
+        repo = SnoRepo(repo_path)
+        H.clear_working_copy()
+
+        with new_sqlserver_db_schema() as (sqlserver_url, sqlserver_schema):
+            repo.config["sno.workingcopy.path"] = sqlserver_url
+            r = cli_runner.invoke(["checkout"])
+
+            # If type-approximation roundtrip code isn't working,
+            # we would get spurious diffs on types that PostGIS doesn't support.
```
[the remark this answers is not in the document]
olsen232: Done

tests/test_working_copy_sqlserver.py
```diff
@@ -0,0 +1,313 @@
...
+            assert repo.head.peel(pygit2.Commit).hex == orig_head
+
+
+def test_approximated_types():
+    assert sqlserver_adapter.APPROXIMATED_TYPES == compute_approximated_types(
+        sqlserver_adapter.V2_TYPE_TO_MS_TYPE, sqlserver_adapter.MS_TYPE_TO_V2_TYPE
+    )
+
+
+def test_types_roundtrip(data_archive, cli_runner, new_sqlserver_db_schema):
```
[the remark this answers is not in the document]
olsen232: No, it just has a working copy with one of every sno datatype, including all different sized integers.
MSSQL is unique in that it's not able to roundtrip geometrytype or CRS (eg there's nowhere to store that a column is `POINT, 4326`) and so that diff has to be suppressed (eg, we take the "geometry" type from the working copy when generating a diff, but we simply copy the other info from the dataset, so that cannot differ and no change to it can be committed). If this wasn't working, any test that diffs geometry types would have a spurious meta diff, this one included.

sno/working_copy/sqlserver.py
olsen232: Okay, done

sno/working_copy/sqlserver_adapter.py
rcoup: >  there is no formula for converting a month into seconds

ah, yes. ick.

Go with what you have already. 👍

sno/working_copy/gpkg.py
```diff
@@ -603,6 +634,19 @@ def _apply_meta_metadata_dataset_json(self, sess, dataset, src_value, dest_value
     def _update_last_write_time(self, sess, dataset, commit=None):
         self._update_gpkg_contents(sess, dataset, commit)
 
+    def _get_geom_extent(self, sess, dataset):
+        """Returns the envelope around the entire dataset as (min_x, min_y, max_x, max_y)."""
```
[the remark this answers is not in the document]
olsen232: Not done.
Ugh just spent a good half hour on this  - for no particular reason, calling GetSpatialIndex from within Sno / sqlalchemy reliably segfaults for me, although I can call it from my `sqlite3` CLI. My spatialite version should be the same for both. Left this as-is.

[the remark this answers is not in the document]
rcoup: Fair enough. 

It's documented as returning a _Spatialite_ geometry, so would need to cast it to GPKG/WKB or extract min/max values from it. Not sure if that's the issue though. 🤷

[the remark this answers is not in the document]
olsen232: Presumably not the issue - I was calling ST_MinX etc on it, it was working in `sqlite3`, just not in Sno

sno/working_copy/postgis.py
```diff
@@ -325,20 +239,25 @@ def delete_meta(self, dataset):
...
-            CREATE INDEX "{dataset.table_name}_idx_{geom_col}"
-            ON {self.table_identifier(dataset)} USING GIST ({self.quote(geom_col)});
-            """
+
+        spatial_index = Index(
+            index_name, table.columns[geom_col], postgres_using="GIST"
         )
+        spatial_index.table = table
+        spatial_index.create(sess.connection())
+        sess.execute(f"""ANALYSE {self.table_identifier(dataset)};""")
```
rcoup, line 259:
```suggestion
        sess.execute(f"""ANALYZE {self.table_identifier(dataset)};""")
```

kinda implies it's not tested though?

olsen232: Switched to ANALYZE

ANALYSE is apparently a synonym for ANALYZE, switching to any other misspelling causes tests to fail

rcoup: aha, someone added a secret alias for proper English? 👍

rcoup (approved):
Awesome! 🏅

PR #364, opened 2021-03-01
Add support for `sno commit-files --amend`

olsen232:
## Description

`sno commit-files --amend` now amends the previous commit:
`sno commit-files --amend -m MESSAGE` changes its commit message
`sno commit-files --amend FILE=@CONTENTS` adds one extra FILE containing CONTENTS to the previous commit

### DIFF
--- a/sno/meta.py
+++ b/sno/meta.py
@@ -201,23 +201,29 @@ def meta_set(ctx, message, dataset, items):
     type=StringFromFile(encoding="utf-8"),
 )
 @click.option("--ref", default="HEAD")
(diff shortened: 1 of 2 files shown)


craigds (approved):

rcoup:
> `sno commit-files --amend FILE=@CONTENTS`

so `FILE` is the path in the repo, and `@CONTENTS` loads content from a file path at `CONTENTS`? Can we expand the docstring to explain this super-clearly? Metavar has `KEY=VALUE` which is different...

PR #370, opened 2021-03-08
Add constraints to geometry column in SQL server

olsen232:
## Description

Ensures that only geometries that match the Sno schema can be inserted into the database column, with constraints like the following:

CHECK (geom.STGeometryType() = 'POINT' AND geom.STSrid = 4326);

Although it looks like this would fail on NULLs (as it would in an OO-language) in SQL Server this works fine on NULLs, as shown in the unit test.

## Related links:

@rcoup requested this in the documentation PR:
https://github.com/koordinates/sno/pull/369

### DIFF
--- a/sno/working_copy/sqlserver_adapter.py
+++ b/sno/working_copy/sqlserver_adapter.py
@@ -1,3 +1,4 @@
+from sno import crs_util
 from sno.schema import Schema, ColumnSchema
 
 from sqlalchemy.sql.compiler import IdentifierPreparer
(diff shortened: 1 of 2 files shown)


sno/working_copy/sqlserver_adapter.py
```diff
@@ -65,12 +66,40 @@ def quote(ident):
 # geometry type info - the geometry type and CRS - is not roundtripped.
 
 
+# Used for constraining a column to be of a certain type, including subtypes of that type.
```
rcoup, line 69: So, PostGIS at least doesn't work this way. If I create a `GEOMETRYCOLLECTION` column that's the _only_ type I can insert into it. Which I think makes sense, because Simple Features for SQL doesn't have any "is-a" type functions beyond [`ST_GeometryType()`](https://postgis.net/docs/ST_GeometryType.html) so it's not easy to actually check whether you have a subtype.

But [GeoPackage explicitly allows assignable subtypes](https://www.geopackage.org/spec120/#r32). So I guess we go with that... (does that cause issues for PostGIS type constraints? Maybe we need some more tests around subtypes)

rcoup: And to be helpful Shapefiles mix MULTIX & X types freely in an "X" column, so datasets originating from there will potentially be a mix, or otherwise we cast to MULTIX. Thoughts @craigds

rcoup (changes requested):
The NULL test you mentioned is part of the layer fixture?

olsen232:
Forgot to push the latest - NULL test was missing

sno/working_copy/sqlserver_adapter.py
olsen232: Does that mean, if you create `GEOMETRY` column, then you can only put `GEOMETRY` types into it? That seems suspect...

rcoup: Geometry is an abstract type, so there's no check constraint and you can add anything into it.

sno/working_copy/sqlserver_adapter.py
```diff
@@ -65,12 +66,40 @@ def quote(ident):
...
+# The CHECK need to explicitly list all types and subtypes, eg for SURFACE:
+# >>> CHECK(geom.STGeometryType() IN ('SURFACE','POLYGON','CURVEPOLYGON'))
+_GEOMETRY_SUBTYPES = {
+    "GEOMETRY": set(["POINT", "CURVE", "SURFACE", "GEOMCOLLECTION"]),
+    "CURVE": set(["LINESTRING", "CIRCULARSTRING", "COMPOUNDCURVE"]),
+    "SURFACE": set(["POLYGON", "CURVEPOLYGON"]),
+    "GEOMCOLLECTION": set(["MULTIPOINT", "MULTICURVE", "MULTISURFACE"]),
```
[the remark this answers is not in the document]
olsen232: Did some experimentation - you are right, it is GEOMETRYCOLLECTION.

sno/working_copy/sqlserver_adapter.py
```diff
@@ -65,12 +66,40 @@ def quote(ident):
 # geometry type info - the geometry type and CRS - is not roundtripped.
 
 
+# Used for constraining a column to be of a certain type, including subtypes of that type.
+# The CHECK need to explicitly list all types and subtypes, eg for SURFACE:
+# >>> CHECK(geom.STGeometryType() IN ('SURFACE','POLYGON','CURVEPOLYGON'))
+_GEOMETRY_SUBTYPES = {
+    "GEOMETRY": set(["POINT", "CURVE", "SURFACE", "GEOMCOLLECTION"]),
```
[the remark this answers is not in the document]
olsen232: Done

sno/working_copy/sqlserver_adapter.py
rcoup: Yeah, I'm happy to go right now with:
- `GEOMETRY` → anything (no constraint)
- other types → them & their subtypes

Then revisit separately and figure out whether we tighten to the PG model or go with the GPKG approach in the general case.

sno/working_copy/sqlserver_adapter.py
```diff
@@ -79,7 +108,51 @@ def v2_schema_to_sqlserver_spec(schema, v2_obj):
...
+
+def _geometry_type_constraint(col_name, geometry_type):
+    result = f"({quote(col_name)}).STGeometryType()"
+
+    geometry_type = geometry_type.upper()
+    geometry_types = _GEOMETRY_SUBTYPES.get(geometry_type)
+
+    if geometry_types:
+        geometry_types = ",".join(f"'{g}'" for g in geometry_types)
+        result += f" IN {geometry_types}"
```
rcoup, line 144: [The docs](https://docs.microsoft.com/en-us/sql/t-sql/spatial-geometry/stgeometrytype-geometry-data-type?view=sql-server-ver15) and some examples on the web seem to have `StGeometryType()` returning `'LineString'` instead of `'LINESTRING'`. 

I guess if the tests pass then the internal comparison is case-insensitive? 🤷

rcoup (approved):
👍 :shipit:

sno/working_copy/sqlserver_adapter.py
olsen232: Ugh. I did wonder about this... tests pass... but it might depend on the collation settings of my SQL Server instance, which of course, shouldn't be relied upon.
Fixing this now.

PR #377, opened 2021-03-15
Make pyodbc work for windows and mac builds 

olsen232:
## Description

We don't yet have a working way of including pyodbc for linux builds (only for unix, if I understand correctly).
This PR makes pyodbc a dependency for windows and mac, but not for linux. Any tests that require pyodbc - that is, all the SQL Server tests - are marked as xfail, but only on linux.

Note that also - the only environment where these tests are even being attempted as part of the CI tests is also linux, as of this PR. That's because the linux environment is spinning up the SQL Server docker image, and the other environments don't. However, it's not being used yet, since these tests are xfailing, since we don't have pyodbc.

The tests can be run locally if you have a SQL Server URI to point them at. They are passing locally.

Stil TODO (in future PRs):
- include pyodbc on linux, once we have it working
- include SQL Server driver in windows, mac, linux builds
- remove xfail markers

### DIFF
--- a/tests/test_working_copy_sqlserver.py
+++ b/tests/test_working_copy_sqlserver.py
@@ -3,6 +3,7 @@ import pytest
 import pygit2
 from sqlalchemy.exc import IntegrityError
 
+from sno import is_linux
 from sno.repo import SnoRepo
 from sno.working_copy import sqlserver_adapter
 from test_working_copy import compute_approximated_types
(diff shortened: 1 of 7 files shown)


tests/test_working_copy_sqlserver.py
```diff
@@ -11,6 +12,7 @@
 H = pytest.helpers.helpers()
 
 
+@pytest.mark.xfail(is_linux, reason="pyodbc is not yet included in linux build")
```
rcoup, line 15: for future reference, you can actually do this [once at the module level](https://docs.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules):

```python
# xfail all tests in this module if we're on Linux
pytestmark = pytest.mark.xfail(is_linux, reason="pyodbc is not yet included in linux build"))
```

PR #378, opened 2021-03-16
import: Add --replace-ids flag

craigds:
## Description

The existing `--replace-existing` flag imports over the top of an
existing dataset.

Imports from huge source tables can be slow, since the entire table
must be read and imported into git, even if only a small subset of the
table changed.

If the caller knows what parts of the table have changed, it makes
sense for them to supply that information to sno, to speed up the
import.

This change adds a `--replace-ids` flag, which supplies a list of
primary key values which sno should look at when determining the
changeset to import.

## Related links:

none

## Checklist:

### DIFF
--- a/sno/init.py
+++ b/sno/init.py
@@ -57,6 +57,20 @@ def _add_datasets_to_working_copy(repo, *datasets, replace_existing=False):
     wc.write_full(commit, *datasets)
 
 
+class GenerateIDsFromFile(StringFromFile):
+    name = "ids"
+
+    def convert(self, value, param, ctx):
+        fp = super().convert(
+            value,
+            param,
+            ctx,
+            # Get the file object, so we don't have to read the whole thing
+            as_file=True,
+        )
+        return iter(fp)
+
+
 @click.command("import")
 @click.pass_context
 @click.argument("source")
@@ -135,6 +149,14 @@ def _add_datasets_to_working_copy(repo, *datasets, replace_existing=False):
     help="Replace existing dataset(s) of the same name.",
 )
 @click.option(
+    "--replace-ids",
+    type=GenerateIDsFromFile(encoding="utf-8"),
+    help=(
+        "Replace only features with the given IDs. IDs should be given one-per-line. "
+        "Use file arguments (--replace-ids=@filename.txt)"
+    ),
+)
+@click.option(
     "--similarity-detection-limit",
     hidden=True,
     type=click.INT,
@@ -180,6 +202,7 @@ def import_table(
     tables,
     table_info,
     replace_existing,
+    replace_ids,
     similarity_detection_limit,
     allow_empty,
     max_delta_depth,
@@ -245,6 +268,18 @@ def import_table(
             description=info.get("description"),
             xml_metadata=info.get("xmlMetadata"),
         )
+        if replace_ids is not None:
+            if repo.version < 2:
+                raise InvalidOperation(
+                    f"--replace-ids is not supported for V{repo.version} datasets"
+                )
+            if not import_source.schema.pk_columns:
+                raise InvalidOperation(
+                    "--replace-ids requires an import source with a primary key"
+                )
+
+            replace_existing = True
+
         if replace_existing:
             if repo.version < 2:
                 raise InvalidOperation(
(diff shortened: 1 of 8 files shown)


olsen232 (commented):
Sorry to be a pain

base_dataset already has a function `get_features` that seems to be basically the same as `features(filter_pks=not None)`

Maybe the naming of `features` for all features and `get_features(some_pks)` are not the best.
However, there is an upside to having two different functions, as opposed to adding an optional query param, that is shown by this PR - if you add an extra function, you can just make a "Not yet implemented" implementation in the base class, and leave all subclasses alone. If you add an extra optional query parameter, you have to change them all.

For this reason and to avoid having too many nearly-identical concepts, I recommend you revert the `filter_to_pks` part of this change and use `get_features` instead. Right now the get_features functionality is only part of the dataset interface, but you can add it to any import sources where you need as well. Also, add a specification for it to the ImportSource interface - by specification I mean something like this:

```
class ImportSource:
...
def get_features(self, row_pks, *, ignore_missing=False):
        """
        Yields a dict for each of the specified features.
        If ignore_missing is True, then failing to find a specified feature does not raise a KeyError.
        """
        raise NotImplementedError()
```

The other approach that doesn't lead to duplicated code / duplicated concepts, would be to get rid of `get_features` and only use `features` everywhere.

craigds:
thanks. sorted that - I implemented `get_features` on the import sources and it's a bit cleaner now

olsen232 (approved):

sno/init.py
```diff
@@ -245,6 +268,18 @@ def import_table(
             description=info.get("description"),
             xml_metadata=info.get("xmlMetadata"),
         )
+        if replace_ids is not None:
```
rcoup, line 271: seems like this is only valid if you specify `--replace-existing` too?

sno/init.py
```diff
@@ -57,6 +57,20 @@ def _add_datasets_to_working_copy(repo, *datasets, replace_existing=False):
...
+
+    def convert(self, value, param, ctx):
+        fp = super().convert(
+            value,
+            param,
+            ctx,
+            # Get the file object, so we don't have to read the whole thing
+            as_file=True,
+        )
+        return iter(fp)
```
rcoup, line 71: do we need to strip newlines off these? Or do any other validation?

sno/ogr_import_source.py
```diff
@@ -387,6 +389,27 @@ def features(self):
         for ogr_feature in self._iter_ogr_features():
             yield self._ogr_feature_to_sno_feature(ogr_feature)
 
+    def get_features(self, row_pks, *, ignore_missing=False):
+        pk_field = self.primary_key
+
+        # TODO: does OGR have any better way of quoting these parameters?
```
rcoup, line 395: Not AFAIK, follows strict SQL formatting and expects you to have escaped already: https://gdal.org/user/ogr_sql_dialect.html

sno/ogr_import_source.py
```diff
@@ -387,6 +389,27 @@ def features(self):
...
+        # there's no params argument to SetAttributeFilter()...
+        def _quote(x):
+            if isinstance(x, int):
+                return str(x)
+            else:
+                return "'{}'".format(str(x).replace("'", "''"))
+
+        # (123,) --> 123. we only handle one pk field
```
rcoup, line 403: assert? Or checked elsewhere?

tests/test_init.py
```diff
@@ -400,6 +400,85 @@ def test_import_replace_existing_with_column_renames(
...
+                    """
+                )
+                conn.execute(
+                    """
+                    UPDATE "nz_waca_adjustments" SET adjusted_nodes = adjusted_nodes + 5
+                    WHERE id IN (4408774, 4408145)
+                    """
+                )
+
+            r = cli_runner.invoke(
```
rcoup, line 438: comment on how the above IDs relate to what we're specifying below, and what we're checking?

rcoup (commented):
LGTM, there's a few tidyups + comments that need to happen.

sno/init.py
craigds: it implies it. i'll add to help text

sno/init.py
craigds: apparently yes we need to strip newlines, not sure how I missed that. will do

validation: there's not much I _can_ do easily; any valid string might be a valid PK

sno/ogr_import_source.py
craigds: currently there's just no way of getting a multi-PK dataset into sno in the first place, but I'll add an assert anyway

PR #379, opened 2021-03-16
Include unixodbc in mac build

olsen232:
## Description

Adds unixodbc to the mac build, delocates it appropriately.
unixodbc is required by pyodbc.

Still TODO:
- On mac: modify pyodbc library rpath to point to unixodbc
- On linux: build unixodbc, build or install pyodbc and point it to unixodbc

This was a long day's work

### DIFF
--- a/vendor/Brewfile
+++ b/vendor/Brewfile
@@ -15,3 +15,5 @@ brew "json-c"
 brew "curl"
 
 brew "sqlite3"
+
+brew "unixodbc"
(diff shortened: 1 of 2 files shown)


rcoup:
Looks good!

> On mac: modify pyodbc library rpath to point to unixodbc

ya, so I think we follow the psycopg2 pattern here and build a pyodbc wheel against your new unixodbc library.

PR #381, opened 2021-03-17
Build pyodbc on mac, depend on unixodbc

olsen232:
## Description

Fixes pyodbc on the mac build to depend on the unixodbc that is also part of the mac build, instead of depending on a system lib that may or may not be there.

One can test that it worked like so:
```
~/code/sno$ git checkout pyodbc
~/code/sno$ make sno
~/code/sno$ venv/bin/python
>>> import pyodbc
>>> pyodbc.__file__
~/code/sno/venv/lib/python3.7/site-packages/pyodbc.cpython-37m-darwin.so
>>> ^D

~/code/sno$ otool -L venv/lib/python3.7/site-packages/pyodbc.cpython-37m-darwin.so
venv/lib/python3.7/site-packages/pyodbc.cpython-37m-darwin.so:
	@rpath/libodbc.2.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 800.7.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.0.0)
	
~/code/sno$ otool -L venv/lib/libodbc.2.dylib
otool -L venv/lib/libodbc.2.dylib
venv/lib/libodbc.2.dylib:
	libodbc.2.dylib (compatibility version 3.0.0, current version 3.0.0)
	@loader_path/libltdl.7.dylib (compatibility version 11.0.0, current version 11.1.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,10 +24,12 @@ vendor/proj/**
 !vendor/proj/Makefile
 vendor/psycopg2/**
 !vendor/psycopg2/Makefile
-vendor/pysqlite3/**
-!vendor/pysqlite3/Makefile
 vendor/pygit2/**
 !vendor/pygit2/Makefile
+vendor/pyodbc/**
+!vendor/pyodbc/Makefile
+vendor/pysqlite3/**
+!vendor/pysqlite3/Makefile
 vendor/spatialindex/**
 !vendor/spatialindex/Makefile
 vendor/spatialite/**
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,8 +28,6 @@ pycparser==2.20
     # via -r requirements/requirements.in
 pygments==2.8.1
     # via -r requirements/requirements.in
-pyodbc==4.0.30 ; platform_system != "Linux"
-    # via -r requirements/requirements.in
 pyrsistent==0.17.3
     # via jsonschema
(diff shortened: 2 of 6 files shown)


rcoup:
* `otool -l | grep -A4 PATH` on the two libraries shows RPATH being set as expected?
* `libltdl.7.dylib` → this ends up in the vendor tgz too?

olsen232:
1. Good catch - it turns out, pyodbc isn't in a module folder, it's just a single .so file. So the .so file sits one level higher up compared to other py libraries, so, I was setting my rpath with one `../` too many. Fixed now.
1b) This means I can't declare the output of the pyodbc rule as being `pyodbc/__init__.py`, but instead, `pyodbc.cpython-37m-darwin.so` (and on linux, `pyodbc.cpython-37m-x86_64-linux-gnu.so`). Is this okay?

2. Yes, this ends up in vendor/env/lib

rcoup:
1b) yup 👍

PR #384, opened 2021-03-24
Rewrote the find-similar-feature algorithm in PK-less import to be faster

olsen232:
When git does exact-rename detection, it simply hashes all of the contents of all of the inserted or deleted files. Then it checks which hashes are the same. We do something similar when we detecting exact-feature-reimports during PK-less import. In our use-case, these features will *not* be renamed, they will be assigned the same PK. This algorithm is fast - linear on the number of features.

When git does similar-rename-detection - finding files that have been both renamed and edited - git has to compare the full contents of every deleted file to every inserted file. This is because the hashes of similar files are not themselves similar, so git has no choice but to compare all of the contents. We are currently doing something similar when we detect similar-feature-reimports during PK-less import. Again, these are the features that, if detected, will *not* be renamed, but assigned the same PKs as previously. This algorithm is slow - quadratic on the number of features.

Because our similarity detection metric is actually "would these two features by identical, if we ignored one field", we can do a trick that git cannot. We can remove all of the fields of a feature, one at a time, and generate a hash for each of those. Then we can look for hash collisions for those hashes. There will be a lot of hashes to keep track of - if every feature has 10 fields, then there will be 10 times as many hashes as features. Instead of being quadratic on the number of features, this algorithm is:
O(N * M) where N is the number of inserts + deletes, and M is the number of fields a feature has.

Also changed the tests to actually make the algorithm work harder - previously it was only trying to find renames among 4 inserts + deletes, now there are 1400 insert + deletes which it has to hunt through. With the algorithm change, the test runs in 4 seconds instead of 5 seconds on my machine - haven't profiled beyond this yet, not sure how much is setup time etc. Actually mirroring a repo which has a lot of churn would give much more useful data anyway.

Also increased the default rename detection limit up to 500.

### DIFF
--- a/sno/import_source.py
+++ b/sno/import_source.py
@@ -119,6 +119,14 @@ class ImportSource:
         """
         return Schema.from_column_dicts(self.get_meta_item("schema.json"))
 
+    def align_schema_to_existing_schema(self, existing_schema):
+        """
+        Aligning the schema with an existing schema means that the pre-existing colunms will keep the same ID
+        that they had last time. Failing to align the schema would mean that some features would be re-encoded
+        even if they hadn't actually chaned.
+        """
+        self.schema = existing_schema.align_to_self(self.schema)
+
     @property
     def has_geometry(self):
         return self.schema.has_geometry
--- a/sno/init.py
+++ b/sno/init.py
@@ -162,7 +162,7 @@ class GenerateIDsFromFile(StringFromFile):
     "--similarity-detection-limit",
     hidden=True,
     type=click.INT,
-    default=500,
+    default=10000,
     help=(
         "When replacing an existing dataset where primary keys are auto-generated: the maximum number of unmatched "
         "features to search through for similar features, so that primary keys can be reassigned for features that "
--- a/sno/init.py
+++ b/sno/init.py
@@ -310,9 +310,7 @@ def import_table(
                     dest_path=dest_path,
                     similarity_detection_limit=similarity_detection_limit,
                 )
-                import_source.schema = existing_ds.schema.align_to_self(
-                    import_source.schema
-                )
+                import_source.align_schema_to_existing_schema(existing_ds.schema)
         import_source.dest_path = dest_path
         import_sources.append(import_source)
 
--- a/sno/pk_generation.py
+++ b/sno/pk_generation.py
@@ -4,9 +4,7 @@ import pygit2
 from .dataset2 import Dataset2
 from .import_source import ImportSource
 from .exceptions import NotYetImplemented
-from .serialise_util import (
-    json_pack,
-)
+from .serialise_util import json_pack, msg_pack
 from .schema import ColumnSchema, Schema
 
 
@@ -106,6 +104,8 @@ class PkGeneratingImportSource(ImportSource):
         self.prev_dest_tree = self._prev_import_dest_tree(repo)
 
         if not self.prev_dest_tree:
+            self.prev_dest_dataset = None
+            self.prev_dest_schema = None
             self.pk_col = self.DEFAULT_PK_COL
             self.primary_key = self.pk_col["name"]
             self.pk_to_hash = {}
@@ -116,6 +116,7 @@ class PkGeneratingImportSource(ImportSource):
             return
 
         self.prev_dest_dataset = Dataset2(self.prev_dest_tree, self.dest_path)
+        self.prev_dest_schema = self.prev_dest_dataset.schema
 
         data = self.prev_dest_dataset.get_meta_item(self.GENERATED_PKS_ITEM)
         self.pk_col = data["primaryKeySchema"]
@@ -200,6 +201,15 @@ class PkGeneratingImportSource(ImportSource):
         cols = self.delegate.schema.to_column_dicts()
         return Schema.from_column_dicts([self.pk_col] + cols)
 
+    def _schema_is_similar_to_last_import(self):
+        if not self.prev_dest_schema:
+            return False
+        dt = self.prev_dest_schema.diff_type_counts(self.schema)
+        if dt["pk_updates"] > 0 or dt["inserts"] > 1 or dt["deletes"] > 1:
+            return False
+
+        return True
+
     def features(self):
         # Next primary key to use if we can't find a historical but unassigned one in hash_to_unassigned_pks.
         next_new_pk = self.first_new_pk
@@ -214,6 +224,9 @@ class PkGeneratingImportSource(ImportSource):
         # become edits.
         buffered_inserts = []
         buffered_insert_limit = self.similarity_detection_insert_limit
+
+        if not self._schema_is_similar_to_last_import():
+            buffered_insert_limit = 0
 
         for orig_feature in self.delegate.features():
             feature = {self.primary_key: None, **orig_feature}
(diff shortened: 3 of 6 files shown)


sno/init.py
```diff
@@ -162,7 +162,7 @@ def convert(self, value, param, ctx):
     "--similarity-detection-limit",
     hidden=True,
     type=click.INT,
-    default=50,
+    default=500,
```
craigds, line 165: since this feature is now linear-time, I'd suggest a much higher limit like 10K or even 100K. assuming 10KB of memory per feature (fairly large features) then 100K features is only 1GB of memory, which seems passable (and in low-memory environments the user can override this flag)

craigds (approved):
looks great

Now that the test handles a slightly decent amount of data, is it worth adding the `benchmark` fixture to the test and benching it? Then comparing the old behaviour to the new, just so we can get a feel for how much difference it makes with 1400-feature changesets.

ofc even 1400 feature changesets is quite small but would probably have caused the old algorithm some trouble; it'd just be nice to verify that improvement with a benchmark

sno/pk_generation.py
```diff
@@ -406,3 +453,45 @@ def feature_blobs(self):
...
+            value = self.feature[col.name]
+            value_hash = hash(msg_pack([col.id, value]))
+            feature_hash ^= value_hash
+            value_hashes.append(value_hash)
+
+        # Add the feature at its overall hash.
+        yield feature_hash
+        # Remove each piece one at a time to make a hash for each variation of the feature missing exactly one field.
+        for value_hash in value_hashes:
+            yield feature_hash ^ value_hash
```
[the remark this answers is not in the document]
craigds: oh - does this algorithm mean that these features have the same hash value?

```python
{'a': 1, 'b': 2, 'c': 3}
{'a': 2, 'b': 3, 'c': 1}
```


because it seems to be only hashing the values and not in an order-dependent way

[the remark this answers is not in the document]
craigds: ah, ignore; I missed the `col.id` in `hash(msg_pack([col.id, value]))` above

sno/pk_generation.py
```diff
@@ -406,3 +452,45 @@ def feature_blobs(self):
...
+    def __eq__(self, other):
+        return other is self
+
+    def all_related_hashes(self):
+        value_hashes = []
+        feature_hash = 0
+        # Generate the overall hash of the feature, but in separate pieces.
+        for col in self.schema.non_pk_columns:
+            value = self.feature[col.name]
+            value_hash = hash(msg_pack([col.id, value]))
```
rcoup, line 477: does `msg_pack()` here add much value over just hashing `(col.id, value)`?

sno/pk_generation.py
```diff
@@ -406,3 +452,45 @@ def feature_blobs(self):
...
+    def try_add_match(self, other):
+        if self.match is None:
+            self.match = other
+        elif self.match == other:
+            pass
+        else:
+            self.match = self._MORE_THAN_ONE_MATCH
+
+    def is_uniquely_matched(self):
+        return self.match is not None and self.match.match is self
```
rcoup, line 496: if `self.match == self. _MORE_THAN_ONE_MATCH` then `self.match.match` will throw an `AttributeError` here?

sno/pk_generation.py
```diff
@@ -406,3 +452,45 @@ def feature_blobs(self):
             pk = self.decode_path_to_1pk(blob.name)
             if self.pk_filter(pk):
                 yield blob
+
+
+class FeaturePlusHashes:
+    _MORE_THAN_ONE_MATCH = object()
```
rcoup, line 458: worth adding [`__slots__`](https://docs.python.org/3.9/reference/datamodel.html#invoking-descriptors) here since we're potentially creating a lot of them?

rcoup (commented):
One potential bug, but overall LGTM 💯

sno/pk_generation.py
craigds: I think I hit this (it's very uncommon though)
```
   [14646] Failed to execute script sno_cli
     File "sno_cli.py", line 4, in <module>
     File "site-packages/click/core.py", line 829, in __call__
     File "site-packages/click/core.py", line 782, in main
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/cli.py", line 138, in invoke
     File "site-packages/click/core.py", line 1259, in invoke
     File "site-packages/click/core.py", line 1066, in invoke
     File "site-packages/click/core.py", line 610, in invoke
     File "site-packages/click/decorators.py", line 21, in new_func
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/init.py", line 329, in import_table
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/fast_import.py", line 208, in fast_import_tables
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/fast_import.py", line 271, in write_blobs_to_stream
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/dataset2.py", line 317, in import_iter_feature_blobs
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/pk_generation.py", line 266, in features
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/pk_generation.py", line 303, in _match_similar_features_and_remove
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/pk_generation.py", line 346, in _fast_pop_similar_pairs
     File "site-packages/sno-0.7.2.dev0+ci.1851.git52675e9d-py3.7.egg/sno/pk_generation.py", line 496, in is_uniquely_matched
   AttributeError: 'object' object has no attribute 'match'
   fatal: stream ends early
```

olsen232: Oops! Fixed it

sno/pk_generation.py
olsen232: Um... it's safe in that, all the values we store are guaranteed to message packable, and the hashes of message packed values are guaranteed to depend on their content, not their identity.
But, just looking through our types, the only weird one is geometry. And that's still just a glorified bytestring. So, you're right, we don't need to message pack here.

rcoup: Sure, feels like we could get another speed up avoiding the encoding.

Even for a `Geometry` type we could add `__hash__()` that returns `hash(self.wkb)` or similar.

sno/import_source.py
```diff
@@ -119,6 +119,14 @@ def _init_schema(self):
         """
         return Schema.from_column_dicts(self.get_meta_item("schema.json"))
 
+    def align_schema_to_existing_schema(self, existing_schema):
+        """
+        Aligning the schema with an existing schema means that the pre-existing colunms will keep the same ID
```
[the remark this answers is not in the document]
olsen232: Done

sno/import_source.py
```diff
@@ -119,6 +119,14 @@ def _init_schema(self):
...
+        that they had last time. Failing to align the schema would mean that some features would be re-encoded
+        even if they hadn't actually chaned.
```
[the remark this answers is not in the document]
olsen232: Done

sno/pk_generation.py
```diff
@@ -200,6 +200,15 @@ def _init_schema(self):
         cols = self.delegate.schema.to_column_dicts()
         return Schema.from_column_dicts([self.pk_col] + cols)
 
+    def _schema_is_similar_to_last_import(self):
```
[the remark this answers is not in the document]
olsen232: Done

PR #386, opened 2021-03-25
Basic but working implementation of MySQL Working Copy.

olsen232:
## Description

Adds MySQL working copy support to Sno. Checking out, editing, diffing and committing of features themselves and of schema.json is supported. Also added tests - although, they won't run in our CI queue until we set up a MySQL database for them.

Still TODO:
- Include the pymysql library and any of its dependencies in our various builds.
- Create (/drop) spatial reference systems using the MySQL `CREATE SPATIAL REFERENCE SYSTEM` command.
- Make meta diffs other than schema.json work (eg crs/something.wkt, title, maybe description)
- Find a way to make the spatial index work
- Add documentation, update changelog

### DIFF
--- a/sno/sqlalchemy/create_engine.py
+++ b/sno/sqlalchemy/create_engine.py
@@ -115,6 +115,30 @@ def postgis_engine(pgurl):
     return engine
 
 
+CANONICAL_MYSQL_SCHEME = "mysql"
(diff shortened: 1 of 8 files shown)


olsen232:
Clicked merge in the wrong tab somehow - reverted - will re-push as a new PR later

PR #388, opened 2021-03-29
Use a server-side cursor for SQL importers

craigds:
## Description

importing from a postgres source, we noticed high memory consumption with large layers. Turns out sno is fetching all the results from the database rather than using a server-side cursor.

This PR makes it use a server side cursor to fetch results. [SQLAlchemy docs](https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=yield_per#engine-stream-results) say that it will use a low fetch size initially and eventually increase up to 1000 rows at a time, which seems reasonable.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
  * `init` now accepts a `--initial-branch` option
  * `clone` now accepts a `--filter` option (advanced users only)
  * `show -o json` now includes the commit hash in the output
+ * `import` from Postgres now uses a server-side cursor, which means sno uses less memory
 
 ## 0.7.1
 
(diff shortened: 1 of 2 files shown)


rcoup:
> use a low fetch size initially and eventually increase up to 1000 rows at a time, which seems reasonable.

Isn't 1000 rows only 1-3s of processing at 300-1000 features/second? Wouldn't 30-60s-ish of data be better in terms of DB load and reduced round-trips?

(Is there an advantage to starting small?)

olsen232 (approved):

craigds:
you might be right; a fetch size of 10K would be better.

Unfortunately I can't see a way to set the cursor fetch size with sqlalchemy 1.3. Upgrading to 1.4 gives us a way to control it, but I didn't want to do that in this PR

craigds:
Bumped fetch size to 10K in 3016a9dccf2e30e2375afb88980838338a4e50e2

PR #390, opened 2021-03-30
Optimised GPKG working copies for large datasets.

olsen232:
## Description

Sets page_size to 64KB - this is on the GPKG itself.
Sets cache_size to 200MB - this is done each time a session is opened.

### DIFF
--- a/sno/sqlalchemy/create_engine.py
+++ b/sno/sqlalchemy/create_engine.py
@@ -16,6 +16,8 @@ from sno import spatialite_path, is_windows
 from sno.geometry import Geometry
 from sno.exceptions import NotFound, NO_DRIVER
 
+GPKG_CACHE_SIZE_MiB = 200
+
 
 def gpkg_engine(path):
     def _on_connect(pysqlite_conn, connection_record):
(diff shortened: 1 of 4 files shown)


olsen232:
cc @hamishcampbell

sno/sqlalchemy/create_engine.py
```diff
@@ -16,6 +16,8 @@
 from sno.geometry import Geometry
 from sno.exceptions import NotFound, NO_DRIVER
 
+GPKG_CACHE_SIZE_MB = 200
```
craigds, line 19: make it in MiB, then the maths below makes sense (the pragma value is in KiB)

craigds (commented):
where'd the numbers you chose come from?

sno/sqlalchemy/create_engine.py
olsen232: Done

rcoup:
Benchmark tests show any difference?

hamishcampbell:
Huge improvement.

3M+ dataset before the change:

```
$ time sqlite3 data.gpkg "SELECT COUNT(building_id) FROM building_outlines;"
3320006
sqlite3 data.gpkg   0.94s user 5.60s system 10% cpu 1:04.05 total
```

After the change:

```
$ time sqlite3 data.gpkg "SELECT COUNT(building_id) FROM building_outlines;"
3320006
sqlite3 data.gpkg   0.24s user 0.25s system 99% cpu 0.488 total
```

olsen232:
Glad it's working! I found and fixed our benchmarks for importing and writing GPKGs, but nothing in particular showed up amongst the noise. Most of our benchmarks measure sno internal performance, not GPKG performance.

PR #392, opened 2021-04-01
Fix libcrypto warning on Ubuntu

craigds:
## Description

In non-dev sno, when running a command that invokes ssh via git
(clone, push, pull, fetch), SSH issues this warning twice:

```
ssh: /opt/sno/libcrypto.so.1.0.0: no version information available (required by ssh)
ssh: /opt/sno/libcrypto.so.1.0.0: no version information available (required by ssh)
```

This is probably because Pyinstaller has set LD_LIBRARY_PATH to find
bundled libraries, but we don't bundle ssh.
So, (system) ssh is loading our custom libcrypto, which doesn't
have the versioning information it expects.

This commit un-overrides LD_LIBRARY_PATH to fix the issue.

## Related links:

https://pyinstaller.readthedocs.io/en/stable/runtime-information.html#ld-library-path-libpath-considerations

## Checklist:

### DIFF
--- a/sno/cli_util.py
+++ b/sno/cli_util.py
@@ -1,8 +1,9 @@
-import json
 import click
+import io
+import json
 import jsonschema
 import os
-import io
+import platform
 
 import pygit2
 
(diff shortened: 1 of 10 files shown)


sno/cli_util.py
```diff
@@ -45,6 +46,21 @@ def startup_load_git_init_config():
         os.environ["GIT_CONFIG_PARAMETERS"] = f"'init.defaultBranch=main'{existing}"
 
 
+def git_remote_environment():
```
rcoup, line 49: This needs to be for every git command, since we don't know with partial clones which commands might need to interact with a promisor remote.

rcoup (changes requested):
Tested as working 👍 

Since we need to expand it to every git invocation, could we just make it _every_ invocation of something external? Editors, pagers, ...

olsen232:
Makes sense. Failing on linux because of a typo - just fixing that

PR #393, opened 2021-04-01
Fix test_gdal_proj_data - files have moved around

olsen232:
Simplest fix for failing GDAL test on Windows - started failing when we upgraded gdal-3.1.4 to gdal-3.2.2 - Rob to verify if this test is sufficient, or if further testing is needed.


olsen232:
Approved (I can't actually approve this since it's my own PR)
But I checked the maths and it makes sense. I'll merge it

PR #394, opened 2021-04-01
Fix reporting of windows tests

olsen232:
Fixes windows test failures to actually show up as a red build.

Uses a divide by zero since, inside the loop, variable substitutions don't work properly (even if you turn on enabledelayedexpansion - not sure why) - so set /A is the only command I could find that could read a variable, so I needed set /A to be the command that causes the build to fail. And outside the loop, the variable is for some reason reset to its pre-loop value, so I needed the command to fail due to a set /A from inside the loop.

This drove me a bit insane actually. I hope it actually works

### DIFF
--- a/makefile.vc
+++ b/makefile.vc
@@ -62,6 +62,7 @@ test: $(app) venv\.test.installed
 
 ci-test:
 	set CI=true
+	set failed=0
 	FOR /L %%I IN (0, 1, 9) DO \
 		pytest \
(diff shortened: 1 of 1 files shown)


olsen232:
I merged the gdal fix earlier today even though it failed on windows - I thought I checked it properly but evidently I did not. I better merge this so that at least we can see what we are doing

Issue #353, opened 2021-02-05
Support for SQL Server working copy


olsen232:
This one:
https://docs.microsoft.com/en-us/sql/relational-databases/spatial/spatial-data-sql-server?view=sql-server-ver15

olsen232:
Fixed as of https://github.com/koordinates/kart/tree/v0.8.0

PR #395, opened 2021-04-02
ODBC packaging fixes

rcoup:
## Description

1. Switch macOS and Windows back to pulling in the PyPI binary wheels for pyodbc. The pyodbc wheels don't include unixODBC. 
2. Get PyInstaller to ignore libodbc on macOS & Linux — this means pyodbc won't import on macOS/Linux without installing unixODBC. And we don't need to build unixODBC on Linux.
3. Make our deb packaging recommend installing unixodbc. RPM doesn't have a concept of "recommends".
4. Get Sno to handle import errors for pyodbc — the MS SQL Server drivers depend on unixodbc, so installing them (via homebrew on macOS + via the package manager on Linux) will resolve (2)
5. Get PyInstaller to ignore libstdc++ & libgcc on Linux — this avoids conflicts across c++ versions by always using the OS version — they're backwards compatible so since we compile against an old version this is ok. This is the same approach manylinux takes.
6. Github actions fix — lots of artifact upload errors including falling through retries as "success". Swap from upload-artifact@v2-preview to upload-artifact@v2.

Tested manually on Windows + macOS, and tested via a [modified e2e-1.sh script](https://gist.github.com/rcoup/015dbc36794eebb790914bc4d9b272da) across all the Linux distro releases covered under the packaging `test-{rpm,dev}-all` tests.

## Related links:

#353 

## Checklist:

- [ ] Have you updated the [changelog](https://github.com/koordinates/sno/blob/master/CHANGELOG.md)? — it's kinda included under the wider SQL Server addition?


olsen232 (approved):
Honestly there's not much chance I'd be able to spot an issue
But looks good from here
I'll test the RC when we get that far

PR #400, opened 2021-04-08
`import --replace-existing` tweaks & optimisations

craigds:
## Description

When importing features, if there is also a schema change, we compare new blobs against corresponding existing blobs. This is so we can avoid writing subtly-different-but equivalent blobs to the repo unnecessarily.

This is a slow process since we need to fetch all the blobs to do that, so until now we were only running the comparison when a schema change occurred in the commit.

However, during a discussion with @olsen232 we realised that a later non-schema-change commit also needs to run the feature blob comparison, since any features imported with the new schema will vary from the previously-imported feature blob with the old schema.

This PR makes that adjustment (run the feature blob comparison during any import in a repo which has previously had a schema change).

Of necessity I looked into performance to try and mitigate the impact this would have. I found a couple of things which are helpful:

* d1f5dbd9f7f46fc666dfccde6423b69047ecff71 sped up the feature comparison by ~30-40% 👍 
* Increasing the libgit2 object cache size to allow for all sno's tree objects improved the performance by a _further_ ~80-90%.

For reference in a repo with 1M features on my laptop, I achieved the following results for the first 200K features imported in the followup revision:

* no feature comparison, on master: `200,000 features... @11.7s`
* with feature comparison before d1f5dbd9f7f46fc666dfccde6423b69047ecff71: `200,000 features... @68.5s`
* with feature comparison after d1f5dbd9f7f46fc666dfccde6423b69047ecff71: `200,000 features... @47.8s`
* with feature comparison after this PR: `200,000 features... @16.5s`

This also speeds up `sno show` on a large (1M rows imported) commit by 3.5x (4m38 --> 1m45), I guess due to the same issue of trees being looked up repeatedly.

## Related links:

https://github.com/libgit2/libgit2/blob/508361401fbb5d87118045eaeae3356a729131aa/include/git2/common.h#L266

## Checklist:

### DIFF
--- a/sno/dataset2.py
+++ b/sno/dataset2.py
@@ -290,9 +290,9 @@ class Dataset2(RichBaseDataset):
             # Optimisation: Try to avoid rewriting features for compatible schema changes.
             # this can take some time, but often results in much fewer git objects produced.
             # For example, consider revision A with this feature:
-            #   {"a": 1, "b": 2}
+            #   {"x": 1, "y": 2}
             # Now at revision B we add a nullable column, so the new version of the feature is:
-            #   {"a": 1, "b": 2, "c": NULL}
+            #   {"x": 1, "y": 2, "z": NULL}
             # In this situation, during the import of revision B, we can avoid writing the
             # new blob, because the existing feature blob from revision A can be 'upgraded'
             # to the new schema without changing anything.
--- a/sno/dataset2.py
+++ b/sno/dataset2.py
@@ -282,40 +282,51 @@ class Dataset2(RichBaseDataset):
                 blob.data,
             )
 
-    def import_iter_feature_blobs(self, resultset, source, replacing_dataset=None):
+    def import_iter_feature_blobs(
+        self, repo, resultset, source, replacing_dataset=None
+    ):
         schema = source.schema
-        if replacing_dataset is not None and replacing_dataset.schema != source.schema:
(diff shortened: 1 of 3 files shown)


olsen232 (commented):
Agree with this in principal - but I don't like that the _should_compare_imported_features_against_old_features is in dataset2 - prefer it to be in some file to do with imports.

We can just change import_iter_feature_blobs to ALWAYS try to reuse features from replacing_dataset - if it is supplied - and document that behaviour. Then, the import code should decide whether or not to supply a dataset, based on whether or not it is possible to avoid rewriting features.

sno/dataset2.py
```diff
@@ -282,40 +282,51 @@ def iter_legend_blob_data(self):
...
-                        )
-                    except KeyError:
-                        # this feature isn't in the dataset we're replacing
-                        yield self.encode_feature(feature, schema)
-                        continue
-
-                    existing_feature_raw_dict = replacing_dataset.get_raw_feature_dict(
-                        pk_values, data=existing_data
+            # this can take some time, but often results in much fewer git objects produced.
+            # For example, consider revision A with this feature:
```
olsen232, line 292: Try to avoid reusing the letters `a` and `b` more than once to mean unrelated things (even in different letter cases). You can use x and y, r1 and r2... anything else you like

olsen232 (approved):

PR #401, opened 2021-04-08
Make `status` faster for bulk workingcopy changes

craigds:
## Description
This commit avoids generating a DeltaDiff for feature changes, and
instead just counts the features for updates/deletes/inserts change
types. The diff was only used to do rename detection.

**caveat: As a result `status` no longer detects renames (updates to the
PK field) which makes it slightly inconsistent with `diff`**

I'm not sure how big a deal that caveat is.

On my laptop in a repo with 500K new inserts in the working copy, this speeds `status` from ~22s (master at 7761817e5b96d28364ad203ddca76ba2ae2f3e5c) to ~17s.

## Related links:

.

## Checklist:

### DIFF
--- a/sno/status.py
+++ b/sno/status.py
@@ -4,6 +4,7 @@ import click
 import pygit2
 
 from .conflicts import list_conflicts
+from .filter_util import UNFILTERED
 from .output_util import dump_json_output
 from .merge import merge_status_to_text
 from .merge_util import MergeContext, MergeIndex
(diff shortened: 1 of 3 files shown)


sno/status.py
```diff
@@ -77,10 +78,30 @@ def get_working_copy_status_json(repo):
...
+    wc_diff = working_copy.diff_to_tree(repo_filter=repo_filter)
     if wc_diff:
         output["changes"] = get_diff_status_json(wc_diff)
 
+    # now add feature counts. saves a bit of time computing diffs for features,
+    # when we really only need the counts here.
+    for dataset in datasets:
+        meta_diff = working_copy.diff_db_to_tree_meta(dataset)
+        feature_counts = working_copy.diff_db_to_tree_feature_counts(
+            dataset, UNFILTERED, meta_diff
```
rcoup, line 96:
```suggestion
            dataset,
            feature_filter=UNFILTERED,
            meta_diff=meta_diff,
```

sno/working_copy/base.py
```diff
@@ -582,26 +583,72 @@ def diff_db_to_tree_feature(
...
 
         if find_renames and (insert_count + delete_count) <= 400:
             self.find_renames(feature_diff, dataset)
 
         return feature_diff
 
+    def diff_db_to_tree_feature_counts(self, dataset, feature_filter, meta_diff):
+        """
+        Returns a dict which counts feature changes between the repo structure and this working copy.
+        The returned dict contains keys 'update', 'delete' and 'insert', and values are integers.
```
rcoup, line 628:
```suggestion
        The returned dict contains keys 'updates', 'deletes' and 'inserts', and values are integers.
```

rcoup (commented):
Why does the test output change?

craigds:
> Why does the test output change?

as noted in:

> caveat: As a result status no longer detects renames (updates to the
> PK field) which makes it slightly inconsistent with diff

Both those two tests have one feature whose PK has been modified

olsen232 (commented):
I'm okay with it, although I was hoping for a bit more than 25% speedup - it barely seems worth losing rename detection...

Can we keep the best of both worlds? We could check the number of tracked features before deciding whether to do the slow way or the fast way. 10,000 tracked features could mean anything - it could mean 0 deltas, or 10,000 deltas - but it most likely means about 10,000 deltas. Which mostly likely means we'll have too many deltas to do rename detection on anyway (unless we have about 9900 mods, and about 100 inserts and deletes). So, 10,000 tracked features in a dataset probably means we may as well run the fast code. But for 10 tracked features, we should probably run the slow code. Is it cheap to count the number of tracked features for a particular dataset?

craigds:
in retrospect this just doesn't seem worth the tradeoffs, and we lost interest. Closing

Issue #405, opened 2021-04-11
startup is a bit slow (200-600ms)


craigds:
`kart branch` (in a non-repo directory) spends 600ms loading code before it can give me an error message. This seems consistent across various commands. Built Kart on macOS seems faster than dev Kart, presumably due to pyinstaller being faster at doing imports, but is still consistently >200ms.

It seems important (for scripted usage especially) that we trim that startup time down - perhaps 200ms would be acceptable. Seems unlikely we'd get below 100ms without rewriting (the entrypoint at least) in a compiled language.

I used https://github.com/nschloe/tuna to do a basic profile and it seems like itt's all just doing regularish python imports. Interestingly quite a bit of the time is in sqlalchemy:

* 10% in `working_copy/*.py`
* 23% loading sqlalchemy

1. It might be helpful to just defer import of workingcopy/* and sqlalchemy until we need to do something with the working copy.

2. Perhaps we could segregate the startup by command: rather than loading all the commands on startup we could just load the one that's being invoked at present. This seems difficult to do with Click though.

craigds:
another idea: write the basic `sno` entrypoint in go, and delegate most commands to python (just execvp the existing python entrypoint I guess).

Commands that just execvp to `git` can trivially be rewritten in Go to save on startup overhead, and perhaps over time other commands can move to Go too.

I say Go because it's reasonably pythonic, high level and has a pretty good package ecosystem, and [`git2go`](https://github.com/libgit2/git2go) looks quite complete. And @olsen232 is already quite familiar with it. Open to other language suggestions.

If we proceed on that path we should do some performance comparison of git2go vs pygit2 for large repos.

pfw:
`kart helper` has been released in v0.11.5. This is c client to a long running, forking process that does the actual kart work.

PR #406, opened 2021-04-11
diff: Add a way to get diff size estimates

craigds:
## Description

Large diffs are currently quite slow. Even if we can optimise a lot,
it's not uncommon to diff an import commit and produce 50M features
of output, which can take minutes/hours to generate and be
difficult or impossible to handle.

This change adds a size estimate feature so scripts can avoid showing
large diffs altogether.

The feature is implemented via a flag:

```
  --only-feature-count [veryfast|fast|medium|good|exact]
                                  Returns only a feature count (the number of
                                  features modified in this diff). If the
                                  value is 'exact', the feature count is exact
                                  (this may be slow.) Otherwise, the feature
                                  count will be approximated with varying
                                  levels of accuracy.
```

`fast` is normally good enough for a ballpark estimate.

## strategies

The different values vary in strategy as well as confidence intervals.

* `veryfast` always samples only one of 65536 trees.
* `exact` always samples all 65536 trees.
* `fast`, `medium` and `good` all iterate until they achieve a particular confidence that the feature count is within 10% of the sample feature count. The algorithm is roughly:
    - start with `sample_size` trees (fast: 2, medium: 8, good: 16)
    - use the sampled trees to estimate the required sample size for a particular confidence interval (fast: 60%; medium: 80%; good: 95%)
    - if the sample size is >= the required sample size, stop (result is the sample mean multiplied by sample size). otherwise, repeat

One small sample seems to be plenty for a large diff, because each tree has enough features that there is a low variance in the results.

For smaller repos, the first sample often yields no features, or very few features, and we have to increase the sample size.

### `exact` (slow):

```
$ time sno diff e39a101444d721abc7f2105e5eb81ecce74011ca^...e39a101444d721abc7f2105e5eb81ecce74011ca --only-feature-count=exact
landonline_setup:
	1040738 features changed

real	2m2.328s
```

(note that generating the full diff takes 7m56s for this commit so even though 2m2 is slow, it's still a bit of a win if you need the exact feature count)

```
$ sno -vv diff cbc25ebc80a670ab1b092d18fd4f323c79b244f8^...cbc25ebc80a670ab1b092d18fd4f323c79b244f8 --only-feature-count=exact
nz_waca_adjustments:
	228 features changed
```

### `good`:

```
$ time sno -vv diff e39a101444d721abc7f2105e5eb81ecce74011ca^...e39a101444d721abc7f2105e5eb81ecce74011ca --only-feature-count=good
2021-04-13 14:13:28,719 T4690804160 DEBUG sno.diff [diff.py:360] - sampling 16 trees for dataset landonline_setup
2021-04-13 14:13:28,845 T4690804160 DEBUG sno.diff [diff.py:408] - sampled 16 trees in 0.126s (ƛ=15.562, s=3.346). required: 12.6 (margin: 101990.4; confidence: 95%)
landonline_setup:
	1019904 features changed

real	0m0.725s
```

```
$ time sno -vv diff cbc25ebc80a670ab1b092d18fd4f323c79b244f8^...cbc25ebc80a670ab1b092d18fd4f323c79b244f8 --only-feature-count=good
2021-04-13 14:18:10,208 T4521086400 DEBUG sno.diff [diff.py:360] - sampling 16 trees for dataset nz_waca_adjustments
2021-04-13 14:18:10,215 T4521086400 DEBUG sno.diff [diff.py:390] - sampled 16 trees in 0.007s, found 0 features; increased sample size to 16384
2021-04-13 14:18:10,215 T4521086400 DEBUG sno.diff [diff.py:360] - sampling 16384 trees for dataset nz_waca_adjustments
2021-04-13 14:18:10,260 T4521086400 DEBUG sno.diff [diff.py:408] - sampled 16384 trees in 0.045s (ƛ=0.004, s=0.062). required: 65536.0 (margin: 24.8; confidence: 95%)
2021-04-13 14:18:10,260 T4521086400 DEBUG sno.diff [diff.py:360] - sampling 65536 trees for dataset nz_waca_adjustments
2021-04-13 14:18:10,415 T4521086400 DEBUG sno.diff [diff.py:408] - sampled 65536 trees in 0.155s (ƛ=0.003, s=0.059). required: 65536.0 (margin: 22.8; confidence: 95%)
nz_waca_adjustments:
	228 features changed

real	0m0.795s
```

### `fast`:

```
$ time sno -vv diff e39a101444d721abc7f2105e5eb81ecce74011ca^...e39a101444d721abc7f2105e5eb81ecce74011ca --only-feature-count=fast
2021-04-13 14:12:41,661 T4425674176 DEBUG sno.diff [diff.py:360] - sampling 2 trees for dataset landonline_setup
2021-04-13 14:12:41,788 T4425674176 DEBUG sno.diff [diff.py:408] - sampled 2 trees in 0.127s (ƛ=13.500, s=0.707). required: 0.0 (margin: 88473.6; confidence: 60%)
landonline_setup:
	884736 features changed

real	0m0.739s
```

```
$ time sno -vv diff cbc25ebc80a670ab1b092d18fd4f323c79b244f8^...cbc25ebc80a670ab1b092d18fd4f323c79b244f8 --only-feature-count=fast
2021-04-13 14:18:36,896 T4787953088 DEBUG sno.diff [diff.py:360] - sampling 2 trees for dataset nz_waca_adjustments
2021-04-13 14:18:36,902 T4787953088 DEBUG sno.diff [diff.py:390] - sampled 2 trees in 0.007s, found 0 features; increased sample size to 2048
2021-04-13 14:18:36,902 T4787953088 DEBUG sno.diff [diff.py:360] - sampling 2048 trees for dataset nz_waca_adjustments
2021-04-13 14:18:36,913 T4787953088 DEBUG sno.diff [diff.py:408] - sampled 2048 trees in 0.011s (ƛ=0.003, s=0.054). required: 2301.8 (margin: 19.2; confidence: 60%)
2021-04-13 14:18:36,913 T4787953088 DEBUG sno.diff [diff.py:360] - sampling 4096 trees for dataset nz_waca_adjustments
2021-04-13 14:18:36,929 T4787953088 DEBUG sno.diff [diff.py:408] - sampled 4096 trees in 0.015s (ƛ=0.003, s=0.058). required: 1971.5 (margin: 22.4; confidence: 60%)
nz_waca_adjustments:
	224 features changed

real	0m0.634s
```

## todo:

- [x] add tests

## Related links:

- #405 is related (even with `veryfast` this command still takes a little while to run)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Please note that compatibility for 0.x releases (software or repositories) isn't
 _When adding new entries to the changelog, please include issue/PR numbers wherever possible._
 
 
+## 0.9.0 (UNRELEASED)
+
+* `diff` now accepts `--only-feature-count`, which produces a feature count for the diff. The feature count can be exact or a fast estimate.
+* `log` now accepts `--with-feature-count` which adds a feature count to each commit when used with `-o json`. The feature count can be exact or a fast estimate.
+
 ## 0.8.0
 
 ### Breaking changes
--- a/sno/diff.py
+++ b/sno/diff.py
@@ -245,13 +245,9 @@ def feature_count_diff(
 
     repo = ctx.obj.repo
     base_rs, target_rs, working_copy = _parse_diff_commit_spec(repo, commit_spec)
-    if working_copy:
-        raise NotImplementedError(
-            "--only-feature-count isn't supported for working-copy diffs yet"
-        )
 
     dataset_change_counts = diff_estimation.estimate_diff_feature_counts(
-        base_rs, target_rs, accuracy
+        base_rs, target_rs, working_copy, accuracy
     )
 
     if output_format == "text":
--- a/sno/diff.py
+++ b/sno/diff.py
@@ -1,9 +1,6 @@
 import logging
 import re
-import statistics
-import subprocess
 import sys
-import time
 from pathlib import Path
 
 import click
@@ -26,6 +23,7 @@ from .exceptions import (
 from .filter_util import build_feature_filter, UNFILTERED
 from .output_util import dump_json_output
 from .repo import SnoRepoState
+from . import diff_estimation
 
 
 L = logging.getLogger("sno.diff")
@@ -233,61 +231,6 @@ def diff_with_writer(
             sys.exit(1)
 
 
-FEATURE_SUBTREES_PER_TREE = 256
-FEATURE_TREE_NESTING = 2
-MAX_TREES = FEATURE_SUBTREES_PER_TREE ** FEATURE_TREE_NESTING
-
-
-def _feature_count_sample_trees(rev_spec, feature_path, num_trees):
-    num_full_subtrees = num_trees // 256
-    paths = [f"{feature_path}{n:02x}" for n in range(num_full_subtrees)]
-    paths.extend(
-        [
-            f"{feature_path}{num_full_subtrees:02x}/{n:02x}"
-            for n in range(num_trees % 256)
-        ]
-    )
-
-    p = subprocess.Popen(
-        [
-            "git",
-            "diff",
-            "--name-only",
-            "--no-renames",
-            rev_spec,
-            "--",
-            *paths,
-        ],
-        stdout=subprocess.PIPE,
-        encoding="utf-8",
-    )
-    tree_samples = {}
-    for line in p.stdout:
-        # path/to/dataset/.sno-dataset/feature/ab/cd/abcdef123
-        # --> ab/cd
-        root, tree, subtree, basename = line.rsplit("/", 3)
-        k = f"{tree}/{subtree}"
-        tree_samples.setdefault(k, 0)
-        tree_samples[k] += 1
-    p.wait()
-    r = list(tree_samples.values())
-    r.extend([0] * (num_trees - len(r)))
-    return r
-
-
-Z_SCORES = {
-    0.50: 0.0,
-    0.60: 0.26,
-    0.70: 0.53,
-    0.75: 0.68,
-    0.80: 0.85,
-    0.85: 1.04,
-    0.90: 1.29,
-    0.95: 1.65,
-    0.99: 2.33,
-}
-
-
 def feature_count_diff(
     ctx,
     output_format,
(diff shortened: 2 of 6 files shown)


UberMouse:
Can we get sno to use this internally when generating the JSON commit log? Need to be able to both see how large of a diff the uncommitted changes in the working copy will be (which this is great for as is) and how large the diff for each commit will be. We need to know how big the commit will be ASAP when clicking a commit because we can't even attempt to fetch a diff before we run this command, which would add 400-600ms of delay. If Sno could calculate it for the commits, we don't need to pay the startup cost for each commit like we would if the app had to calculate it for each commit one by one.

craigds:
interesting idea. I don't see why not. I guess that would be something along the lines of `sno log -o json a...b --feature-count-sample-trees=256`. Though obviously it will slow down the log

and if you don't specify the option then the size estimate is omitted from the log (doesn't make sense to include it by default since regular users won't want this)

I guess the option would be rejected or ignored for a non-json log; or would we want to display the estimated size in some kind of commit-trailer form? it's probably a moot point since this option is for app integrations and not for end-user usage.

I was under the impression you needed this in `diff` for estimating arbitrary diffs; is that correct?

I'm open to better option names btw :)

UberMouse:
>I was under the impression you needed this in diff for estimating arbitrary diffs; is that correct?

Yes, need it for `diff` to check that the uncommitted changes aren't too large to display, and for commit log to determine if the commit is too large to display.

craigds:
Currently this PR doesn't do anything with uncommitted changes. I could add in something but I don't have any clever suggestions for making it fast.

Essentially the best I can think of is: generate the diff for N% of the tracking table, counting features as we go, then extrapolate the feature count to the rest of the tracking table. I would guess 10% of the diff would be minimum required for a ballpark estimate.

rcoup:
> Currently this PR doesn't do anything with uncommitted changes. I could add in something but I don't have any clever suggestions for making it fast.

Isn't a useful estimate for the count of uncommitted changes simply (GPKG):
```sql
SELECT table_name, COUNT(*) AS change_count_estimate
FROM gpkg_sno_track
GROUP BY table_name;
```
Combined with "has the schema changed?", feels like in virtually all cases it'd be perfectly accurate? The obvious thing it gets wrong is "delete all the features then reinsert them with the same values", but surely that's an edge case?

olsen232:
Rob is right - last time we were looking at messing with `sno status` we were still talking about what it would do generally - so we were reluctant to break rename detection, or, let's call it, "no-op detection". But if we're just looking for a rough estimate, SELECT COUNT(*) is almost always good, and it's okay if we occasionally get it wrong.

craigds:
I've updated this PR to use a less confusing CLI approach - the user now needs to know nothing about trees or statistics, and only needs to select the relative importance of a good estimate vs a fast estimate, from 5 possible choices.

I'll work on adding it to `log` also.

olsen232 (approved):
Meets our fast-get-commit-size requirement 👍 
Is probably a little bit over-engineered 👍 
Looks good to me

olsen232 (approved):

PR #407, opened 2021-04-14
Faster startup by delayed loading of commands

craigds:
## Description

This trims ~150ms from startup in my dev setup, by just avoiding
import of all commands at startup.

We load the one command that's being run, unless we don't recognise
any command, or the user is running `help` (we need to load all commands
to provide useful help text)

## TODO

* [x] make the tests pass

## Related links:

re #405

## Checklist:

### DIFF
--- a/platforms/sno_cli.py
+++ b/platforms/sno_cli.py
@@ -1,4 +1,4 @@
 import sno.cli
 
 if __name__ == "__main__":
-    sno.cli.cli()
+    sno.cli.entrypoint()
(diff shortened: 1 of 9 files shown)


olsen232 (approved):
Good plan - I think you just need to fix it for those commands that have two names, ie a human-readable name "import" and another name that isn't a python keyword ("import_table" - possibly we should be calling it "import_" since that is the pythonic thing to do when you collide with a keyword)

PR #409, opened 2021-04-15
Make sno repos support kart branding

olsen232:
Only affects local-only objects: 
- folder names
- config variables
- working copy tables

Git objects - objects that are version controlled and can be pushed and pulled across repos, still support only Sno branding. Rebranded these with kart will be done separately - so that even if our mirrors continue to have Sno branded git objects indefinitely, and so will the clones of them, at least the clones of them will have Kart branded local-only objects.
Git objects are relatively hidden anyway, so are lower priority than, eg, folder names.

Sno repo branding is still the default - no visible way for user to enable Kart branding.
Repos that use Sno branding work just as before.
Tested=test pass
- need to add new tests to make sure Kart branded repos work too.

### DIFF
--- a/sno/upgrade/__init__.py
+++ b/sno/upgrade/__init__.py
@@ -73,7 +73,7 @@ def upgrade(ctx, source, dest):
     # action!
     click.secho(f"Initialising {dest} ...", bold=True)
     dest.mkdir()
-    dest_repo = SnoRepo.init_repository(dest, wc_path=None, bare=True)
+    dest_repo = SnoRepo.init_repository(dest, wc_location=None, bare=True)
 
     # walk _all_ references
     source_walker = source_repo.walk(
@@ -129,7 +129,7 @@ def upgrade(ctx, source, dest):
     click.secho("\nCompacting repository ...", bold=True)
     dest_repo.gc()
 
-    if "sno.workingcopy.path" in source_repo.config:
+    if source_repo.workingcopy_location:
         click.secho("\nCreating working copy ...", bold=True)
         subctx = click.Context(ctx.command, parent=ctx)
         subctx.ensure_object(context.Context)
@@ -213,12 +213,12 @@ def upgrade_to_tidy(source):
         )
 
     source_repo.ensure_supported_version()
-    if source_repo.is_tidy_style_sno_repo():
+    if source_repo.is_tidy_style:
         raise InvalidOperation(
             "Cannot upgrade in-place - source repo is already tidy-style"
         )
 
-    wc_path = source_repo.workingcopy_path
+    wc_loc = source_repo.workingcopy_location
     is_bare = source_repo.is_bare
 
     source_repo.free()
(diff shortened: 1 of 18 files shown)


sno/upgrade/__init__.py
```diff
@@ -241,7 +241,103 @@ def upgrade_to_tidy(source):
...
+    Changes the following:
+     - config variables:
+       * sno.repository.version -> kart.repostructure.version
+       * sno.workingcopy.path -> kart.workingcopy.location
+     - hidden directory name from .sno to .kart (if repo is tidy-style)
+     - rewrites .sno/index / .kart/index to contain "kart" extension
+     - recreates the working copy, if one exists:
+       * _sno_state table (or similar) -> _kart_state
+       * _sno_track table (or similar) -> _kart_track
+       * triggers related to _sno_track also have new names
```
rcoup, line 265:
```
+       * rewrites SNO_README.txt -> KART_README.txt
```

sno/upgrade/__init__.py
```diff
@@ -241,7 +241,103 @@ def upgrade_to_tidy(source):
...
 
     click.secho("In-place upgrade complete: repo is now tidy", fg="green", bold=True)
+
+
+@click.command()
+@click.pass_context
+@click.argument("source", type=click.Path(exists=True, file_okay=False), required=True)
+def upgrade_to_kart(ctx, source):
+    """
+    Upgrade in-place a Sno repository to be a Kart repository.
```
rcoup, line 255: Maybe need to clarify this _doesn't_ change any of the stored data (blobs/trees/commits/tags)?

tests/test_kart.py
```diff
@@ -0,0 +1,164 @@
...
+    table_ref,
+    data_archive,
+    tmp_path,
+    chdir,
+    cli_runner,
+    insert,
+    request,
+    monkeypatch,
+):
+    # Like test_e2e.py, but, uses a Kart branded repo instead of a Sno one.
```
rcoup, line 39: maybe swap this to be `test_e2e`, and make the Sno-flavoured one `test_e2e_sno`?

rcoup (commented):
👍

tests/test_kart.py
olsen232: Yeah I still need to do this - 
- when we turn on kart branding for new repos (not turned on as of this PR), test_e2e.py will become basically the same as test_kart - so at that point, I'll need to modify them as you suggest so that we have the standard e2e (kart) plus another e2e_sno.
- a lot of other tests will break (eg a test that creates a new repo with a new GPKG, and then tries to read the tracking table at sno_track). I'll need to update those to not break.
- eventually, I should upgrade the test .tgz repos that we have to be kart branded, not sno branded. This is lower priority though - it doesn't affect the user experience

sno/upgrade/__init__.py
olsen232: Done

sno/upgrade/__init__.py
olsen232: Done

PR #413, opened 2021-04-23
Fix URIs to refer to koordinates/kart

olsen232:
Can be merged as soon as github.com/koordinates/sno is renamed to github.com/koordinates/kart
Doesn't need to be merged immediately, since github.com promises to redirect from old name to new name.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,10 +4,10 @@
 
 ## Related links:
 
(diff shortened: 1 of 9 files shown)


craigds (changes requested):
`ack koordinates/sno` still finds quite a bit of stuff

craigds (approved):

PR #414, opened 2021-04-28
Add dependency on pymysql

olsen232:
Adds pymysql dependency to support MySQL working copy - see https://github.com/koordinates/kart/pull/399


craigds (approved):

olsen232:
Merged into https://github.com/koordinates/kart/pull/399

PR #416, opened 2021-04-29
Cache diff feature counts as annotations in sqlite

craigds:
### Description

We have a need for caching of slow computed results such as feature
counts (estimates and exact counts) for diffs and logs.

This change introduces a sqlite database, stored in the backend repo
directory, which stores generic annotations about objects in the repo.

Initially it stores feature counts (of any accuracy) computed using
`kart diff --only-feature-count=...`. Stored values are used as the
output of that command if they exist, rather than computing a new
value.

The value for `--only-feature-count=exact` is additionally
cached by consuming a full diff from `kart diff` or `kart show`.

The caching results can be exposed with `-vv`:
```
$ kart -vv diff 19fc5c60eb7dd8fb85b73838a42f6cb1c8b719c2^...19fc5c60eb7dd8fb85b73838a42f6cb1c8b719c2 --only-feature-count=exact
2021-04-29 17:08:03,062 T4531244544 DEBUG kart.annotations [__init__.py:68] - failed to retrieve: feature-change-counts-exact for 5f23ba76c999b17272c810a49e42b05e11031a4f...9108df6be590e932cacfb0372b4cad7d878af8a3
2021-04-29 17:08:03,174 T4531244544 DEBUG kart.annotations [__init__.py:32] - storing: feature-change-counts-exact for 5f23ba76c999b17272c810a49e42b05e11031a4f...9108df6be590e932cacfb0372b4cad7d878af8a3: {"layers_limit_of_moderate_wave_action": 2098}
layers_limit_of_moderate_wave_action:
    2098 features changed
```

```
$ kart -vv diff 19fc5c60eb7dd8fb85b73838a42f6cb1c8b719c2^...19fc5c60eb7dd8fb85b73838a42f6cb1c8b719c2 --only-feature-count=exact
2021-04-29 17:06:46,040 T4548242944 DEBUG kart.annotations [__init__.py:61] - retrieved: feature-change-counts-exact for 5f23ba76c999b17272c810a49e42b05e11031a4f...9108df6be590e932cacfb0372b4cad7d878af8a3: {'layers_limit_of_moderate_wave_action': 2098}
layers_limit_of_moderate_wave_action:
    2098 features changed
```

Additionally this change introduces a `build-annotations` command which explicitly builds annotations for commits, and may be useful to prepopulate annotations before they are needed.

```
$ kart build-annotations --all-reachable
Enumerating reachable commits...
Building feature change counts...
(1/6): 908a295 Title and description changed
(2/6): 5bbc380 34 features added, 329 mods, 16 dels
(3/6): 19fc5c6 516 features added, 1,059 mods, 523 dels
(4/6): 476c6ee 188 features added, 1,152 mods, 119 dels
(5/6): bb6787d 148 features added, 456 mods, 18 dels; title changed
(6/6): e4da31d 9,495 features added; description added
done.
```

### todo

* [x] update changelog
* [x] add tests


kart/annotations/cli.py
```diff
@@ -0,0 +1,74 @@
...
+    Builds annotations against commits; stores the annotations in a sqlite database.
+    """
+    repo = ctx.obj.repo
+    if all_reachable:
+        if refishes:
+            raise click.UsageError(
+                "--all-reachable and refishes are mutually exclusive"
+            )
+        click.echo("Enumerating reachable commits...")
+        commits = list(gen_reachable_commits(repo))
```
rcoup, line 55: why materialise it into a list? if it's for the `if` statement, doesn't seem to achieve much?

kart/annotations/db.py
```diff
@@ -0,0 +1,51 @@
...
+
+from kart.sqlalchemy.create_engine import sqlite_engine
+
+Base = declarative_base()
+
+
+class KartAnnotation(Base):
+    __tablename__ = "kart_annotations"
+    id = Column(Integer, nullable=False, primary_key=True)
+    # TODO: look into what indexes we should have here
```
rcoup, line 16: presumably a multi-column index... depends on what queries we'll likely need for _multiple objects_?

1. (`object_type`, `object_id`, `annotation_type`) would allow efficiently enumerating all annotations for object types, or a specific object.
2. (`annotation_type`, `object_type`, `object_id`) would allow enumerating all objects that have a specific annotation.

kart/annotations/db.py
```diff
@@ -0,0 +1,51 @@
...
+    object_type = Column(Text, nullable=False)
```
rcoup, line 17: does `object_type` provide any useful benefit?

kart/annotations/db.py
```diff
@@ -0,0 +1,51 @@
...
+    object_id = Column(Text, nullable=False)
+    annotation_type = Column(Text, nullable=False)
+    data = Column(Text, nullable=False)
```
rcoup, line 20: blob? I guess if it's predominantly json text is useful.

kart/sqlalchemy/create_engine.py
```diff
@@ -18,6 +18,21 @@
...
 
+def sqlite_engine(path):
+    """
+    An engine for non-spatial, non-GPKG sqlite databases.
+    """
+
+    def _on_connect(pysqlite_conn, connection_record):
+        pysqlite_conn.isolation_level = None
+        dbcur = pysqlite_conn.cursor()
+        dbcur.execute("PRAGMA foreign_keys = ON;")
```
rcoup, line 29: need to set journal mode to WAL on DB creation otherwise multiple processes will clash.

kart/sqlalchemy/create_engine.py
```diff
@@ -18,6 +18,21 @@
...
+
+    engine = sqlalchemy.create_engine(f"sqlite:///{path}", module=sqlite)
```
rcoup, line 31: This should handle running in a read-only repository where we can't create or write to the DB.

kart/annotations/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+        with annotations_session(self.repo) as session:
+            object_id = self._object_id(base_rs, target_rs)
+            data = json.dumps(data)
+            L.debug(
+                "storing: %s for %s: %s",
+                annotation_type,
+                object_id,
+                data,
+            )
+            session.add(
```
rcoup, line 34: Needs to upsert for concurrency... or handle that via catching the `COMMIT` error and ignoring?

rcoup: And read-only DBs/directories will obviously fail to write

kart/annotations/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+class DiffAnnotations:
+    def __init__(self, repo):
+        self.repo = repo
+
+    def _object_id(self, base_rs, target_rs):
+        # this is actually symmetric, so we can marginally increase hit rate by sorting first
+        if base_rs.tree.id < target_rs.tree.id:
+            return f"{base_rs.tree.id}...{target_rs.tree.id}"
+        else:
+            return f"{target_rs.tree.id}...{base_rs.tree.id}"
```
rcoup, line 17: I know we use `(base|target)_rs` throughout diff, but what does it mean?

rcoup:
```suggestion
        tree_ids = sorted(rs.tree.id for rs in (base_rs, target_rs))
        return f"{tree_ids[0]}...{tree_ids[1]}"
```

rcoup (changes requested):
Looking good! 🚀 

1. needs error handling
2. and tests
3. maybe for later, do we want a cleanup command (part of gc/fsck?) that either nukes the db altogether or cleans out unreachable stuff.

kart/annotations/cli.py
```diff
@@ -0,0 +1,74 @@
...
+        commit_shas_seen.add(commit.oid.hex)
+        yield commit
+
+
+@click.command(name="build-annotations")
+@click.pass_context
+@click.option(
+    "--all-reachable", is_flag=True, help="Build annotations for all reachable commits"
+)
+@click.argument(
```
rcoup, line 40: Presumably a common way to use this is in a post-receive hook. Normally it's the hook's responsibility to resolve pushed refs to a set of commit IDs which can then be passed to `build-annotations`, but there could be a a lot. Maybe a way to accept refishes via stdin?

kart/annotations/cli.py
craigds: I show the length (`1/7`) in the output

kart/annotations/db.py
craigds: Maybe/maybe not. I expect to annotate trees at some stage (to provide things like full-dataset feature counts).

If we were to have annotations for trees _and_ commits, they'll be opaque OIDs. They won't conflict (because OIDs don't) but it's still nice to be able to see what they are.

kart/annotations/cli.py
craigds: This is probably _only_ useful as a low-level command to be run in hooks; users aren't going to run it. I'll change it so the only way to send it IDs is via stdin, or use `--all-reachable`

kart/annotations/__init__.py
craigds: > upsert

this is only necessary if there is a unique constraint on `(object_type, object_id, annotation_type)`.

It's easiest to just not have that constraint. Then there's a possibility of more than one transaction inserting the same annotation, but we can just allow that and not have to worry about concurrency (in `get` I just fetch the first one, and don't worry too hard about the presence of others)

kart/annotations/__init__.py
craigds: Added notes to docstring about the repostructure objects

kart/annotations/db.py
craigds: really unless we're expecting _at least_ hundreds of thousands of these, any indexes at all might be completely unnecessary.

So far annotations don't get looked up for multiple objects at once. A future enhancement for `log` or `build-annotations` might change to doing that though.

All lookups at present are based on all three of `(object_type, annotation_type, object_id)`

rcoup: A crude representative sqlite test has the index as 3 times quicker even for 500 rows. Block cache obviously makes it go away fairly quickly, but considering this stuff will almost never be in the cache seems like minimal disk hits are key?

kart/annotations/db.py
rcoup: For how often it'd be used, is `git cat-file -t {oid}` (or `... | git cat-file --batch-check=%(objecttype)` in bulk) enough?

kart/annotations/__init__.py
rcoup: A unique constraint can have a default/automatic `ON CONFLICT` resolution, _and_ it's supported by SQLAlchemy... https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#on-conflict-support-for-constraints

You're presuming `data` is immutable as well?

kart/annotations/cli.py
rcoup: seems sane 👍

kart/annotations/__init__.py
craigds: yes data is immutable if we always use object IDs in the `object_id` field, and not refs or whatever

kart/annotations/db.py
craigds: ok

`(annotation_type, object_type, object_id)` seems like the best choice to me.

kart/annotations/__init__.py
rcoup: well, it's not really - if the counting method changes, we record different things (eg: adding add/mod/delete counts with some sort of "renames detection happened" flag) then `data` will change. Or even JSON not serialising deterministically (though iirc it's sorted in py3?)

rcoup:
Presume for the last 3 commits you did tests with r/o db files?

craigds:
well, for the last commit I did (`chflags uchg ./annotations.db` on macOS)

kart/annotations/__init__.py
craigds: ok, done

table now looks like 
```sql
CREATE TABLE kart_annotations (
	id INTEGER NOT NULL,
	object_id TEXT NOT NULL,
	annotation_type TEXT NOT NULL,
	data TEXT NOT NULL,
	PRIMARY KEY (id),
	CONSTRAINT kart_annotations_multicol UNIQUE (annotation_type, object_id) ON CONFLICT REPLACE
)
```

craigds:
> needs error handling

what did you have in mind specifically?

kart/annotations/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Returns a diff annotation from the sqlite database.
+        Returns None if it isn't found.
+
+        base_rs: base RepoStructure object for this diff (revA in a 'revA...revB' diff)
+        target_rs: target RepoStructure object for this diff (revB in a 'revA...revB' diff)
+        """
+        with annotations_session(self.repo) as session:
+            object_id = self._object_id(base_rs, target_rs)
+            for annotation in session.query(KartAnnotation).filter(
```
rcoup, line 64: presumably this will raise an exception if the DB/table doesn't exist. Need a test without an annotations DB.

kart/annotations/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        data = json.dumps(data)
+        try:
+            with annotations_session(self.repo) as session:
+                L.debug(
+                    "storing: %s for %s: %s",
+                    annotation_type,
+                    object_id,
+                    data,
+                )
+                session.add(
```
rcoup, line 39: presumably this won't create the table if the DB exists but is empty? Do we need a `CREATE TABLE IF NOT EXIST` query on connect?

kart/annotations/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+                    KartAnnotation(
+                        object_id=object_id,
+                        annotation_type=annotation_type,
+                        data=data,
+                    )
+                )
+        except OperationalError as e:
+            # ignore errors from readonly databases.
+            if "readonly database" in str(e):
```
rcoup, line 48: if the directory isn't writeable does this raise "readonly database" or an IO error? Via the sqlite CLI I get `Error: unable to open database "doo.db": unable to open database file`

tests/test_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+            r.message
+            for r in caplog.records
+            if "feature-change-counts-exact" in r.message
+        ]
+        assert messages == [
+            "retrieved: feature-change-counts-exact for 8feb827cf21831cc4766345894cd122947bba748...a8fa3347aed53547b194fc2101974b79b7fc337b: {'nz_pa_points_topo_150k': 5}"
+        ]
+
+
+def test_diff_feature_count_populates_annotations(data_archive, cli_runner, caplog):
```
rcoup, line 36: should run this in a few scenarios:
- no annotations DB
- read-only dir; can't create annotations DB
- annotations DB exists; has no table
- annotations DB exists; has table
- annotations DB exists; is read-only

tests/test_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+        assert r.exit_code == 0, r.stderr
+        messages = [
+            r.message
+            for r in caplog.records
+            if "feature-change-counts-fast" in r.message
+        ]
+        assert messages == [
+            "missing: feature-change-counts-fast for 8feb827cf21831cc4766345894cd122947bba748...a8fa3347aed53547b194fc2101974b79b7fc337b",
+            'storing: feature-change-counts-fast for 8feb827cf21831cc4766345894cd122947bba748...a8fa3347aed53547b194fc2101974b79b7fc337b: {"nz_pa_points_topo_150k": 5}',
+        ]
```
rcoup, line 52: I think we should check the expected data is actually in the annotations DB in these tests.

kart/annotations/__init__.py
rcoup: i guess this comes back to the basic behaviour expectations. I assumed:

* if you've run `build-annotations`, populate the DB and use it after that
* if you haven't, don't auto-create the DB.

Does that match your thinking?

kart/annotations/__init__.py
craigds: `annotations_session()` creates the db and the table and indexes.

kart/annotations/__init__.py
craigds: the table is created and tables populated before this, as soon as we hit `annotations_session` above.

kart/annotations/__init__.py
craigds: have dealt with that now - it will switch to an in-memory database if the real db can't be created

tests/test_annotations.py
craigds: done

tests/test_annotations.py
craigds: this log message does the same thing though; it identifies that the rows are being _retrieved_, which is the same as testing that they're present (and easier to check)

kart/annotations/__init__.py
craigds: there's a test without an annotations db (actually there already was, but now there's a range of starting states)

kart/annotations/__init__.py
rcoup: How's an in-memory database useful?

kart/annotations/__init__.py
rcoup: That's fine, was just my misunderstanding. Can you add a brief document for this? How it works, when it's created, where it's used, what's stored, etc?

kart/annotations/__init__.py
rcoup: > `# but we still need _some_ sqlalchemy session to yield, otherwise all the annotations code will have to handle this specifically. so we create a in-memory database and make it look readonly [so it isn't used]`

that makes a bit more sense.

All the low-level sqlite error handling seems to be the cache caller's responsibility which I'm not keen on, but at least it's only used in two places at the moment — can be improved later.

kart/annotations/__init__.py
craigds: Added [ANNOTATIONS.md](https://github.com/koordinates/kart/pull/416/commits/6d32e5db29da54a6744646628905837ec4010fbc)

rcoup (approved):

PR #419, opened 2021-04-30
Handle SIGTERM by also SIGTERMing child processes

olsen232:
Otherwise, child processes keep running for no reason.

### DIFF
--- a/kart/__init__.py
+++ b/kart/__init__.py
@@ -8,8 +8,11 @@ __all__ = (
 )
 
 import os
+import logging
 import platform
 import sys
+
+L = logging.getLogger("kart.__init__")
 
 is_frozen = getattr(sys, "frozen", None) and hasattr(sys, "_MEIPASS")
 is_darwin = platform.system() == "Darwin"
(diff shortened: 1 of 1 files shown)


rcoup:
> `# Windows solution is simple - just iterate over children and SIGTERM them - but racy.`

So, `SIGTERM` on Windows maps to the equivalent of `SIGKILL` on posix (calling [`TerminateProcess()`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess)). Notably, there's no handlers/cleanup run.

Looking at [Git's windows compatibility code](https://github.com/git/git/blob/master/compat/mingw.c), it appears to:
1. connect to `SIGINT` (ie. `ctrl+c`), and calling `TerminateProcess()` on children
2. creates child processes using [`DETACHED_PROCESS`](https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags#detached_process) if there is no console (AFAICT this maps to checking `sys.stdout.isatty() is False`). This is more for helpful info, maybe in the cases where we're capturing output (ie. _not_ the `exec*()` cases) we should be following the same approach?

Other TIL:
* `CTRL_C_EVENT`/`SIGINT` and `CTRL_BREAK_EVENT`/`SIGBREAK` don't interrupt `time.sleep()`, so need to loop over `sleep(1)` or something.
* `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` don't seem to work unless you call `kernel32.SetConsoleCtrlHandler(None, False)`, as described in a roundabout fashion in https://stackoverflow.com/questions/7085604/sending-c-to-python-subprocess-objects-on-windows

So... on Windows:

1. connecting to `SIGTERM` doesn't really do anything, since the handler never gets run
2. connecting to `SIGINT` (and `SIGBREAK`?) then passing that to immediate git child processes might be a better approach? And they'll kill their children aggressively (but don't necessarily have to).
3. then anything sending us SIGINT we can pass on SIGINT
4. if we have children processes that _don't_ handle SIGINT then we should call TerminateProcess/kill on them

rcoup:
@olsen232 can you add some more description to the PR about why this is the right solution? ie. what does behaviour look like with/without the `SetConsoleCtrlHandler()` call.

PR #424, opened 2021-05-07
Add support for custom CRS to MySQL WC

olsen232:
This adds better CRS support to MySQL. Previously, you could check out only commits that used built-in EPSG CRS definitions - no custom CRS allowed - and you could not see diffs or commit changes to the CRS (whether you were changing from one EPSG definition to another, or to custom CRS). These are now both supported.

All 4 working copies are modified by this change. The following description tries to explain why:

MySQL is a bit fussier about CRS definitions compared to other working copies. It has the following rules:
- If a table uses CRS x, then CRS x must already exist in the CRS table.
- While a table references CRS x, CRS x must not be modified.
- There are several thousand built-in EPSG CRS's. These must also not be modified.

Therefore, we now apply the following logic to writing CRS's, to avoid breaking the rules:
- Before creating a table that uses CRS x, write CRS x to the CRS table - but only if:
  o CRS x is not a built-in (or doesn't appear to be, eg, isn't an EPSG definition)
  o CRS x is not currently being used by another table
  o Writing CRS x would have an effect (if CRS x already exists as we want it, we specifically don't overwrite it, since doing so would require permissions which we might not have)
- After deleting a table, we don't delete any CRS's. We can never be 100% sure that we wrote the CRS's, so we shouldn't be too eager to clean them up once nobody is using them.
- Suppress any diffs that simply show a difference between a built-in working copy CRS x and a dataset CRS x (with the same ID), with the following reasoning:
a) They probably don't show anything useful - the built-in WC CRS called EPSG:4326 is probably basically the same as the user-supplied CRS called EPSG:4326. They might just have one more digit of accuracy or something like that.
b) If we show a diff, it counts as a dirty working copy, and the user will try and action it - either commit the built-in WC CRS to the dataset, or checkout the dataset CRS to the WC. We don't really want them to do either of these things. We don't want different user's choice of working copies to lead to lots of committing different versions of the same basic CRS over the top of each other, and, we don't want the user to mess with the built-in data that comes with their database (if they even have permission to do so).

Other working copy types weren't previously as strict and so we had laxer versions of the rules.
- if we got a new version of a CRS definition, we would just overwrite the old one (without checking if another table is still using the old version, and, without checking if overwriting if would actually have any effect).
- PostGIS did avoid overwriting built-in definitions, but GPKG did not - there is really only one built-in, EPSG 4326, and we were overwriting it with the user-supplied the CRS.

GPKG has a slightly different variant of these rules - since we control the entire working copy, in GPKG we are more pro-active in tidying up CRS's that are not being used - we can always recreate them when we do need them. 
SQL Server doesn't have the ability to create new CRS definitions so these rules aren't a thing there.

I've also added tests for custom CRS for all four working copy types, which were lacking - specifically that
- you can check out a revision that has custom CRS, and not see any spurious diffs (ie, it roundtrips through the WC)
- or a revision that has standard (built-in) CRS, and not see any spurious diffs
- but if you check out the revision with one CRS and modify it to be a different CRS, then you can see that in the diff
(except for SQL Server, which can't store custom CRS and can't store CRS IDs except by embedding them per-field).

More new MySQL things:
MySQL demands that the CRS contains an explicit axis order.
MySQL won't accept a CRS that contains newlines. (This doesn't cause any problems for us since we normalise whitespace before diffing anyway, so we can change the formatting without any side effects.)


craigds (approved):
👍

PR #430, opened 2021-05-23
Switch to sqlalchemy import source for GPKG.

olsen232:
## Description

No longer using OGR at all for GPKG imports - now only using sqlalchemy.

Kind of a proof-of-concept, getting the structure set up right - what we actually want is to use sqlalchemy for PostGIS imports. That would allow us to be more discerning about types than OGR, which would let us distinguish between timestamp-in-UTC and timestamp-without-timezone.

As a bonus, I'll probably get SQL Server and MySQL imports working too.

### DIFF
--- a/kart/sqlalchemy/gpkg.py
+++ b/kart/sqlalchemy/gpkg.py
@@ -35,32 +35,30 @@ class Db_GPKG(BaseDb):
         return SQLiteIdentifierPreparer(engine.dialect)
 
     @classmethod
-    def list_tables(cls, sess, db_schema=None, with_titles=False):
+    def list_tables(cls, sess, db_schema=None):
         if db_schema is not None:
             raise RuntimeError("GPKG files don't have a db_schema")
 
-        if with_titles:
-            gpkg_contents_exists = sess.scalar(
-                "SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='gpkg_contents';",
+        gpkg_contents_exists = sess.scalar(
+            "SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='gpkg_contents';",
+        )
+        if gpkg_contents_exists:
+            r = sess.execute(
+                """
+                SELECT SM.name, GC.identifier FROM sqlite_master SM
+                LEFT OUTER JOIN gpkg_contents GC ON GC.table_name = SM.name
+                WHERE SM.type='table'
+                AND SM.name NOT LIKE 'sqlite%' AND SM.name NOT LIKE 'gpkg%' and SM.name NOT LIKE 'rtree%'
+                ORDER BY SM.name;
+                """
             )
-            if gpkg_contents_exists:
-                r = sess.execute(
-                    """
-                    SELECT SM.name, GC.identifier FROM sqlite_master SM
-                    LEFT OUTER JOIN gpkg_contents GC ON GC.table_name = SM.name
-                    WHERE SM.type='table'
-                    AND SM.name NOT LIKE 'sqlite%' AND SM.name NOT LIKE 'gpkg%' and SM.name NOT LIKE 'rtree%';
-                    """
-                )
-                return {row['name']: row['identifier'] for row in r}
+            return {row['name']: row['identifier'] for row in r}
 
         r = sess.execute(
             """
             SELECT name FROM sqlite_master SM WHERE type='table'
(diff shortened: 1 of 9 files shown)


kart/sqlalchemy/gpkg.py
```diff
@@ -27,6 +25,42 @@ def _on_connect(pysqlite_conn, connection_record):
...
+                )
+                return {row['name']: row['identifier'] for row in r}
+
+        r = sess.execute(
+            """
+            SELECT name FROM sqlite_master SM WHERE type='table'
+            AND name NOT LIKE 'sqlite%' AND name NOT LIKE 'gpkg%' and name NOT LIKE 'rtree%';
+            """
+        )
+        if with_titles:
```
craigds, line 63: IMO just always use `with_titles` behaviour and ditch the flag. Returning a dict _sometimes_ is confusing

kart/sqlalchemy_import_source.py
```diff
@@ -0,0 +1,242 @@
...
+        path_length = db_type.path_length(spec)
+        longest_allowed_path_length = db_type.path_length_for_table
+        shortest_allowed_path_length = max(
+            db_type.path_length_for_table_container - 1, 0
+        )
+
+        if (
+            path_length > longest_allowed_path_length
+            or path_length < shortest_allowed_path_length
+        ):
```
craigds, line 49:
```suggestion
        if not (
            shortest_allowed_path_length <= path_length <= longest_allowed_path_length
        ):
```

kart/sqlalchemy_import_source.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __init__(
+        self,
+        original_spec,
+        db_type,
+        engine,
+        db_schema,
+        table,
+        **meta_overrides,
+    ):
```
craigds, line 87: with this many arguments, most/all should probably be keyword-only
```suggestion
    def __init__(
        self,
        original_spec,
        *,
        db_type,
        engine,
        db_schema,
        table,
        **meta_overrides,
    ):
```
(note: also requires changes to callers)

kart/sqlalchemy/gpkg.py
```diff
@@ -27,6 +25,42 @@ def _on_connect(pysqlite_conn, connection_record):
...
+                "SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='gpkg_contents';",
+            )
+            if gpkg_contents_exists:
+                r = sess.execute(
+                    """
+                    SELECT SM.name, GC.identifier FROM sqlite_master SM
+                    LEFT OUTER JOIN gpkg_contents GC ON GC.table_name = SM.name
+                    WHERE SM.type='table'
+                    AND SM.name NOT LIKE 'sqlite%' AND SM.name NOT LIKE 'gpkg%' and SM.name NOT LIKE 'rtree%';
+                    """
```
craigds, line 53:
```suggestion
                    ORDER BY SM.name
                    """
```

these flow through to the table chooser prompt, so makes sense to get them in a pretty order

craigds (approved):
Do we rely on any of the special spatial-index things that OGR does? https://gdal.org/drivers/vector/gpkg.html

e.g. it adds some extra SQL functions (`CreateSpatialIndex` etc)

kart/sqlalchemy/gpkg.py
olsen232: Done

kart/sqlalchemy_import_source.py
olsen232: Ooh I think this is the first language I've used that can do this

kart/sqlalchemy_import_source.py
olsen232: Done

kart/sqlalchemy/gpkg.py
olsen232: Done

olsen232:
Lots of reasons why we don't need OGR's CreateSpatialIndex and other spatial SQL functions -
- this only affects how read things to import them - doesn't affect how we create working copies. And, we weren't creating any spatial indexes while reading from an import source.
- We were already using sqlalchemy to actually read the features during GPKG import, in the SQLAlchemyOgrImportSource - so we can't have been using any extra spatial SQL functions there anyway.
- Tests are passing so we're successfully importing things.

PR #431, opened 2021-05-26
Switch PostGIS import to use sqlalchemy

olsen232:
Uses the same code from working-copy adapters to import PostGIS tables into Kart.

Allows us to be more discerning about the difference between different PostGIS types, which is needed for mirroring.

### DIFF
--- a/kart/sqlalchemy_import_source.py
+++ b/kart/sqlalchemy_import_source.py
@@ -4,6 +4,7 @@ import sys
 
 import click
 import sqlalchemy
+from sqlalchemy.orm import sessionmaker
 
 from .exceptions import (
     NotFound,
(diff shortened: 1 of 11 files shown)


kart/sqlalchemy_import_source.py
```diff
@@ -205,21 +212,41 @@ def get_meta_item(self, name):
...
+        with sessionmaker(bind=self.engine)() as sess:
             return dict(
                 self.db_type.adapter.all_v2_meta_items(
-                    conn, self.db_schema, self.table, id_salt
+                    sess, self.db_schema, self.table, id_salt
                 )
             )
 
+    def override_primary_key(self, new_primary_key):
+        """Modify the schema such that the column named 'new_primary_key' is only the primary key column."""
```
craigds, line 223:
```suggestion
        """Modify the schema such that the given column is the primary key"""
```

craigds (approved):

Issue #434, opened 2021-06-01
init/import: problems when importing from PG, error in ORG type LineString and others schemas than public


diogooliveira-dsg:
I'm having some problems testing kart in Postgres/PostGIS.

First i tried to do the init --import command, but i recivied the "OGR type=StringList" error:

> kart init --import postgresql://user:password@localhost/db_master
> Initialized empty Git repository in /home/user/repo_db_master/.kart/
> Error: Unsupported column type for import: OGR type=StringList

So I decided to init without any parameter and then do the import, but now seems that the import does not have the ability to handle more schemas then public:

> kart import postgresql://user:password@localhost/db_master schema1.table1
> Starting git-fast-import...
> Importing 4 features from postgresql://user:password@localhost/db_master:schema1.table1 to schema1.table1/ ...
> Traceback (most recent call last):
>   File "site-packages/sqlalchemy/engine/base.py", line 1706, in _execute_context
>   File "site-packages/sqlalchemy/engine/default.py", line 681, in do_execute
> psycopg2.errors.UndefinedTable: relation "schema1.table1" does not exist
> LINE 1: ...a2a1ad90_1" CURSOR WITHOUT HOLD FOR SELECT * FROM "schema1.tabl...
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "kart_cli.py", line 4, in <module>
>   File "site-packages/kart-0.9.0-py3.7.egg/sno/cli.py", line 333, in entrypoint
>   File "site-packages/click/core.py", line 829, in _call_
>   File "site-packages/click/core.py", line 782, in main
>   File "site-packages/kart-0.9.0-py3.7.egg/sno/cli.py", line 157, in invoke
>   File "site-packages/click/core.py", line 1259, in invoke
>   File "site-packages/click/core.py", line 1066, in invoke
>   File "site-packages/click/core.py", line 610, in invoke
>   File "site-packages/click/decorators.py", line 21, in new_func
>   File "site-packages/kart-0.9.0-py3.7.egg/sno/init.py", line 342, in import_
>   File "site-packages/kart-0.9.0-py3.7.egg/sno/fast_import.py", line 393, in fast_import_tables
>   File "site-packages/kart-0.9.0-py3.7.egg/sno/dataset2.py", line 358, in import_iter_feature_blobs
>   File "site-packages/kart-0.9.0-py3.7.egg/sno/ogr_import_source.py", line 656, in features
[...]

PR #439, opened 2021-06-04
Postgres imports: Hide postgis internal views

craigds:
## Description

Hides these views/tables from `kart import '<postgres-url>'`:

```
 extname | extschema |      relname      | relkind
---------+-----------+-------------------+---------
 postgis | public    | spatial_ref_sys   | r
 postgis | public    | geography_columns | v
 postgis | public    | geometry_columns  | v
 postgis | public    | raster_columns    | v
 postgis | public    | raster_overviews  | v
```

These aren't importable things and they just cause potential problems,
especially when used with `--all-tables` (but they are noisy when
listing/prompting for tables also)

## Related links:

re #434

## Checklist:

### DIFF
--- a/tests/test_import_source.py
+++ b/tests/test_import_source.py
@@ -2,6 +2,8 @@ import os
 
 from osgeo import gdal
 
+import pytest
+
 from kart.ogr_import_source import postgres_url_to_ogr_conn_str
 from kart.repo import KartRepo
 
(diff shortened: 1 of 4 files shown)


tests/test_import_source.py
```diff
@@ -115,3 +128,21 @@ def test_import_various_field_types(tmp_path, postgis_db, cli_runner):
         "reel": {"dataType": "float", "size": 64},
         "tumeric": {"dataType": "float", "size": 64},
     }
+
+
+def test_list_postgres_tables(postgis_db, postgres_table_with_types, cli_runner):
+    # Using postgres here because it has the best type preservation
```
olsen232, line 134: "# Using postgres here because it has the best type preservation"
That's not why you're using postgres here

olsen232 (approved):

tests/test_import_source.py
craigds: truth

PR #440, opened 2021-06-09
Add documentation for datasets V3

olsen232:
## Description

New dataset path layout is not backwards compatible with Kart 0.9, so it needs its own repository structure version number. It will be Datasets V3.

The path layout we used was not previously particularly documented, so I've had to document both the old way and the new way(s) so that I can explain what the difference between v2 and v3 is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Kart v0.10.0 introduces a new repository structure, which is the default, dubbed
 
  * Entire repositories can be upgraded from V2 to V3 using `kart upgrade EXISTING_REPO NEW_REPO`.
  * Anything which works in a V2 repo should work in a V3 repo and vice versa.
- * V3 repos are much more performant when dealing with extremely large datasets.
+ * V3 repos are more performant for large datasets - compared to V2 repos where size-on-disk climbs quickly once dataset size exceeds 16 million features.
 
 ### Other major changes in this release
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,22 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.10.0 (UNRELEASED)
 
-### Major changes in this release
+Kart v0.10.0 introduces a new repository structure, which is the default, dubbed 'Datasets V3'. Datasets V2 continues to be supported, but all newly created repos are V3 going forward.
+
+### Datasets V3
+
+ * Entire repositories can be upgraded from V2 to V3 using `kart upgrade EXISTING_REPO NEW_REPO`.
+ * Anything which works in a V2 repo should work in a V3 repo and vice versa.
+ * V3 repos are much more performant when dealing with extremely large datasets.
+
+### Other major changes in this release
+
 * The working copy can now be a MySQL database (previously only GPKG, PostGIS and SQL Server working copies were supported). The commands `init`, `clone` and `create-workingcopy` now all accept working copy paths in the form `mysql://HOST/DBNAME` [#399](https://github.com/koordinates/kart/pull/399)
   - Read the documentation at [docs/MYSQL_WC.md](docs/MYSQL_WC.md)
 * Import of tables using `kart import` is now supported from any type of database that Kart also supports writing to as a working copy - namely, GPKG, PostGIS, SQL Server and MySQL.
+* Support for rapidly calculating or estimating feature-counts - see below.
 
-### Other changes
+### Other minor changes
 * Change to `kart data ls` JSON output, now includes whether repo is Kart or Sno branded.
 * Importing from a datasource now samples the first geometry to check the number of dimensions - in case the datasource actually has 3 or 4 dimensions but this fact is not stored in the column metadata (which is not necessarily required by all source types). [#337](https://github.com/koordinates/kart/issues/337)
 * Bugfix: Creating a working copy while switching branch now creates a working copy with the post-switch branch checked out, not the pre-switch branch.
--- a/docs/DATASETS_v3.md
+++ b/docs/DATASETS_v3.md
@@ -1,22 +1,32 @@
-Datasets V2
+Datasets V3
 -----------
 
 ### Background
 
-Kart was previously known as Sno. Sno 0.2 introduced Datasets V1, and Sno 0.5 introduced Datasets V2.
-Datasets V1 and V2 are storage formats for database tables where each table row is stored in a separate file.
-This means table rows can be stored using git-style version control, resulting in a storage format for a database-table which has version history.
+Kart's internal repository structure - currently called "Datasets V3" - has been through a few iterations:
 
-Datasets V2 is very similar to Datasets V1 - the main difference is that the schema of a Datasets V2 table can be changed in isolation without having
-to rewrite every row in the table. Rows that were written with a previous schema are adapted to fit the current schema when read.
+Kart Version | Kart's name at the time | Repository structure version
+--- | --- | ---
+0.0 to 0.1 | Snowflake | Datasets V0
+0.2 to 0.4 | Sno | Datasets V1
+0.5 to 0.8 | Sno | Datasets V2
+0.9 | Kart | Datasets V2
+0.10 | Kart | Datasets V3 (but v2 still supported)
 
-In Sno 0.4, only Datasets V1 is supported. In Sno 0.5, both Datasets V1 and Datasets V2 are supported, but a particular Sno repository must be entirely one or the other. From Sno/Kart 0.8 onwards, only Datasets V2 is supported - V1 and earlier must be upgraded first using `kart upgrade SOURCE DEST`.
+Datasets V3 is a storage formats for database tables where each table row is stored in a separate file. This means table rows can be stored using git-style version control, resulting in a storage format for a database-table which has version history.
 
-The following is a technical description of Datasets V2.
+The main improvement of Datasets V3 is how the rows are divided into different folders (or "trees" in git terminology) for more efficient storage when there are a large number of revisions and features. See [DATASETS_v2](DATASETS_v2.md) for more information on the preivous system.
+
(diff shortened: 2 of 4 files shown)


docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+
+`urlsafe_b64encode(msgpack.packb(primary_key_value_array))`
+
+In the example feature path above, there is only one primary key column, and the feature being stored is the feature with primary key 77. So the primary key values are an array of length one containing 77: `[77]`. So the filename was generated as follows:
+
+`[77]` -> MessagePack -> `bytes([0x91, 0x4d])` -> Base64 -> `kU0=`
+
+#### Path to the feature file
+
+For technical reasons, it is best if only a relatively small number of features are stored together in a single directory, and similarly if only a small number of directories are stored together in a single directory. Ideally, the features created at the same time or likely to be edited at the same time should be stored together, rather than spread out among all the other features - so, neighbouring primary key values should be neighbouring files where possible.
```
rcoup, line 310:
```suggestion
For technical reasons, it is best if only a relatively small number of features are stored together in a single directory, and similarly if only a small number of directories are stored together in a single directory. Ideally, the features created at the same time or likely to be edited at the same time should be stored together, rather than spread out among all the other features - so, neighbouring primary key values should be neighbouring file paths where possible.
```

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+  "levels": 4,
+  "encoding": "base64"
+}
+```
+
+The `"scheme": "int"` tells us that this path-structure is used for a dataset which has a single primary key column of type integer, and that value will be used directly to generate the path to the file. (The only other supported scheme is `"msgpack/hash"` - see below).
+
+The next two parameters - `"branches": 64, "levels": 4` indicate that there are 4 levels of directory hierarchy, and at each level, there are up to 64 different directories branching out, such that a dataset with a huge number of features will have them spread across `64 ** 4 = 16777216` leaf-node directories - so a dataset could have `64 ** 5 = 1073741824` features and no directory would contain more than 64 directories or features. (Directories are only created when needed, so a dataset with only one feature with primary key 1 would create only four nested folders in which to store it, eg `A/A/A/A`.)
+
+Each directory is named after a character in the base64 alphabet - this is the `"encoding": "base64"`, and this encoding only supports a branch factor of 64. The other valid encoding is `"hex"`, which supports a branch factor of 16 or 256.
```
rcoup, line 327:
```suggestion
Each directory is named after a character in the [URL-safe Base64 alphabet](https://en.wikipedia.org/wiki/Base64#The_URL_applications) - this is the `"encoding": "base64"`, and this encoding only supports a branch factor of 64. The other valid encoding is `"hex"`, which supports a branch factor of 16 or 256.
```

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+
+So to encode the example before where the primary-key-value-array is `[77]` - since the scheme is "int" we know there is only one primary key value, an integer, which we can use as input for the subsequent steps: `77`. Encoding an integer (rather than a string of bytes) using Base64 works similarly to encoding integers in other bases such as hexadecimal. A quick primer: 0 is `A`, 1 is `B`, 64 is `BA`, and 77 is `BN`. We pad the left side with `A` (which stands for `0`) as needed: `AAABN`, and we remove the last character since we want to only change the path every 64 features, not every feature, giving us `AAAB`. (Feature filenames already have their own scheme which distiguishes them from every other feature in the same folder). Treating this as a path 4 levels long gives us `A/A/A/B`.
+
+So, feature with primary key values `[77]` would be written at `A/A/A/B/kU0=` using this path-structure.
```
rcoup, line 331: worth an example where we're encoding a much larger number to illustrate how we max out at 64^5 dirs and we use the LSBs to allocate to trees?

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+
+`bytes([0x3c, 0x57, 0x8e, 0x75, ...])` -> Base64 encode first 24 bits -> `PFeO` -> treat as path -> `P/F/e/O`
+
+So, feature with primary key values `[77]` would be written at `P/F/e/O/kU0=` using this path-structure.
+
+The paths to the files are more opaque in this scheme and provide less information about the feature's primary keys - however, just as in the last scheme, the feature's filename by itself can be decoded back into the primary key values. The paths are simply there to spread out the features for performance reasons.
+
+#### Legacy path-structure
+
+Datasets V2 only supports a single path structure, which is not stored in the dataset, but hard-coded. If no path-structure information is stored in the dataset, then the Datasets V2 structure is assumed. See [DATASETS_v2](DATASETS_v2.md)
```
rcoup, line 355:
```suggestion
Datasets V2 only supports a single path structure, which is not stored in the dataset, but hard-coded. If no path-structure information is stored in the dataset, then the Datasets V2 structure is assumed. See [DATASETS_v2](DATASETS_v2.md). This is the same structure as `msgpack/hash`.
```

rcoup (approved):
👍

CHANGELOG.md
```diff
@@ -7,12 +7,22 @@ _When adding new entries to the changelog, please include issue/PR numbers where
...
 ## 0.10.0 (UNRELEASED)
 
-### Major changes in this release
+Kart v0.10.0 introduces a new repository structure, which is the default, dubbed 'Datasets V3'. Datasets V2 continues to be supported, but all newly created repos are V3 going forward.
+
+### Datasets V3
+
+ * Entire repositories can be upgraded from V2 to V3 using `kart upgrade EXISTING_REPO NEW_REPO`.
+ * Anything which works in a V2 repo should work in a V3 repo and vice versa.
+ * V3 repos are much more performant when dealing with extremely large datasets.
```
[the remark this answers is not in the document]
olsen232: Done

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
+Datasets V3
+-----------
+
+### Background
+
+Kart's internal repository structure - currently called "Datasets V3" - has been through a few iterations:
+
+Kart Version | Kart's name at the time | Repository structure version
+--- | --- | ---
+0.0 to 0.1 | Snowflake | Datasets V0
```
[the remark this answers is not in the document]
olsen232: Done

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+0.10 | Kart | Datasets V3 (but v2 still supported)
+
+Datasets V3 is a storage formats for database tables where each table row is stored in a separate file. This means table rows can be stored using git-style version control, resulting in a storage format for a database-table which has version history.
+
+The main improvement of Datasets V3 is how the rows are divided into different folders (or "trees" in git terminology) for more efficient storage when there are a large number of revisions and features. See [DATASETS_v2](DATASETS_v2.md) for more information on the preivous system.
+
+The main improvement of Datasets V2 is that the schema of a table can be changed in isolation without having to rewrite every row in the table. Rows that were written with a previous schema are adapted to fit the current schema when read. See [DATASETS_v1](DATASETS_v1.md) for more information on the previous system.
+
+To upgrade a Kart repository to the latest supported repository structure, run
+`kart upgrade SOURCE DEST` where `SOURCE` is the path to the existing repo, and `DEST` is the path to where the upgraded repo will be created.
```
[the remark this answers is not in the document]
olsen232: Done

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+This is the type of data which is stored in this column. A complete list of allowed types is found in the "Data types" section below.
+
+###### `primaryKeyIndex`
+This controls whether or not this column is a primary key. If this value is `null` or not present, then the column is not a primary key. If it is any non-negative integer, then the column is a primary key. The first primary key column should have a `primaryKeyIndex` of `0`, the next primary key column should have `1`, and so on.
+
+Those are all of the fields that apply to any column. Certain dataTypes can have extra fields that help specify the type of data that the column should hold - see the "Extra type info" section below.
+
+##### Data types
+
+The following data types are supported by Kart. When a versioned Kart dataset is converted to a database table (ie, when `kart checkout` updates the working copy) then these Kart data types will be converted to equivalent data types in the database table, depending on what is supported by the database in question.
```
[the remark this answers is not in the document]
olsen232: Done

docs/DATASETS_v3.md
```diff
@@ -0,0 +1,355 @@
...
+1. Geometries must use the StandardGeoPackageBinary type.
+2. GeoPackage binary headers must always use little-endian byte ordering.
+3. The WKB geometry must always use little-endian byte ordering.
+4. All non-empty geometries must have an envelope, except for POINT types:
+   - Points and empty geometries have no envelope.
+   - Geometries with a Z component have an XYZ envelope.
+   - Other geometries have an XY envelope.
+5. The `srs_id` is always 0, since this information not stored in the geometry object but is stored on a per-column basis in `meta/schema.json` in the `geometryCRS` field.
+
+**Note on axis-ordering:** As required by the GeoPackageBinary format, which Kart uses internally for geometry storage, Kart's axis-ordering is always `(longitude, latitude)`. This same axis-ordering is also used in Kart's JSON and GeoJSON output.
```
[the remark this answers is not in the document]
olsen232: Done

docs/DATASETS_v3.md
olsen232: Done

docs/DATASETS_v3.md
olsen232: Done, although this is incomplete - the datasets V2 structure is not just msgpack/hash, but all of the following:

```json
{
  "scheme": "msgpack/hash",
  "branches": 256,
  "levels": 2,
  "encoding": "hex"
}
```

rcoup: Ah, so maybe something more like this would be better? 

> This is the `msgpack/hash` scheme with differing parameters — see DATASETS_v2 for details.

docs/DATASETS_v3.md
olsen232: Done

docs/DATASETS_v3.md
olsen232: Done

PR #447, opened 2021-06-18
Optimisations for upgrading mirrors

olsen232:
Upgrading a repo from V2 -> V3 is already supported, but this change adds the following optimisations:

- upgrading in place is now supported, and for V2 -> V3, this means the feature blobs do not have to be reserialized since they remain unchanged.
- commits which only affect one dataset are now processed using the existing `replace_ids` argument of fast-import - so, the next-upgraded-commit starts from the previous upgrade commit and fast-import commands are issued to modify only certain features, as opposed to previously when each source commit would be processed in isolation and the upgraded commit would be built from scratch.


craigds (approved):
LGTM: 👍

Issue #448, opened 2021-06-24
upgrade loses extra data in V2 -> V3 upgrade


olsen232:
`kart upgrade` works by treating every dataset at every commit in the V2 repo as a datasource to be imported, presenting the same interface as it would if you imported from a GPKG or a shape file. 

This is good from a reuse of code perspective and is likely to work fine for our users data. However, our own repos have extra data that is not generally importable from GPKG or similar, but which we need to preserve and is not currently being preserved. Specifically -
- attachments (eg license.txt, note this is not yet fully supported by kart) 
- extra kart-specific metadata (`generated-pks.json`, which contains information to help us recognise new versions of existing features from pk-less datasources where we can't use the PK to do so).

Within the kart dataset tree, these blobs are found at the following places:
```
<dataset-path>/
    LICENSE.txt <-- attachment not be preserved by kart upgrade
    .sno-table/
        meta/
            title
            description
            schema.json
            generated-pks.json <-- extra metadata not preserved by kart upgrade
```

PR #449, opened 2021-06-24
Fix for kart upgrade so it handles complicated repos

olsen232:
Most mirrored repos are complicated - user repos, where they exist at all, are simpler.

A complicated repo might have meta-items that are non-standard eg "generated-pks.json" - these are considered "non-standard" as they don't describe the state of the data itself, or even metadata about the data such as title, description, or schema. Instead, they are extra metadata that the data-source itself maintains so that it can accurately provide the actual data / metadata. As such, they currently don't show up in diffs.

A complicated repo might have attachments, which are currently mostly hidden from users - but still need to be in the mirrors to future proof the mirrors since we can't keep changing the mirrored history, and we will eventually support them.

https://github.com/koordinates/kart/issues/448

### DIFF
--- a/kart/fast_import.py
+++ b/kart/fast_import.py
@@ -42,7 +42,7 @@ class _CommitMissing(Exception):
     pass
 
 
-def _safe_walk_repo(repo):
+def _safe_walk_repo(repo, from_commit):
     """
     Contextmanager. Walk the repo log, yielding each commit.
     If a commit isn't present, raises _CommitMissing.
@@ -50,7 +50,7 @@ def _safe_walk_repo(repo):
     """
     do_raise = False
     try:
-        for commit in repo.walk(repo.head.target):
+        for commit in repo.walk(from_commit.id):
             try:
                 yield commit
             except KeyError:
@@ -65,7 +65,10 @@ def _safe_walk_repo(repo):
 
 
 def should_compare_imported_features_against_old_features(
-    repo, source, replacing_dataset
+    repo,
+    source,
+    replacing_dataset,
+    from_commit,
 ):
     """
     Returns True iff we should compare feature blobs to the previous feature blobs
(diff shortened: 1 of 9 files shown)


kart/fast_import.py
```diff
@@ -149,14 +152,22 @@ def fast_import_clear_trees(*, procs, replace_ids, replacing_dataset, source):
...
     for i, proc in enumerate(procs):
         if replace_ids is None:
             # Delete the existing dataset, before we re-import it.
             proc.stdin.write(f"D {source.dest_path}\n".encode("utf8"))
         else:
-            # delete and reimport meta/
+            # Delete and reimport any attachments at dest_path
+            attachment_names = [
+                obj.name for obj in replacing_dataset.tree if obj.type_str == "blob"
+            ]
```
craigds, line 165: should this look deeper? it only looks for blobs which are immediate children of the dataset tree. I feel like we're going to miss something here if we ever add more complex hierarchies of 'attachments'.

kart/upgrade/__init__.py
```diff
@@ -53,8 +60,13 @@ def get_meta_item(self, name, missing_ok=True):
                     return next(iter(metadata_xml))["text/xml"]
         return result
 
+    def attachment_items(self):
+        attachments = [obj for obj in self.tree if obj.type_str == "blob"]
```
craigds, line 64: (same again here)

craigds (approved):

kart/fast_import.py
olsen232: Right now these are the only attachments which exist in V2 repos, and they really only exist in repos which we created (not user repos) and the set of all V2 repos is not supposed to grow very much (or at least, not with new types of V2 repos which we haven't yet seen).

There is another type of hypothetical attachment which won't currently be preserved and which we can't really do anything about, at least not with the current structure - attachments which are not attached to any particular dataset, but are at some other path. But again, these don't exist either. So, not too motivated to make this more complicated for no real benefit

craigds: okay 👍

PR #452, opened 2021-06-30
Request the logs of multiple commits at once

olsen232:
Generally, git log takes any number of commits-plus-ancestors you want to log - `git log A B C` will log all the commits that are ancestors of any of A, B, or C (or that *are* A, B or C). `kart log` works the same.

The kart app currently makes three log calls in order to find these three lists:
All the commits that are ancestors of A only.
All the commits that are ancestors of B only.
All the commits that are the ancestors of the common ancestor of A and B.

This change adds a hidden flag - `--separate-logs` - so that each commit+ancestors is kept separate from the others. Calling `kart log --separate-logs A B` will now return the following JSON object:
```
{
  "A": [...all the commis that are ancestors of A only...],
  "B": [...all the commits that are ancestors of B only...], 
  "*": [...all the commits that are the ancestors of the common ancestor of A and B...]
}
 ```

Works similarly for more commits eg `kart log --separate-logs A B C`

https://github.com/koordinates/kart/issues/170

## Checklist:

- [ ] Hidden feature, not including in changelog

### DIFF
--- a/kart/log.py
+++ b/kart/log.py
@@ -8,6 +8,8 @@ from .cli_util import tool_environment
 from .exec import execvp
 from .exceptions import SubprocessError
 from .output_util import dump_json_output
+from .repo import KartRepoState
+from .structs import CommitWithReference
 from .timestamps import datetime_to_iso8601_utc, timedelta_to_iso8601_tz
 from . import diff_estimation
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by olsen232** [comment]
@UberMouse

**Comment by UberMouse** [comment]
Not sure how I'm supposed to use this if it doesn't support ranges, since we currently use ranges to resolve the logs?

What's the equivalent to `@..@{upstream}` and `@{upstream}..@` with this?

**Comment by craigds** [comment]
so basically `kart log --separate-logs A B` provides the same information as:

* `kart log A..B`
* `kart log B..A`
* `kart log "$(git merge-base A B)"`

The code looks fine but tbh I don't like the concept. This changes the output from a JSON array into an object and calls `git log` 3 times. The result is the same as if the caller had just called `kart log` 3x with the specific ranges they wanted. The purpose seems to be a workaround for incurring the cost of kart startup overhead (~200ms?) three times...


Additionally I think all the required information is actually present in `kart log A B -o json`:

```bash
$ kart log main other -o json
[
  {
    "commit": "3a11f62c8bc81990738ecbd0e368335b62210590",
    "message": "Update metadata for chatham_islands_lakes_polygons_v14",
    "parents": [
      "2609b38b4ee085bdc7ef5d29ca4262f92eed09bc"
    ],
    "refs": [
      "other"
    ]
  },
  {
    "commit": "381ea37ce732ed4cdd4f300b3e2361c3316865bf",
    "message": "Replaced 211 features with 211 features; schema changed; ...\n\nAlso: Dublin Core metadata added; Title and description changed",
    "parents": [
      "2609b38b4ee085bdc7ef5d29ca4262f92eed09bc"
    ],
    "refs": [
      "main"
    ]
  },
  {
    "commit": "2609b38b4ee085bdc7ef5d29ca4262f92eed09bc",
    "message": "211 features added; description and 1 file added",
    "parents": [],
    "refs": []
  }
]
```

^ Here is a repo with two branches: `other` and `main`. The graph of them looks like

```
other  ->  3a11f62c     2609b38b  <-  main
                  \      /
                  2609b38b4
```

The single log call with both refs listed gives all that's needed to rebuild this graph quite easily
by following `parents`.

I'd prefer we work on reducing the startup overhead more.
In the meantime, the caller could just call `kart log` 3x in parallel rather than in series?

I'm -0. cc @rcoup

olsen232:
Craig has a different plan... so this could all end up moot. But you would just do this:
`kart log --separate-logs @ @{upstream}`
Then you would get this object:
```
{
"@": [... all commits that are ancestors of @ only ...]
 "@{upstream}": [... all commits that are ancestors of @{upstream} only ...]
 "*": [... all commits that are ancestors of the common ancestor of @ and @{upstream} ...]
 }
 ```
By "@ only" I mean, it is not also an ancestor of the other, @{upstream} and is not found in any other list. So, these are actually ranges.

UberMouse:
>In the meantime, the caller could just call kart log 3x in parallel rather than in series?
Already do that

olsen232:
I'm more convinced by Craig's plan than by this PR now actually. There is some complexity that is not handled well by @UberMouse's current code, nor by this PR which works in the same way.

Check out this presentation with an unusual commit graph. https://docs.google.com/presentation/d/1_wkkz7uui6VlG_X250yyWjvfz_n6EBYcvGo-4RNn6xQ/edit
Basically, there can be rare circumstances where there can be a commit that can be an ancestor of A, and an ancestor of B, but not an ancestor of the merge-parent of A and B. In this case, this commit is part of both M..A and M..B where M is the merge-parent, so the same commit could be shown twice in the UI. (Where A and B could be, for instance, @ and @{upstream})

The best solution to this unusual situation is probably to do what @craigds suggests and just run `kart log A B` and parse the results including building a graph (or if not actually building the whole graph, at least finding which commits are reachable from A or B respectively). Handling this unusual situation may be low priority, but it does make me think I shouldn't merge this PR, which makes the inability to handle this situation properly part of our hidden API. And as Craig says, `kart log A B` does actually have all the information you need if you parse the parents, so, this PR isn't really adding a lot.

PR #454, opened 2021-07-08
Use UTF8 as text file encoding on all platforms

craigds:
## Description

The test added in 6bdc2b2a190eb3df3c8a84ce399d64cbb6e80c42 fails on Windows because it seems files are opened using the `charmap` encoding there, unlikely MacOS/Linux which are using utf-8.

We want utf-8 to be the default text file encoding on _all_ platforms, otherwise we'll get inconsistent results (a file written by one user may not be readable by another)

This fix forces the locale encoding to UTF-8 without touching the locale's language, and thus hopefully fixes the test on Windows 🤞 

## Related links:

none

## Checklist:

### DIFF
--- a/kart/__init__.py
+++ b/kart/__init__.py
@@ -8,9 +8,19 @@ __all__ = (
 )
 
 import os
+import locale
 import logging
 import platform
 import sys
+
+# Always use UTF-8 for opening text-mode files.
+# This is the default on POSIX but not on Windows for some reason.
+# But we want consistent results across platforms, so we override it.
+# We *would* set PYTHONUTF8=1 to enable python's UTF-8 mode,
+# but that can't be changed after the interpreter
+# has started up, so it's harder for us to control.
+lang = locale.getlocale()[0]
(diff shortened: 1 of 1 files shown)


craigds:
Foiled again - it seems this doesn't work on windows either. I think enabling utf8 mode might be the only way to properly set the default encoding to utf8 in windows.

craigds:
d95c32dae266a6d85429d481622c51a17a80c21d fixes the specific test failure, but I think a general fix is still in order if we can figure out how.

@rcoup you've done some windows dev, do you have any suggestions on how to fix this?

craigds:
https://pyinstaller.readthedocs.io/en/stable/spec-files.html#giving-run-time-python-options suggests we can pass _some_ options to the python interpreter, but it has a very limited whitelist and `-X utf-8` isn't in it.

I can't see anything in the pyinstaller docs about customizing the _environment_ for the python interpreter unfortunately.

craigds:
foiled by Windows, seems unlikely we (I at least) can resolve this.

PR #458, opened 2021-07-12
Configure env variables for GDAL and PROJ on windows

olsen232:
## Description

Set GDAL and PROJ data folder environment variables at Kart launch Windows. This means Kart working properly no longer depends on the global value of these variables, which other tools may also change.
The proper location of GDAL and PROJ data folder was decided empirically by searching through installed Kart folders, without referring to our changing the makefiles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please note that compatibility for 0.x releases (software or repositories) isn't
 
 _When adding new entries to the changelog, please include issue/PR numbers wherever possible._
 
+## 0.10.3 (UNRELEASED)
+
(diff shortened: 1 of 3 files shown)


rcoup (changes requested):
Looks reasonable. Can we re-enable the test at https://github.com/koordinates/kart/blob/master/tests/test_core.py#L272-L297?

I think it's worth splitting into two:
- one doing config stuff (L272-285), this should work on all platforms
- one doing a CRS transform using the grid(L287-297): this should work on Linux, and on Windows  should return the 7-parameter transform result (which is expected atm).

Issue #456, opened 2021-07-12
Add spatial filtering support to Kart


olsen232:
It should be possible to work with a "spatially filtered" Kart repo. This is a repo which has a particular geometry which is the stored in the config which is treated as a spatial filter, such that -

- when the working copy is written, only features that match the spatial filter are included
- when cloning and fetching features from the server, only features that match (or possibly match) the spatial filter are included
- when generating diffs, changes that happen entirely outside the spatial filter are omitted.

There is some more complexity around these concepts (such as, making sure the user doesn't accidentally reuse the PK from a feature that doesn't match their current filter, since PKs must be globally unique) but that is the basic idea.

olsen232:
Kart spatial filtering `a.intersects(b)` test is naive right now -
- doesn't make sure long polygon edges are segmented into short lines before reprojecting, so the geographical location of the middle of the polygon's edge can change
- doesn't consider the anti-meridians at all - might behave unexpectedly if the spatial filter or the layer crosses the antimeridian
- doesn't check if a spatial filter geometry can be wholly (or partially) reprojected into a target CRS, so this will simply fail if it cannot

Generally works fine in spite of this but these will all be needed at some point

olsen232:
Setting a spatial filter locally during a kart clone or kart init is mostly working - however, there are lots of gaps that need to be revisited. Tracking them here so they don't get forgotten:

Setting and applying spatial filters locally:
- better intersection tests (see above https://github.com/koordinates/kart/issues/456#issuecomment-880258848)
- no supported way to change the spatial filter after the repo after it has been created (workarounds exist eg if you manually edit the config and then recreate the workingcopy)
- make sure refs/filters/* are being cloned as we expect, think about if we support other refs
- update sequence generators in the working copy to try and avoid probably-accidental-PK-collisions
- detect and handle probably-accidental-PK-collisions somehow - maybe warn or maybe reassign PKs

Applying spatial filters to data sent over network - partial clone:
This isn't even started at all, but basically kart needs to support the following:
- Sending the spatial filter or spatial filter reference via git clone
- Detecting when features are missing due to partial clone and generally just skipping those features
- Fetching missing objects when they are needed after all (ie, in a merge conflict, or, to find out what the first unassigned PK is)

olsen232:
As of #475, spatial-index changes to kart have begun - these depend on S2 / pywraps2, and on git 2.33 or later.
[...]

PR #460, opened 2021-07-16
Spatial filtering for working copies #2 - reset()

olsen232:
## Description

Spatial filtering (which is a not-yet-exposed kart config option) previously only worked when write_full was called - that is, when the entire dataset is rewritten to the working copy scratch. This PR makes it also work when the dataset is modified in the working copy without rewriting the entire dataset. This can happen in response to a user command like `kart checkout`, `kart restore`, or `kart reset`, depending on how different the two states are that the user is moving between.

All the TODO bullet points of the previous spatial filter PR are still required for spatial filters - this PR doesn't address them.
See https://github.com/koordinates/kart/pull/457

Tracking issue:

### DIFF
--- a/kart/spatial_filters.py
+++ b/kart/spatial_filters.py
@@ -30,18 +30,17 @@ class SpatialFilter:
     def from_spec(cls, geometry_wkt, crs_spec):
         if not geometry_wkt:
             return SpatialFilter.MATCH_ALL
+        if not crs_spec:
+            raise ValueError("SpatialFilter requires a CRS")
 
         # TODO - use PreparedGeometry.
         filter_geometry_ogr = Geometry.from_wkt(geometry_wkt).to_ogr()
-        if crs_spec is not None:
-            try:
-                crs = make_crs(crs_spec)
-            except RuntimeError as e:
-                raise click.BadParameter(
-                    f"Invalid or unknown coordinate reference system configured in spatial filter: {crs_spec!r} ({e})"
-                )
-        else:
-            crs = None
+        try:
+            crs = make_crs(crs_spec)
+        except RuntimeError as e:
+            raise click.BadParameter(
+                f"Invalid or unknown coordinate reference system configured in spatial filter: {crs_spec!r} ({e})"
+            )
 
         return SpatialFilter(filter_geometry_ogr, crs)
 
@@ -54,18 +53,18 @@ class SpatialFilter:
         crs - The CRS used to interpret the spatial filter. An OGR SpatialReference object.
         match_all - if True, this filter is the default match-everything filter.
         """
+        self.match_all = match_all
+
         if match_all:
-            self.filter_ogr = None
-            self.filter_env = None
+            self.filter_ogr = self.filter_env = self.crs = None
+            self.geom_column_name = None
         else:
             self.filter_ogr = filter_geometry_ogr
             self.filter_env = self.filter_ogr.GetEnvelope()
+            self.crs = crs
+            self.geom_column_name = geom_column_name
 
-        self.crs = crs
-        self.geom_column_name = geom_column_name
-        self.match_all = match_all
-
-    def matches(self, feature_or_geometry):
+    def matches(self, feature):
         """
         Returns True if the given feature geometry matches this spatial filter.
         The feature to be tested is assumed to be using the same CRS as this spatial filter,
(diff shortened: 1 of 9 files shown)


kart/working_copy/base.py
```diff
@@ -828,9 +830,10 @@ def _insert_or_replace_state_table_tree(self, sess, tree_id):
         )
         return r.rowcount
 
-    def write_full(self, commit, *datasets, spatial_filter=SpatialFilter.MATCH_ALL):
+    def write_full(self, commit, *datasets):
         """
-        Writes a full layer into a working-copy table
+        Writes a full layer into a working-copy table.
+        Only writes features that match the given spatial filter (defaults to the repo spatial filter.)
```
craigds, line 836:
```suggestion
        Only writes features that match the repo's spatial filter
```

kart/spatial_filters.py
```diff
@@ -37,34 +45,46 @@ def from_spec(cls, geometry_wkt, crs_spec):
...
+        feature_geometry - either a feature dict (in which case self.geom_column_name must be set)
+            or a geometry.Geometry object.
         """
-        if self.match_all or self.filter_ogr is None or feature_geometry is None:
+        if self.match_all or feature_or_geometry is None:
             return True
 
+        if isinstance(feature_or_geometry, dict):
+            feature_geometry = feature_or_geometry[self.geom_column_name]
+        elif isinstance(feature_or_geometry, Geometry):
```
craigds, line 83: is this used? the two calls I see only pass features in. In which case this function might as well only take features

kart/spatial_filters.py
```diff
@@ -105,26 +125,41 @@ def matches(self, feature_geometry):
...
+            if not ds_crs_defs:
+                return self
+            if len(ds_crs_defs) > 1:
+                raise NotYetImplemented(
+                    f"Sorry, spatial filtering dataset {ds_path!r} with multiple CRS is not yet supported"
+                )
+            ds_crs_def = list(ds_crs_defs.values())[0]
+            return result.transform_for_crs(ds_crs_def, ds_path)
+        else:
+            return result
```
craigds, line 148: what does it mean for a spatial filter to have a null CRS? we just assume it's the same as the datasets? Maybe it'd be better to force it to be explicit, or possibly assume it's WGS84

kart/spatial_filters.py
```diff
@@ -105,26 +125,41 @@ def matches(self, feature_geometry):
 
     def transform_for_dataset(self, dataset):
         """Transform this spatial filter so that it matches the CRS of the given dataset."""
-        if self.match_all or self.filter_ogr is None or self.crs is None:
-            return self
+        if self.match_all or dataset is None:
```
craigds, line 128: needs a comment why `dataset=None` is sensible

kart/spatial_filters.py
rcoup: I don't think a null CRS is really a good idea: Server-side, we need to get the filter to WGS84 in order to do the clone filtering, and to do that conversion the source needs to be known.

kart/spatial_filters.py
olsen232: Good point, this no longer receives mere geometries. Removed this capability.

kart/working_copy/base.py
olsen232: Done

kart/spatial_filters.py
olsen232: Default CRS is now EPSG:4326, if it is not specified in the config.

kart/spatial_filters.py
olsen232: Probably is not sensible - I thought needed it for doing diff(old_dataset, new_dataset, spatial_filter), which kind of turns into diff(old_dataset + old_spatial_filter, new_dataset + new_spatial_filter) - but old_dataset or new_dataset could be None, in which case, there's no particular spatial-filter that applies to them except MATCH_ALL (or MATCH_NONE would work too).
But I didn't end up structuring the code that way so never mind!

tests/test_spatial_filters.py
```diff
@@ -10,6 +11,20 @@ def ring_as_wkt(*points):
     return "(" + ",".join(f"{x} {y}" for x, y in points) + ")"
 
 
+def bbox_as_wkt_polygon(min_x, max_x, min_y, max_y):
```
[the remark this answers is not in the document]
olsen232: Not done - this particular piece of code is a test helper that is only used in this file, so, as long as it is a useful test helper, it can have its arguments in any reasonable order. Right now none of the tests test anti-meridian crossing spatial filters so it doesn't make any difference.
The actual production code does need to actually deal with CRS's that cross the anti-meridian, at some point, but this is still on the list of TODOs. When it does, I think we will still have some (min-x, max-x, min-y, max-y) though, since that is the official GPKG spec, and we store our geometries according to that spec. In any case, not dealt with in this PR.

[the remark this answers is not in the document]
rcoup: > some (min-x, max-x, min-y, max-y) though, since that is the official GPKG spec, and we store our geometries according to that spec.

Potentially makes sense for a single geometry in some situations, falls apart for anything else. Add a comment that a CRS-aware bounds method should be used instead of this unless absolutely necessary.

PR #461, opened 2021-07-20
Filter out diff deltas that don't match the spatial filter.

olsen232:
## Description

Feature deltas are only relevant to the user if either the old or the new version of the feature is inside the spatial filter. The exception to this rule is local working-copy edits, which are always relevant to the user. This PR filters out irrelevant deltas that do not match the spatial filter.

## Related links:

### DIFF
--- a/kart/base_diff_writer.py
+++ b/kart/base_diff_writer.py
@@ -11,10 +11,10 @@ from .exceptions import (
     CrsError,
     InvalidOperation,
     NotFound,
-    NotYetImplemented,
     NO_WORKING_COPY,
 )
 from .key_filters import RepoKeyFilter
+from .spatial_filters import SpatialFilter
 
 
 L = logging.getLogger("kart.diff_writer")
@@ -80,6 +80,8 @@ class BaseDiffWriter:
 
         self.user_key_filters = user_key_filters
         self.repo_key_filter = RepoKeyFilter.build_from_user_patterns(user_key_filters)
+
+        self.spatial_filter = repo.spatial_filter
 
         base_ds_paths = {ds.path for ds in self.base_rs.datasets}
         target_ds_paths = {ds.path for ds in self.target_rs.datasets}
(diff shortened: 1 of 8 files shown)


rcoup (approved):
> Spatial filtered diff output is not yet properly streamed - every feature is loaded before the first feature is output. (In contrast to diffs where there is no spatial filter, where each feature is loaded just in time). This should be fixed at the very least for json-lines output, which is where streaming is most beneficial.

Is there a reason that yielding matching features one at a time isn't straightforward? Or simply to get it going?

olsen232:
> > Spatial filtered diff output is not yet properly streamed - every feature is loaded before the first feature is output. (In contrast to diffs where there is no spatial filter, where each feature is loaded just in time). This should be fixed at the very least for json-lines output, which is where streaming is most beneficial.
> 
> Is there a reason that yielding matching features one at a time isn't straightforward? Or simply to get it going?

I've done it now, so, it's not impossible - it just took a bit of restructuring. As the code was, filtering feature deltas immediately before they were output was not possible, since at that point, there was no way to tell which deltas were from commits and which were from the WC (and the two sources are filtered differently). However, filtering feature deltas at the point where we did have this information was too soon - this information was lost before the first feature was output, so if we do spatial filtering at that point, it means loading every feature before outputting the first one.

The fix is just to propagate which features are from commits and which features are from the WC.

Also added a warning to `kart create-patch` so that TODO is done now too - we can always revisit this if we decide `kart create-patch` should not have be affected by spatial filtering at all.

kart/base_diff_writer.py
```diff
@@ -237,54 +245,63 @@ def get_dataset_diff(self, dataset_path, apply_spatial_filter=True, prune=True):
...
-        Removes all the feature-deltas from the given dataset diff that don't match the self.spatial_filter.
-        Features that are present in the working-copy diff for this dataset - if supplied - are not
-        removed from the diff, since working-copy edits are always relevant to the user even if outside the
-        spatial filter.
+        Yields the key, delta for only those feature-deltas from the given dataset diff that match
+        self.spatial_filter. Note that feature-deltas are always considered to match the spatial-filter
+        if they are marked as working-copy edits, since working-copy edits are always relevant to the user
+        even if they are outside the spatial filter.
         """
         # NOTE: This function has to load every feature if it is to do any filtering at all.
```
rcoup, line 266: Comments not relevant anymore?

olsen232: Removed

PR #465, opened 2021-08-01
Allow for changing the spatial filter with kart checkout

olsen232:
## Description

`kart checkout --spatial-filter=SPEC` now updates the spatial filter. The syntax is the same as for `kart init` and `kart clone`.
The spatial filter cannot be changed while there are uncommitted changes in the working copy (this avoids certain potential problems with handling probably-accidental-PK-collisions - note that these aren't actually handled yet however), and it works by rewriting the entire working copy from scratch (anything else would be an optimisation, and would only speed things up in the case that the intersection between the old and the new spatial filter is large compared to the their total size). 

Not done in this PR:
- Handling of probably-accidental-PK-collisions (see above)
- Notifying user when a spatial filter that is stored by reference has changed and/or updating it automatically when possible

## Related links:

### DIFF
--- a/kart/checkout.py
+++ b/kart/checkout.py
@@ -11,9 +11,11 @@ from .exceptions import (
 
 from .exceptions import DbConnectionError
 from .key_filters import RepoKeyFilter
+from .output_util import InputMode, get_input_mode
+from .spatial_filters import SpatialFilterString, spatial_filter_help_text
 from .structs import CommitWithReference
 from .working_copy import WorkingCopyStatus
-from .output_util import InputMode, get_input_mode
+
 
 _DISCARD_CHANGES_HELP_MESSAGE = (
     "Commit these changes first (`kart commit`) or"
@@ -38,12 +40,23 @@ def reset_wc_if_needed(repo, target_tree_or_commit, *, discard_changes=False):
         working_copy.create_and_initialise()
         datasets = list(repo.datasets(target_tree_or_commit))
         working_copy.write_full(target_tree_or_commit, *datasets)
+        return
+
+    spatial_filter_matches = repo.spatial_filter.matches_working_copy(repo)
+    if not spatial_filter_matches:
+        # TODO - support spatial filter changes without doing full rewrites.
+        click.echo(f"Updating {working_copy} with new spatial filter...")
+        datasets = list(repo.datasets(target_tree_or_commit))
+        working_copy.rewrite_full(
+            target_tree_or_commit, *datasets, force=discard_changes
+        )
+        return
 
     db_tree_matches = (
         working_copy.get_db_tree() == target_tree_or_commit.peel(pygit2.Tree).hex
     )
 
-    if discard_changes or not db_tree_matches:
+    if discard_changes or not db_tree_matches or not spatial_filter_matches:
         click.echo(f"Updating {working_copy} ...")
         working_copy.reset(target_tree_or_commit, force=discard_changes)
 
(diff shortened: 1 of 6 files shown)


kart/spatial_filters.py
```diff
@@ -134,36 +155,67 @@ def __init__(self, ref_or_oid):
...
+            return ref, oid, self._resolve_object_contents(obj)
+
+        ref_desc = " or ".join(set([oid, ref]))
+        raise NotFound(
+            f"No spatial filter object was found in the repository at {ref_desc}",
+            exit_code=NO_SPATIAL_FILTER,
+        )
+
+    def resolve(self, repo):
+        # Returns the ResolvedSpatialFilterSpec that this ReferenceSpatialFilterSpec points to.
```
craigds, line 198: This comment needs clarifying I think (and turning into a docstring)

kart/spatial_filters.py
```diff
@@ -134,36 +155,67 @@ def __init__(self, ref_or_oid):
...
         try:
-            obj = repo[self.ref_or_oid]
+            obj = repo[oid]
         except (KeyError, ValueError):
             pass
 
         if obj is not None:
+            return None, oid, self._resolve_object_contents(obj)
+
+        ref = self.ref_or_oid
```
craigds, line 177: So this function is assuming (AFAICT) that at this point the `self.ref_or_oid` is the name of a ref rather than an OID. Is that true? It seems like it could be an OID of an unfetched object also. Is there any way to tell?

kart/working_copy/base.py
```diff
@@ -1010,6 +1046,18 @@ def _delete_meta(self, sess, dataset):
...
+    def rewrite_full(self, commit, *datasets, force=False):
+        """
+        Rewrites all of the given datasets from scratch to match the given commit, and updates the state table tree.
+        Since write_full honours the current repo spatial filter, this also ensures that the working copy spatial
+        filter is up to date.
+        """
+        if not force:
+            self.check_not_dirty()
+
+        self.drop_table(commit, *datasets)
```
craigds, line 1058: should rename to `drop_tables` ?

kart/working_copy/base.py
```diff
@@ -1010,6 +1046,18 @@ def _delete_meta(self, sess, dataset):
         """
         raise NotImplementedError()
 
+    def rewrite_full(self, commit, *datasets, force=False):
```
craigds, line 1049: should this method be done in one transaction? Otherwise a client watching carefully will see tables be dropped, and then recreated shortly afterwards. For big WCs those could be a while apart.

craigds (approved):

kart/checkout.py
```diff
@@ -38,12 +40,23 @@ def reset_wc_if_needed(repo, target_tree_or_commit, *, discard_changes=False):
...
+            target_tree_or_commit, *datasets, force=discard_changes
+        )
+        return
 
     db_tree_matches = (
         working_copy.get_db_tree() == target_tree_or_commit.peel(pygit2.Tree).hex
     )
 
-    if discard_changes or not db_tree_matches:
+    if discard_changes or not db_tree_matches or not spatial_filter_matches:
```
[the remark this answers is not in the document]
olsen232: Done

kart/spatial_filters.py
olsen232: Expanded the docstring and moved it to a superclass.

kart/spatial_filters.py
olsen232: It does assume that if it doesn't find the object, it must be a filter - or missing. It could be that it is an unfetched object however. But whether non-existent or unfetched, the end results will be this:
`No spatial filter object was found in the repository at 9d5d8e7 or refs/filters/9d5d8e7` - which is true and describes the problem accurately.

I'll add a TODO to handle unfetched objects here somehow, but it's part of a bigger problem - we can't currently tell if objects are non-existent / corrupted or merely unfetched, and we don't specifically handle unfetched objects anywhere.

kart/working_copy/base.py
olsen232: Done

kart/working_copy/base.py
olsen232: Done

PR #466, opened 2021-08-03
Feature-count estimate improvements for string-PK datasets

craigds:
## Description

This is part I of a two part series to rework the feature-count estimates to be more performant, more accurate and more reliable.

Datasets with string PKs already use quite a nice algorithm for generating feature-count estimates. This PR has two incremental improvements:

1. actually take the supplied accuracy into account (`fast` vs `good` etc) by sampling more trees when higher accuracy is desired
2. When we find 30 children in a tree with a branching factor of 64, (and the children are also trees, not blobs) the current code was just returning 30 for the number of estimated features. This is subtly incorrect since features are assigned to trees randomly and thus multiple features sometimes collide into the same tree. This makes the correct result closer to 40, which can be given by this formula:

```python
>>>  log(1 - 30/64) / log(1-1/64)
40.164352384390334
```
[(derivation is in this google doc)](https://docs.google.com/document/d/11CeJKbiNQoLmhDcYIM68cJSA_nKBHW7kYVybh2N-Lww/edit#heading=h.7z95y6hc62gn)

## Related links:

none

## Checklist:

### DIFF
--- a/kart/diff_estimation.py
+++ b/kart/diff_estimation.py
@@ -1,7 +1,10 @@
 import logging
+import math
 import statistics
 import subprocess
 import time
+
+import pygit2
 
 from kart.exceptions import SubprocessError
 
@@ -110,7 +113,7 @@ def get_approximate_diff_blob_count(
 
     if path_encoder.DISTRIBUTED_FEATURES:
 
-        diff_count, samples_taken = _recursive_distributed_diff_count(
+        diff_count, samples_taken = _recursive_distributed_diff_estimate(
             repo, tree1, tree2, path_encoder.branches, 16
         )
         return int(round(diff_count))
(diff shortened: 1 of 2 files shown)


kart/diff_estimation.py
```diff
@@ -197,24 +206,39 @@ def _nonrecursive_diff(tree_a, tree_b):
...
     L.debug(f"Found {diff_size} diffs, checking next level:")
 
     total_subsample_size = 0
     total_subsamples_taken = 0
     total_samples_taken = 0
     for tree1, tree2 in diff.values():
-        subsample_size, samples_taken = _recursive_distributed_diff_count(
-            repo, tree1, tree2, branch_count, total_samples_to_take
-        )
+        if isinstance(tree1, pygit2.Blob) or isinstance(tree2, pygit2.Blob):
```
olsen232, line 235: FYI instead of checking the type of the object, you can access its `type_str` attribute which will be either "blob" or "tree". Makes no difference here so I don't care which you use - but it can make tests easier to write since you can create test objects that present the same interface as pygit2.Blob or pygit2.Tree, without actually using pygit2 at all. So, just keep in mind if you need it

olsen232 (approved):

kart/diff_estimation.py
craigds: sounds good. Here though, the trees might also be None, so I'd have to guard against that if I started accessing attributes on them, whereas the `isinstance` check works regardless

PR #468, opened 2021-08-04
Enable auto-incrementing PKs for all WC types.

olsen232:
When the working copy is written, Kart now sets up a sequence which supplies the next unassigned PK value and sets it as the default value for the PK column.

This is normally simply a convenience for the user - the user doesn't have to think of a new PK, the next unassigned one is
automatically supplied for them.

However, it serves another purpose when a spatial filter is active. The PK that is automatically supplied is guaranteed not to clash with any feature inside OR outside of the spatial filter - the user may have no other easy way of providing such a PK themselves without disabling the spatial filter.

## Related links:

### DIFF
--- a/kart/sqlalchemy/adapter/postgis.py
+++ b/kart/sqlalchemy/adapter/postgis.py
@@ -72,6 +72,17 @@ class KartAdapter_Postgis(BaseKartAdapter, Db_Postgis):
     ZM_FLAG_TO_STRING = {0: "", 1: "M", 2: "Z", 3: "ZM"}
 
     @classmethod
+    def v2_column_schema_to_sql_spec(cls, col, v2_obj=None, has_int_pk=False):
+        col_name = cls.quote(col.name)
+        sql_type = cls.v2_type_to_sql_type(col, v2_obj)
+
+        # Make int PKs auto-increment.
+        if has_int_pk and col.pk_index is not None:
+            sql_type = sql_type.replace("INT", "SERIAL")
+
+        return f"{col_name} {sql_type}"
+
+    @classmethod
     def v2_type_to_sql_type(cls, col, v2_obj=None):
         sql_type = super().v2_type_to_sql_type(col, v2_obj)
 
--- a/kart/sqlalchemy/adapter/postgis.py
+++ b/kart/sqlalchemy/adapter/postgis.py
@@ -1,4 +1,5 @@
 import decimal
+import re
 
 from osgeo.osr import SpatialReference
 from psycopg2.extensions import Binary
@@ -78,7 +79,8 @@ class KartAdapter_Postgis(BaseKartAdapter, Db_Postgis):
 
         # Make int PKs auto-increment.
         if has_int_pk and col.pk_index is not None:
-            sql_type = sql_type.replace("INT", "SERIAL")
+            # SMALLINT, INTEGER, BIGINT -> SMALLSERIAL, SERIAL, BIGSERIAL
+            sql_type = re.sub("INT(EGER)?", "SERIAL", sql_type)
 
         return f"{col_name} {sql_type}"
 
--- a/kart/working_copy/base.py
+++ b/kart/working_copy/base.py
@@ -983,7 +983,7 @@ class BaseWorkingCopy:
 
     def _initialise_sequence(self, sess, dataset):
         """Sets up an auto-increment integer PK sequence so that a useful default PK value is provided."""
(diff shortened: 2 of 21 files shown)


kart/working_copy/sqlserver.py
```diff
@@ -140,6 +140,32 @@ def _drop_spatial_index(self, sess, dataset):
         # SQL server deletes the spatial index automatically when the table is deleted.
         pass
 
+    def _initialise_sequence(self, sess, dataset):
+        start = dataset.feature_path_encoder.find_start_of_unassigned_range(dataset)
+        if start:
+            seq_name = f"{dataset.table_name}_{dataset.primary_key}_seq"
+            sess.execute(
+                f"CREATE SEQUENCE {self.table_identifier(seq_name)} START WITH {start};",
+            )
```
rcoup, line 149: this gets called for every working-copy write... won't `CREATE SEQUENCE` fail on a WC update where the sequence already exists?

kart/sqlalchemy/adapter/postgis.py
```diff
@@ -71,6 +71,17 @@ class KartAdapter_Postgis(BaseKartAdapter, Db_Postgis):
...
     ZM_FLAG_TO_STRING = {0: "", 1: "M", 2: "Z", 3: "ZM"}
 
+    @classmethod
+    def v2_column_schema_to_sql_spec(cls, col, v2_obj=None, has_int_pk=False):
+        col_name = cls.quote(col.name)
+        sql_type = cls.v2_type_to_sql_type(col, v2_obj)
+
+        # Make int PKs auto-increment.
+        if has_int_pk and col.pk_index is not None:
+            sql_type = sql_type.replace("INT", "SERIAL")
```
rcoup, line 81: there's also `SMALLSERIAL` (16bit) and `BIGSERIAL` (64bit)

rcoup (approved):
Looking good 👍

kart/sqlalchemy/adapter/postgis.py
rcoup: oh... this replaces the substring in the type name 😬  Comment?

kart/working_copy/postgis.py
```diff
@@ -192,6 +193,21 @@ def _drop_spatial_index(self, sess, dataset):
         # PostGIS deletes the spatial index automatically when the table is deleted.
         pass
 
+    def _initialise_sequence(self, sess, dataset):
+        start = dataset.feature_path_encoder.find_start_of_unassigned_range(dataset)
+        if start:
+            quoted_seq_identifier = sess.scalar(
```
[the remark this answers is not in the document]
olsen232: It is quoted! Perhaps it has to be because it's the only unambiguous way for postgresql to return a fully qualified identifier - for instance, I think an example of a valid but difficult-to-convey SCHEMA_NAME.SEQUENCE_NAME would be `"one.two"."three.four"`

kart/working_copy/base.py
```diff
@@ -977,6 +981,15 @@ def _drop_spatial_index(self, sess, dataset):
         """Inverse of _create_spatial_index_* - deletes the spatial index."""
         pass
 
+    def _initialise_sequence(self, sess, dataset):
+        """Sets up an auto-increment integer PK sequence so that a useful default PK value is provided."""
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
olsen232: If you like!

kart/working_copy/sqlserver.py
olsen232: It only gets called during `write_full` which both creates and populates the table - and I try to delete it again when deleting the table. So, it's actually working fine. But there is still room for improvement:
1. I may as well make this a CREATE or ALTER depending on in the sequence exists - it's more robust in case something goes wrong / the user messes with the working copy in some way.
2. Just realised that the sequence should be updated every time you checkout a different commit, even in the case where that commit doesn't cause the table to be dropped + rewritten. This isn't done yet. Will see what's involved and add it to this PR or to a follow up PR.

kart/sqlalchemy/adapter/postgis.py
olsen232: Added a comment.
Forgot about INTEGER - fixed this bug.

kart/working_copy/sqlserver.py
olsen232: Done and done

PR #470, opened 2021-08-09
Add warnings / errors for probably-accidental-PK-conflicts

olsen232:
## Description

Modifies the existing BaseDiffWriter so that the entire DiffWriter family of classes can better keep track of which features match or do not match the spatial filter and why they do or don't match. Three new features are built on top of this:

- A warning is shown when you run `kart diff` and it turns out that you are guilty of a "probably-accidental-PK-collision" - you have inserted a new feature, but you have reused a PK that is already in use by a feature outside your current spatial filter.
- An error if you try to commit a change that has a "probably-accidental-PK-collision". (There is a flag that can force this change to commit anyway, in which case, you will overwrite the feature that used to have this PK with your new feature).
- Automatic removal of newly committed features from the working copy that do not match the spatial filter

Still TODO - closely related to this, `kart status` still doesn't tell you anything about probably-accidental-PK-collisions - not in text mode nor in JSON mode.


kart/commit.py
```diff
@@ -35,13 +37,55 @@
...
+            self.RECORD_SPATIAL_FILTER_STATS = True
+            self.spatial_filter_pk_conflicts = {
+                ds_path: [] for ds_path in self.all_ds_paths
+            }
+            self.remove_from_wc_post_commit = {
+                ds_path: [] for ds_path in self.all_ds_paths
+            }
+        else:
+            self.RECORD_SPATIAL_FILTER_STATS = False
+            self.spatial_filter_pk_conflicts = self.remove_from_wc_post_commit = None
```
craigds, line 54: minor: IMO avoid `a = b = c` lines; it's always clearer to just put them on two lines

kart/commit.py
```diff
@@ -35,13 +37,55 @@
 from .repo import KartRepoFiles
 
 
+class CommitDiffWriter(BaseDiffWriter):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if not self.spatial_filter.match_all:
+            self.RECORD_SPATIAL_FILTER_STATS = True
```
craigds, line 45: this thing seems like it shouldn't be styled like a constant

kart/commit.py
```diff
@@ -35,13 +37,55 @@
...
+
+    def get_repo_diff(self):
+        repo_diff = super().get_repo_diff()
+
+        if not self.RECORD_SPATIAL_FILTER_STATS:
+            return repo_diff
+
+        for ds_path, ds_diff in repo_diff.items():
+            # Applies the spatial filter but ignores the result, just so that spatial_filter_stats are recorded.
+            for _ in self.filtered_ds_feature_deltas(ds_path, repo_diff[ds_path]):
```
craigds, line 64:
```suggestion
            for _ in self.filtered_ds_feature_deltas(ds_path, ds_diff):
```

kart/commit.py
```diff
@@ -51,7 +95,18 @@
...
+)
+@click.option(
+    "--allow-pk-conflicts",
+    is_flag=True,
+    default=False,
+    help=(
+        "Usually, inserting some new features into the working copy that have the same primary key values as existing "
+        "features that are not in the working copy because they do not match the spatial filter is a mistake - if "
+        "these were committed, it could accidentally overwrite existing features you were unaware of. This option "
+        "bypasses the safety."
```
craigds, line 109: the first clause here is a marathon, I suggest breaking it up a bit
```suggestion
        "Usually, it is a mistake to check features into the working copy that have the same primary key "
        "as something that is outside the spatial filter. If such features were committed, it could "
        "accidentally overwrite existing features you were unaware of. So by default, this is not 
        "allowed. This option bypasses the safety."
```

kart/commit.py
```diff
@@ -89,12 +144,21 @@ def commit(ctx, message, allow_empty, output_format, filters):
...
 
     if not wc_diff and not allow_empty:
         raise NotFound("No changes to commit", exit_code=NO_CHANGES)
 
+    pk_conflicts = commit_diff_writer.spatial_filter_pk_conflicts
+    if not allow_pk_conflicts and pk_conflicts and any(pk_conflicts.values()):
+        commit_diff_writer.write_warnings_footer()
+        raise InvalidOperation(
+            "Aborting commit due to conflicting primary key values - use --allow-pk-conflicts to commit anyway "
+            "(this will overwrite the existing features with these PK values)",
```
craigds, line 158: I'd suggest making mention of the spatial filter here, because otherwise I don't think this will make much sense to a user:

```suggestion
            "Aborting commit due to conflicting primary key values - use --allow-pk-conflicts to commit anyway "
            "(this will overwrite some existing features outside of the current spatial filter)",
```

craigds: oh, except `write_warnings_footer` is just above this, so that probably provides enough context for the user to figure this out 👍

craigds (approved):

kart/commit.py
olsen232: Done

kart/commit.py
olsen232: Done

kart/commit.py
olsen232: Done

kart/commit.py
olsen232: Done

kart/commit.py
olsen232: Done anyway

PR #472, opened 2021-08-18
log: Add `--output-format=json-lines`

craigds:
## Description

Adds the `json-lines` output format to the `log` command (already exists for `diff`, `show`)

Just outputs the same stuff as the `json` output, but with each item on its own line for easy stream consumption.

## Related links:

none

## Checklist:

### DIFF
--- a/kart/output_util.py
+++ b/kart/output_util.py
@@ -190,7 +190,7 @@ def dump_json_output(
 
     fp = resolve_output_path(output_path)
 
-    highlit = json_style == "pretty" and can_output_colour(fp)
+    highlit = can_output_colour(fp)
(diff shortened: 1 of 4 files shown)


kart/output_util.py
```diff
@@ -190,7 +190,7 @@ def dump_json_output(
 
     fp = resolve_output_path(output_path)
 
-    highlit = json_style == "pretty" and can_output_colour(fp)
+    highlit = can_output_colour(fp)
```
olsen232, line 193: What's the reasoning for this change? Prettier json-lines output?

olsen232 (approved):

kart/output_util.py
craigds: yep. optional I suppose. It still doesn't colour the output if it's not going to a TTY though

PR #475, opened 2021-09-14
Adds a new command `spatial-tree index`

olsen232:
Command only works if pywraps2 is manually installed into the kart venv.
Although tests exist, they are currently skipped for this reason.
(Avoiding modifying the kart build while it is red - hopefully fixed sometime soon).

When command is run, creates a spatial-index using s2 cells, but this isn't yet used for anything.
Unlike the proof-of-concept, can be used to index a set of commits and their ancestors, minus another set of commits and their ancestors (ie the ones that we already indexed last time).

Changelog is not yet updated - will wait until the command works and does something useful.

### DIFF
--- a/kart/spatial_tree.py
+++ b/kart/spatial_tree.py
@@ -44,7 +44,7 @@ DS_PATH_PATTERN = r'(.+)/\.(sno|table)-dataset/'
 
 
 def _parse_revlist_output(line_iter, rel_path_pattern):
-    full_path_pattern = DS_PATH_PATTERN + rel_path_pattern
+    full_path_pattern = re.compile(DS_PATH_PATTERN + rel_path_pattern)
 
     for line in line_iter:
         parts = line.split(" ", maxsplit=1)
@@ -52,7 +52,7 @@ def _parse_revlist_output(line_iter, rel_path_pattern):
             continue
         oid, path = parts
 
-        m = re.match(full_path_pattern, path)
+        m = full_path_pattern.match(path)
         if not m:
             continue
         ds_path = m.group(1)
@@ -97,11 +97,12 @@ class CrsHelper:
         return transforms
 
     def iter_crs_oids(self, ds_path):
-        cmd = (
-            _revlist_command(self.repo)
-            + ["--all", "--"]
-            + list(self.all_crs_paths(ds_path))
-        )
+        cmd = [
+            *_revlist_command(self.repo),
+            "--all",
+            "--",
+            *self.all_crs_paths(ds_path),
+        ]
         try:
             r = subprocess.run(
                 cmd,
@@ -144,22 +145,33 @@ class CrsHelper:
 
 
 class SpatialTreeTables(TableSet):
-    """Spatial Tree spec tables - see http://www.geopackage.org/spec/#table_definition_sql"""
+    """Tables for associating a variable number of S2 cells with each feature."""
 
     def __init__(self):
         super().__init__()
 
+        # "blobs" tracks all the features we have indexed (even if they do not overlap any s2 cells).
         self.blobs = Table(
             "blobs",
             self.sqlalchemy_metadata,
+            # From a user-perspective, "rowid" isjust an arbitrary integer primary key.
+            # In more detail: This column aliases to the sqlite rowid of the table.
+            # See https://www.sqlite.org/lang_createtable.html#rowid
+            # Using the rowid directly as a foreign key (see "blob_cells") means faster joins.
+            # The rowid can be used without creating a column that aliases to it, but you shouldn't -
+            # rowids might change if they are not aliased. See https://sqlite.org/lang_vacuum.html)
             Column("rowid", Integer, nullable=False, primary_key=True),
+            # "blob_id" is the git object ID (the SHA-1 hash) of a feature, in binary (20 bytes).
+            # Is equivalent to 40 chars of hex eg: d08c3dd220eea08d8dfd6d4adb84f9936c541d7a
             Column("blob_id", BLOB, nullable=False, unique=True),
             sqlite_autoincrement=True,
         )
 
+        # "blob_cells" associates 0 or more S2 cell tokens with each feature that we have indexed.
         self.blob_cells = Table(
             "blob_cells",
             self.sqlalchemy_metadata,
+            # Reference to blobs.rowid.
             Column(
                 "blob_rowid",
                 Integer,
@@ -167,6 +179,8 @@ class SpatialTreeTables(TableSet):
                 nullable=False,
                 primary_key=True,
             ),
+            # S2 cell token eg "6d6dd90351b31cbf".
+            # To locate an S2 cell by token, see https://s2.sidewalklabs.com/regioncoverer/
             Column(
                 "cell_token",
                 Text,
@@ -200,11 +214,20 @@ def iter_feature_oids(repo, commit_spec):
         )
 
 
-def update_spatial_tree(
-    repo, commit_spec, crs_helper, verbosity=1, clear_existing=False
-):
+def update_spatial_tree(repo, commit_spec, verbosity=1, clear_existing=False):
+    """
+    Index the commits given in commit_spec, and write them to the s2_index.db repo file.
+
+    repo - the Kart repo containing the commits to index, and in which to write the index file.
+    commit_spec - a list of commits to index (ancestors of these are implicitly included).
+        Commits can be exluded by prefixing with '^' (ancestors of these are implicitly excluded).
+        (See git rev-list for the full list of possibilities for specifying commits).
+    verbosity - how much non-essential information to output.
+    clear_existing - when true, deletes any pre-existing data before re-indexing.
+    """
     import pywraps2 as s2
 
+    crs_helper = CrsHelper(repo)
     feature_oid_iter = iter_feature_oids(repo, commit_spec)
 
     s2_coverer = s2.S2RegionCoverer()
@@ -223,8 +246,9 @@ def update_spatial_tree(
 
         SpatialTreeTables.create_all(sess)
 
-    t0 = time.monotonic()
     click.echo(f"Indexing {' '.join(commit_spec)} ...")
+    t0 = time.monotonic()
+    i = 0
 
     # Using sqlite directly here instead of sqlalchemy is about 10x faster.
     # Possibly due to huge number of unbatched queries.
@@ -316,20 +340,15 @@ def _point_f2_cells(s2_coverer, geom, transforms):
     import pywraps2 as s2
 
     g = gpkg_geom_to_ogr(geom)
-
     one_transform = len(transforms) == 1
-    if not one_transform:
-        result = set()
 
+    result = set()
     for transform in transforms:
-        g_transformed = _apply_transform(g, transform, one_transform)
+        g_transformed = _apply_transform(g, transform, overwrite_original=one_transform)
         p = g_transformed.GetPoint()[:2]
         s2_ll = s2.S2LatLng.FromDegrees(p[1], p[0]).Normalized()
         s2_token = s2.S2CellId(s2_ll.ToPoint()).ToToken()
-        if one_transform:
-            return (s2_token,)
-        else:
-            result.add(s2_token)
+        result.add(s2_token)
 
     return result
 
@@ -341,10 +360,10 @@ def _general_s2_cells(s2_coverer, geom, transforms):
     if e is None:
         return ()  # Empty.
 
-    sw_src, ne_src = (e[0::2], e[1::2])
+    sw_src = e[0], e[2]
+    ne_src = e[1], e[3]
 
     result = set()
-
     for transform in transforms:
         s2_ll = []
         for p_src in (sw_src, ne_src):
@@ -403,11 +422,9 @@ def index(ctx, index_all_commits, clear_existing, commits):
     commit_spec = ["--all"] if index_all_commits else list(commits)
 
     repo = ctx.obj.get_repo(allowed_states=KartRepoState.ALL_STATES)
-    crs_helper = CrsHelper(repo)
     update_spatial_tree(
         repo,
         commit_spec,
-        crs_helper,
         verbosity=ctx.obj.verbosity + 1,
         clear_existing=clear_existing,
     )
--- a/kart/spatial_tree.py
+++ b/kart/spatial_tree.py
@@ -1,0 +1,413 @@
+import functools
+import logging
+import re
+import subprocess
+import time
+
+import click
+from osgeo import osr, ogr
+from pysqlite3 import dbapi2 as sqlite
+from sqlalchemy import Column, ForeignKey, Integer, Table, Text
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.types import BLOB
+
+
+from .cli_util import add_help_subcommand, tool_environment
+from .crs_util import make_crs, normalise_wkt
+from .exceptions import SubprocessError
+from .geometry import Geometry, GeometryType, geom_envelope, gpkg_geom_to_ogr
+from .repo import KartRepoState, KartRepoFiles
+from .sqlalchemy import TableSet
+from .sqlalchemy.sqlite import sqlite_engine
+from .serialise_util import msg_unpack
+
+L = logging.getLogger("kart.spatial_tree")
+
+
+S2_MAX_CELLS_INDEX = 8
+S2_MAX_LEVEL = 15
+
+
+def _revlist_command(repo):
+    return [
+        "git",
+        "-C",
+        repo.path,
+        "rev-list",
+        "--objects",
+        "--filter=object:type=blob",
+        "--missing=allow-promisor",
+    ]
+
+
+DS_PATH_PATTERN = r'(.+)/\.(sno|table)-dataset/'
+
+
+def _parse_revlist_output(line_iter, rel_path_pattern):
+    full_path_pattern = DS_PATH_PATTERN + rel_path_pattern
+
+    for line in line_iter:
+        parts = line.split(" ", maxsplit=1)
+        if len(parts) != 2:
+            continue
+        oid, path = parts
+
+        m = re.match(full_path_pattern, path)
+        if not m:
+            continue
+        ds_path = m.group(1)
+        yield ds_path, oid
+
+
+class CrsHelper:
+    """
+    Loads all CRS definitions for a particular dataset,
+    and creates transforms
+    """
+
+    def __init__(self, repo):
+        self.repo = repo
+        self.ds_to_transforms = {}
+
+    def transforms_for_dataset(self, ds_path):
+        transforms = self.ds_to_transforms.get(ds_path)
+        if transforms is None:
+            transforms = self._load_transforms_for_dataset(ds_path)
+            self.ds_to_transforms[ds_path] = transforms
+        return transforms
+
+    def _load_transforms_for_dataset(self, ds_path):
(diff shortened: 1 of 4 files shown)


kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+    @functools.lru_cache()
+    def transform_from_wkt(self, wkt):
+        src_crs = make_crs(wkt)
+        if src_crs.IsGeographic():
+            transform = None
+            desc = f"IDENTITY({src_crs.GetAuthorityCode(None)})"
+        else:
+            target_crs = src_crs.CloneGeogCS()
+            transform = osr.CoordinateTransformation(src_crs, target_crs)
+            desc = f"{src_crs.GetAuthorityCode(None)} -> {target_crs.GetAuthorityCode(None)}"
```
craigds, line 142: this is incorrect I think - we don't want to project to the nearest geographic CRS, we want to project everything to a single CRS. Otherwise S2 cells won't be comparable between different source CRSes.

Different geographic CRSes have different meridians, axis order, possibly units (e.g. radians) etc. We should pick one and always transform to that. Normal choice being EPSG:4326

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+        return transform, desc
+
+
+class SpatialTreeTables(TableSet):
+    """Spatial Tree spec tables - see http://www.geopackage.org/spec/#table_definition_sql"""
```
craigds, line 147: docstring seems wrong, these tables aren't in the GPKG spec

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+        self.blobs = Table(
+            "blobs",
+            self.sqlalchemy_metadata,
+            Column("rowid", Integer, nullable=False, primary_key=True),
```
craigds, line 155: can you add a comment to each of these columns describing what they store? With examples if possible

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+    # Using sqlite directly here instead of sqlalchemy is about 10x faster.
+    # Possibly due to huge number of unbatched queries.
+    # TODO - investigate further.
+    db = sqlite.connect(f"file:{db_path}", uri=True)
+    with db:
+        dbcur = db.cursor()
+
+        for i, (ds_path, feature_oid) in enumerate(feature_oid_iter):
+            if i and progress_every and i % progress_every == 0:
+                click.echo(f"  {i:,d} features... @{time.monotonic()-t0:.1f}s")
```
craigds, line 238: I think this should go to stderr

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+            transforms = crs_helper.transforms_for_dataset(ds_path)
+            if not transforms:
+                continue
+            geom = get_geometry(repo, feature_oid)
+            if geom is None:
+                continue
+            try:
+                s2_cell_tokens = find_s2_cells(s2_coverer, geom, transforms)
+            except Exception as e:
+                L.warning(f"Couldn't locate S2 cells for {feature_oid}:\n{e}")
```
craigds, line 249: should this be a fatal error rather than a warning?

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+        yield from _parse_revlist_output(p.stdout, r"feature/.+")
+    except subprocess.CalledProcessError as e:
+        raise SubprocessError(
+            f"There was a problem with git rev-list: {e}", called_process_error=e
+        )
+
+
+def update_spatial_tree(
+    repo, commit_spec, crs_helper, verbosity=1, clear_existing=False
+):
```
craigds, line 205: could use a docstring

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def _general_s2_cells(s2_coverer, geom, transforms):
+    import pywraps2 as s2
+
+    e = geom_envelope(geom)
+    if e is None:
+        return ()  # Empty.
+
+    sw_src, ne_src = (e[0::2], e[1::2])
```
craigds, line 344: split this into two lines, remove the stride slice. too hard to read

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+
+    for transform in transforms:
+        g_transformed = _apply_transform(g, transform, one_transform)
+        p = g_transformed.GetPoint()[:2]
+        s2_ll = s2.S2LatLng.FromDegrees(p[1], p[0]).Normalized()
+        s2_token = s2.S2CellId(s2_ll.ToPoint()).ToToken()
+        if one_transform:
+            return (s2_token,)
+        else:
+            result.add(s2_token)
```
craigds, line 332: seems unnecessary to special-case the single-transform case (and return a different type?). Does it make something faster?

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+@click.argument(
+    "commits",
+    nargs=-1,
+)
+@click.pass_context
+def index(ctx, index_all_commits, clear_existing, commits):
+    """
+    Indexes all features added by the supplied commits and all of their ancestors.
+    To stop recursing at a particular ancestor or ancestors (eg to stop at a commit
+    that has already been indexed) prefix that commit with a caret: ^COMMIT.
```
craigds, line 396: so if I want to run this in a post-receive hook, and someone has pushed ref `myref`, changing it from `abc123` to `def456`, I would run this like

```
kart spatial-tree index def456 ^abc123
```

That's not super intuitive to me (and I don't think `^{commit}` is an established convention?)

Would it make more sense to take a range argument?

```
kart spatial-tree index abc123..def456
```

tests/test_spatial_tree.py
```diff
@@ -0,0 +1,128 @@
...
+        assert stats.features == 2148
+        assert stats.avg_cell_tokens_per_feature == 1
+        assert stats.avg_cell_token_length == 16
+        assert stats.distinct_cell_tokens == 2143
+
+
+@pytest.mark.skip(reason=SKIP_REASON)
+def test_index_points_idempotent(data_archive, cli_runner):
+    # Indexing the commits one at a time (and backwards) and then indexing --all should
+    # also give the same result, even though everything will have been indexed twice.
```
craigds, line 50: Reckon we can avoid this? it'd be helpful to be able to just `kart spatial-tree index {refname}` without having to specify revisions, and not have it fall off a performance cliff when everything's already been done.

In a post-receive hook, we do know what's been pushed (e.g. `abc123..def456`) but being able to simply run `kart spatial-index def456` means that we need no bootstrapping the first time the post-receive hook is deployed. Whereas if the hook only indexes `abc123..def456` then we also need to run an initial index of `abc123`

I guess it'd just be a matter of storing root tree hashes (or commit hashes) in a separate table in the db, and then using that to skip doing them if they already exist

craigds (changes requested):
Looking good!

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+def update_spatial_tree(
+    repo, commit_spec, crs_helper, verbosity=1, clear_existing=False
+):
+    import pywraps2 as s2
+
+    feature_oid_iter = iter_feature_oids(repo, commit_spec)
+
+    s2_coverer = s2.S2RegionCoverer()
+    s2_coverer.set_max_cells(S2_MAX_CELLS_INDEX)
+    s2_coverer.set_max_level(S2_MAX_LEVEL)
```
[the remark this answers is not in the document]
olsen232: Not sure - @rcoup tuned this number, he can explain why it's this. I imagine there's a tradeoff but I don't know what the tradeoff is.

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def _revlist_command(repo):
+    return [
+        "git",
+        "-C",
+        repo.path,
+        "rev-list",
+        "--objects",
+        "--filter=object:type=blob",
```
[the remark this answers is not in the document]
olsen232: Good point - will add it to the tracking bug, but won't fix build issues for now.

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+def _parse_revlist_output(line_iter, rel_path_pattern):
+    full_path_pattern = DS_PATH_PATTERN + rel_path_pattern
+
+    for line in line_iter:
+        parts = line.split(" ", maxsplit=1)
+        if len(parts) != 2:
+            continue
+        oid, path = parts
+
+        m = re.match(full_path_pattern, path)
```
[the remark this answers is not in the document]
olsen232: Done

kart/spatial_tree.py
```diff
@@ -0,0 +1,413 @@
...
+                )
+        L.info(f"Loaded CRS transforms for {ds_path}: {', '.join(descs)}")
+        return transforms
+
+    def iter_crs_oids(self, ds_path):
+        cmd = (
+            _revlist_command(self.repo)
+            + ["--all", "--"]
+            + list(self.all_crs_paths(ds_path))
+        )
```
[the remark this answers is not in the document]
olsen232: Done

kart/spatial_tree.py
olsen232: Hmm I think you are right. However again I must defer to @rcoup who wrote the prototype - I imagine he had his reasons for doing it this way. If not, I'll change it.

The "Geocentric vs Geodetic coordinates" section in https://s2geometry.io/about/overview might be relevant, but I'm not even sure about that.

kart/spatial_tree.py
olsen232: Fixed

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
olsen232: Not done - it's not an error and `kart import` also writes this to stdout. If this needs to change, will change them both in a separate commit.

kart/spatial_tree.py
olsen232: Probably not -
- a single broken feature / a bug affecting the indexing of a single feature, shouldn't prevent you from indexing your repository.
- if a single feature fails to index, that's no big deal - it will just never be excluded from a clone, regardless of the spatial-filter (much like non-spatial data).

kart/spatial_tree.py
craigds: separate commit is fine, but I think it should go to stderr. it's debatable but e.g. in `kart diff` it's more obvious the diff goes to stdout and any kind of progress or logging must go to stderr to avoid messing up the diff. Here there's not really any stdout but to be consistent logging/progress should still go to stderr

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
olsen232: So actually these arguments are just forwarded verbatim to `git rev-list`. That means:
- ^X is actually an established convention (although I hadn't heard of it)
- X..Y and X...Y actually already work. I just didn't document them since they're syntactic sugar for some X, Y, ^Z way of writing the some specification. 

The X, Y, ^Z way of specifying commits is more expressive since if you have new branch heads N1, N2, N3, N4, and old branch heads O1, O2 that you already indexed last time, you can just index: `N1 N2 N3 N4 ^O1 ^O2` without worrying which old branch head(s) are ancestors of which new branch heads (it might be true that all are ancestors of all).

tests/test_spatial_tree.py
olsen232: Yeah this could be good. Won't do it in this PR though - it would be a separate first step in order to find the commit-spec that we want to use to run `git rev-list <commit-spec>`. So all this current code would remain pretty much unchanged and I can later add the other code that makes it less dumb.
Disinclined to do it as the next step - going to try and get the proof-of-concept working again but using some of the new pieces.

kart/spatial_tree.py
olsen232: Yeah this was a silly and untested overoptimization - no longer special cased

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
craigds: Maybe change the docstring to avoid using that and just say you can pass it anything `git rev-list` accepts

tests/test_spatial_tree.py
craigds: ok

craigds (approved):

kart/spatial_tree.py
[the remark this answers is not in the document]
edcorkery: The precision should _eventually_
1. Be user modifiable/set
2. Be more precise for “smaller” datasets automatically on a sliding scale

So long as these are possible later

kart/spatial_tree.py
rcoup: Logic was avoiding needless & expensive datum transforms. But Craig's right, axis-order/units/meridians can be in different places so that's not a great plan.

WGS84 seems sensible to me.

kart/spatial_tree.py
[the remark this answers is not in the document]
rcoup: 1. the client-side repo→working-copy precision is accurate
2. we don't need high precision for server-side filtering, we need high speed and minimal storage space. Including a pile of bonus features won't make much meaningful difference to clone-time, but going overboard on precision has a sizeable impact on storage costs and potentially slows down every fetch.
3. if your data is small it'll be *much much* faster to do a full clone than going anywhere near a filter. Full clones take advantage of bitmaps, existing packfiles, etc which provide a big speed boost.

We can mess with the settings later once it's in use.

kart/spatial_tree.py
```diff
@@ -0,0 +1,430 @@
...
+
+        SpatialTreeTables.create_all(sess)
+
+    click.echo(f"Indexing {' '.join(commit_spec)} ...")
+    t0 = time.monotonic()
+    i = 0
+
+    # Using sqlite directly here instead of sqlalchemy is about 10x faster.
+    # Possibly due to huge number of unbatched queries.
+    # TODO - investigate further.
```
rcoup, line 255: Yeah, this needs to perform well... can we TODO it now?

Using SQLAlchemy shouldn't be magnitudes slower.

kart/spatial_tree.py
```diff
@@ -0,0 +1,430 @@
...
+            )
+
+    t1 = time.monotonic()
+    click.echo(f"Indexed {i} features in {t1-t0:.1f}s")
+
+
+NO_GEOMETRY_COLUMN = object()
+
+
+def get_geometry(repo, feature_oid):
```
rcoup, line 302: Presumably we have this functionality somewhere else? Can we re-use it?

tests/test_spatial_tree.py
rcoup: @craigds what's the use case for running `spatial-tree index {ref}`?

kart/spatial_tree.py
```diff
@@ -0,0 +1,430 @@
...
+    with db:
+        dbcur = db.cursor()
+
+        for i, (ds_path, feature_oid) in enumerate(feature_oid_iter):
+            if i and progress_every and i % progress_every == 0:
+                click.echo(f"  {i:,d} features... @{time.monotonic()-t0:.1f}s")
+
+            transforms = crs_helper.transforms_for_dataset(ds_path)
+            if not transforms:
+                continue
```
rcoup, line 266: can't this be pulled out of the loop?

kart/spatial_tree.py
```diff
@@ -0,0 +1,430 @@
...
+
+    click.echo(f"Indexing {' '.join(commit_spec)} ...")
+    t0 = time.monotonic()
+    i = 0
+
+    # Using sqlite directly here instead of sqlalchemy is about 10x faster.
+    # Possibly due to huge number of unbatched queries.
+    # TODO - investigate further.
+    db = sqlite.connect(f"file:{db_path}", uri=True)
+    with db:
```
rcoup, line 257: presume this deals with concurrency ok? Very possible multiple of these will be running/queued at the same time for a repository. Think two pushes in quick succession.

kart/spatial_tree.py
olsen232: It's using sqlite directly right now instead of sqlalchemy so it's fast right now. But I can take some time and try and figure out what's making sqlalchemy slow (might be relevant to working copies also)

kart/spatial_tree.py
olsen232: It's using WAL which I think should be okay - I can try it out

kart/spatial_tree.py
olsen232: This one works a bit different - mostly we load the repository at a particular revision. Then every feature we try to read has an attached legend ID, and we can combine that legend with the current schema to see which one is the geometry column. However, in this case, I'm running a magic git command that iterates through all features added in a certain set of commits. Its not straight-forward when using this command to see exactly which commit the feature was added at, so its not obvious what should be done in order to look up the attached legend or the current schema. But, it is possible to find the geometry field without doing those things, so that's what this code does.

kart/spatial_tree.py
olsen232: Okay. Will switch this to convert to WGS 84. Is there a way we can tell it not to bother doing the expensive and needless thing?

kart/spatial_tree.py
olsen232: No - different datasets have different transforms. This indexes features from all datasets in the same loop, possibly in commit order. Eg this loop indexes features from commit X then commit Y then commit Z, and these commits could each involve a few different datasets, and therefore different transforms - there's no loop that only deals with one dataset and therefore one set of transforms. However, transforms are cached so it doesn't take long to find them again.

kart/spatial_tree.py
rcoup: `OSRSpatialReference.IsSame(other)`?

tests/test_spatial_tree.py
olsen232: I also don't know exactly what Craig intends by `spatial-tree index {ref}`, but I agree with him it would be cool if you could just call `spatial-tree index --all` and it would return immediately if everything was already up to date, which I think is mostly what he's getting at.

craigds: That, as well as building for this not being 100% reliable all the time. If it dies once and fails to index some revisions, we want it to automatically catch up during the next push. Which won't happen if we have to explicitly provide a revision range.

rcoup: it's simplistic to think of it as a revision range... we get a list of refs with (Old, New) commit oids (and Old or New can be null)... then we turn that into `^old1 ^old2 ^old3 new2 new3 new4` which spits out all the commit IDs in the new but not the old refs.

> we want it to automatically catch up during the next push

We want it to catch up during the first of maintenance or after the next push.

So, we save root tree IDs?

olsen232: I guess after we finish indexing, we store all of the branch head commit oids in a table somewhere - but in a minimal form, eg, no need to store a commit OID if we are also storing its descendant. Every time an index operation succeeds we replace this minimal list with the new minimal list.

rcoup: equivalent of `git for-each-ref refs/heads/ refs/tags/`?

PR #480, opened 2021-09-27
Update spatial-tree index to not need arguments

olsen232:
## Description

As suggested in the last PR, rather than having to remember to run `spatial-tree index OLD_BRANCH_HEADS NEW_BRANCH_HEADS` any time anything new to index is pushed, it would be easier if you could just run `spatial-tree index` on any kind of regular-basis and have it keep track of where it was up to itself.

The former way means that if the spatial-index doesn't succeed for some reason after a particular push - perhaps the machine crashed - and it isn't rerun, then the index will be incomplete from then on. The latter way means that any time the command is run, it will just carry on wherever it got up to last time it succeeded, including reindexing the last bit if it crashed partway through last time.

It's still possible to index only part of the history, by specifying a set of commits - those commits and their ancestors will be indexed. Indexing isolated commits without indexing their ancestors is not possible.

### DIFF
--- a/kart/spatial_tree.py
+++ b/kart/spatial_tree.py
@@ -17,9 +17,11 @@ from .crs_util import make_crs, normalise_wkt
 from .exceptions import SubprocessError
 from .geometry import Geometry, GeometryType, geom_envelope, gpkg_geom_to_ogr
 from .repo import KartRepoState, KartRepoFiles
+from .serialise_util import msg_unpack
+from .structs import CommitWithReference
 from .sqlalchemy import TableSet
 from .sqlalchemy.sqlite import sqlite_engine
-from .serialise_util import msg_unpack
+
 
 L = logging.getLogger("kart.spatial_tree")
 
@@ -150,6 +152,17 @@ class SpatialTreeTables(TableSet):
     def __init__(self):
         super().__init__()
 
+        # "commits" tracks all the commits we have indexed.
+        # A commit is only considered indexed if ALL of its ancestors are also indexed - this means
+        # relatively few commits need to be recorded as being indexed in this table.
+        self.commits = Table(
+            "commits",
+            self.sqlalchemy_metadata,
+            # "commit_id" is the commit ID (the SHA-1 hash), in binary (20 bytes).
+            # Is equivalent to 40 chars of hex eg: d08c3dd220eea08d8dfd6d4adb84f9936c541d7a
+            Column("commit_id", BLOB, nullable=False, primary_key=True),
+        )
+
         # "blobs" tracks all the features we have indexed (even if they do not overlap any s2 cells).
         self.blobs = Table(
             "blobs",
@@ -198,8 +211,8 @@ def drop_tables(sess):
     sess.execute("DROP TABLE IF EXISTS blobs;")
 
 
-def iter_feature_oids(repo, commit_spec):
-    cmd = _revlist_command(repo) + commit_spec
+def iter_feature_oids(repo, start_commits, stop_commits):
+    cmd = [*_revlist_command(repo), *start_commits, "--not", *stop_commits]
     try:
         p = subprocess.Popen(
             cmd,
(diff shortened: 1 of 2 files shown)


kart/spatial_tree.py
```diff
@@ -214,21 +227,75 @@ def iter_feature_oids(repo, commit_spec):
...
+        r = subprocess.run(
+            cmd,
+            encoding="utf8",
+            check=True,
+            capture_output=True,
+            env=tool_environment(),
+        )
+    except subprocess.CalledProcessError as e:
+        raise SubprocessError(
+            f"There was a problem with git show-ref: {e}", called_process_error=e
```
craigds, line 242: command was merge-base, not show-ref

kart/spatial_tree.py
```diff
@@ -214,21 +227,75 @@ def iter_feature_oids(repo, commit_spec):
         )
 
 
-def update_spatial_tree(repo, commit_spec, verbosity=1, clear_existing=False):
+def _minimal_description_of_commit_set(repo, commits):
```
craigds, line 230: could use a docstring

kart/spatial_tree.py
```diff
@@ -214,21 +227,75 @@ def iter_feature_oids(repo, commit_spec):
...
+
+
+def _build_on_last_index(repo, start_commits, sess):
+    commits_table_exists = sess.scalar(
+        "SELECT count(*) FROM sqlite_master WHERE name = 'commits';"
+    )
+    if commits_table_exists:
+        stop_commits = set(
+            row[0].hex() for row in sess.execute("SELECT commit_id FROM commits;")
+        )
```
craigds, line 254:
```suggestion
        stop_commits = {
            row[0].hex() for row in sess.execute("SELECT commit_id FROM commits;")
        }
```

kart/spatial_tree.py
```diff
@@ -214,21 +227,75 @@ def iter_feature_oids(repo, commit_spec):
...
+            env=tool_environment(),
+        )
+    except subprocess.CalledProcessError as e:
+        raise SubprocessError(
+            f"There was a problem with git show-ref: {e}", called_process_error=e
+        )
+    return set(r.stdout.splitlines())
+
+
+def _build_on_last_index(repo, start_commits, sess):
```
craigds, line 247: needs a docstring

kart/spatial_tree.py
```diff
@@ -214,21 +227,75 @@ def iter_feature_oids(repo, commit_spec):
...
+        repo, start_commits | stop_commits
+    )
+    start_commits = all_independent_commits - stop_commits
+    return (start_commits, stop_commits, all_independent_commits)
+
+
+def _format_commits(repo, commit_ids):
+    if not commit_ids:
+        return None
+    length = len(repo[next(iter(commit_ids))].short_id)
```
craigds, line 268:
```suggestion
    length = len(repo[commit_ids[0]].short_id)
```

kart/spatial_tree.py
```diff
@@ -292,6 +365,11 @@ def update_spatial_tree(repo, commit_spec, verbosity=1, clear_existing=False):
                 params,
             )
 
+        # Update indexed commits.
+        params = [(bytes.fromhex(commit_id),) for commit_id in all_independent_commits]
+        dbcur.execute("DELETE FROM commits;")
```
craigds, line 370: there's a race condition here, if two of these things run in parallel.

can/should/do we prevent them from running in parallel? Perhaps by just taking an exclusive lock on the db before we fetch the commit IDs up above. (but that'd mean we'd have to use the same connection/session throughout)

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
olsen232: Done

kart/spatial_tree.py
olsen232: I'm using lots of sets so they're not subscriptable.

kart/spatial_tree.py
olsen232: Talked about this offline - since we're using WAL, a single writer gets to hold the lock on the DB and update in a self-consistent way. Since the reads are done separately to the write, there's a possibility of two different writers both trying to re-index the same commits, but since that's supposed to be idempotent (see the tests) that's just wasted effort - it won't actually lead to any inconsistency. WDYT?

craigds (approved):

Issue #477, opened 2021-09-24
`apply`: Accept patches containing a head commit hash and no `-` values


craigds:
## background

At present, kart applies patches where edits and deletes are represented by a `-` chunk, e.g.:

```javascript
// ...
      "feature": [
        /* a deleted feature */
        {
          "-": {
            "ID": 12110,
            "GEOMETRY": "010123456000000FE0D18E02DFA3741F4AC",
            "DATE_MOD": "2010-11-22",
            "HEIGHT": 0.0,
            "WEIGHT": 0.0,
          }
        },
        /* an updated feature */
        {
          "-": {
            "ID": 12111,
            "GEOMETRY": "010123456000000FE0D18E02DFA3741F4AC",
            "DATE_MOD": "2019-11-22",
            "HEIGHT": 0.0,
            "WEIGHT": 0.0,
          },
          "+": {
            "ID": 12111,
            "GEOMETRY": "01012345600CE3D4F302EFA3741C2A6FC2B",
            "DATE_MOD": "2020-11-19",
            "HEIGHT": 0.0,
            "WEIGHT": 0.0,
          }
        }
// ...

```

however, this is unfortunate for a few reasons:

1. clients generating patches have to keep track of the 'original' blob as well as the new blob, doubling memory usage for the working feature set
2. individual features are represented very verbosely. In the above example, 2 of the fields are unchanged, so don't really need to be included. They are useful for conflict detection though - if this patch is applied to another HEAD and those fields have been edited to different values, we want the patch application to fail.

This situation (especially point 1) doesn't adapt well to low-resource environments.

However, we can't just remove the `-` blobs, and we can't just remove unchanged fields either - to do that would mean we get no conflict detection when the patch is applied.

## proposal

Instead, I propose inserting a HEAD commit hash into the patch itself as an alternative means of conflict resolution:

```javascript
{
  "kart.diff/v1+hexwkb": {...},
  "kart.patch/v1": {
//    ...,
    "original_head": "abcdef1234567890abcdef1234567890abcdef12"
  }
```

This should be the commit on top of which the patch is guaranteed to apply. So if the patch was generated from a commit, it should be that commit's parent; if the patch was constructed on top of some repo data, it should be the hash of the repo's resolved HEAD at the time when the data was fetched.

With this, we no longer need to include either:

 * `-` blobs, except for deletes - and then we only need the primary key field
 * unchanged fields in edited features

Kart can detect conflicts by comparing updated features against the old version of the feature.

[...]

PR #482, opened 2021-09-29
apply: Initial support for 'minimal' style patches

craigds:
## Description

This supports applying patches where edits can omit the `-` objects for
each change, provided that an `originalParent` key exists in the patch
header, and that its value is set to a commit OID which is present in
the repo.

Such patches are now produced by `kart create-patch --patch-type=minimal`

I intend to also allow these patches omit unchanged fields from feature blobs, but that is _not_ in this PR.

## Related links:

none

## Checklist:


kart/json_diff_writers.py
```diff
@@ -64,14 +68,20 @@ def write_diff(self):
         )
         self.write_warnings_footer()
 
+    def _make_meta_delta_serializable(self, delta):
```
olsen232, line 71: There could maybe be a better name for this - yes it's serialisable but there might be a more specific or positive way to describe what we're doing here

olsen232 (approved):

tests/test_apply.py
```diff
@@ -236,49 +241,209 @@ def test_apply_onto_other_ref(data_working_copy, cli_runner):
...
+    data_archive, cli_runner, tmp_path
+):
+
+    patch = points_patch(
+        {
+            "feature": [
+                {
+                    # one edit (in minimal style it looks like an insert, but FID 1182 already exists)
+                    "+": {
+                        "fid": 1182,
```
rcoup, line 339: I haven't reviewed it all, but this is reasonably fundamental (maybe should be on #477): should the implementation of this be passing the commit as a parameter rather than putting it _in_ the payload?

The comment here worries me a bit that we're writing patches that look like inserts but are actually edits.

craigds: > should the implementation of this be passing the commit as a parameter rather than putting it in the payload?

I'm unsure what you're suggesting here. The original parent commit hash is stored in the patch header (`originalParent`, feel free to suggest a better name; I don't really like that key but I couldn't think of a better one) because it's part of the patch - omitting it would mean we can't apply this patch because we don't know the starting state and hence can't resolve conflicts when the `-` blobs are missing.


> The comment here worries me a bit that we're writing patches that look like inserts but are actually edits.

what if we changed the key from `+` to something else for updates in minimal patches? e.g. `-+`? I'd be happy to do that, it would make the intent of the patch much clearer.

rcoup: > I'm unsure what you're suggesting here.

Would something like `kart apply <TO-WHAT> <PATCH-FILE>` → `kart apply c0ffee my.patch` be clearer or more/less useful? Compared with putting `c0ffee` _into_ `my.patch`

craigds: I don't think so - without that `c0ffee` the patch itself is useless and can't be applied, so it doesn't make sense to me to exclude that info from the patch. It isn't the thing you're applying the patch _to_ (you're always applying the patch to HEAD) but it tells kart what state you generated the patch _from_. You're right that we _could_ pass this info in a sideband but I don't see why we would - it's an indispensible part of the patch. Without that all we have is the state of a bunch of features from some unknown source...

craigds: put it another way: at `apply` time i wouldn't expect the user to know what revision the patch was generated from. So they shouldn't have to specify it on the command line

rcoup: Makes sense, thanks 👍

rcoup: The second issue:

>> The comment here worries me a bit that we're writing patches that look like inserts but are actually edits.
>
> what if we changed the key from "+" to something else for updates in minimal patches? e.g. "-+"?

I'm keen on something instead of `+`, and I'm not excited about `'-+'` (or `'+-'`). I'd probably prefer something from `*$~!@` etc

craigds:
I was torn about this. I realised that probably it'd look a lot cleaner if the `*` deltas were resolved into `-` and `+` up-front, before calling `rs.commit_diff(...)`. At present the `resolve_missing_values_from_rs` is passed all the way down several layers during the commit stage and it _feels_ like the wrong place for it.

However, that wouldn't leave room for the optimisation that the current approach has: if the `base` is the _same_ commit as the target HEAD, we don't have to look at the old blob data at all, which should provide a nice speedup (haven't tested, but I expect ~20-30%). ([That optimisation is here](https://github.com/koordinates/kart/blob/e73110e922e454b4fd70cc1396cfffeb96f57156/kart/rich_base_dataset.py#L393))

Open to suggestions if you think it's not worth the optimisation.

rcoup:
> I was torn about this. I realised that probably it'd look a lot cleaner if the * deltas were resolved into - and + up-front, before calling rs.commit_diff(...). At present the resolve_missing_values_from_rs is passed all the way down several layers during the commit stage and it feels like the wrong place for it.

Fundamentally agree, but not highly concerned about it. Optimisation seems worth it.

What about a `Patch` object to encapsulate it (where Patch either is-a- or has-a- Diff), or somehow folding it into a `Diff` object?

PR #489, opened 2021-10-06
Fix GeoJSON diff regressions in v0.10.1

### LINKED ISSUE #487: Issues with most recent format from kart diff
Looks like diff output format has changed in the latest release. Before, if several layers had changes, the output could be sent to a folder, and kart created several files, one for each layer, with the corresponding changes. Now all changes are written to a single file, with no information about the layer that each feature belongs to. It can maybe be figured out from the attributes schema, but that adds complexity to the workflow, and there can be cases where it not possible to do it (several layers might have the same schema, or the user might be looking at an older diff where the layer had a different schema to its most recent one...)

Would it be possible to add an additional property somewhere in the feature definition, so it can be used to separate the affected features into groups, depending on the layer they belong to?

craigds:
## Description
* Revert `kart diff --output-format=geojson` output handling to the
  pre-0.10.1 behaviour. That is:
  - by default, output goes to stdout
  - if output goes to stdout and there are more than one dataset,
    a usage error occurs (exit code 2) with an error message.
  - `--output=mydirectory` can be used to set a directory into which to
    write output files; one file per dataset.
  - if an existing directory is given, any `*.geojson` files in the
    directory are deleted before creating output files
  - if a path to an existing regular file is given, a usage error
    occurs.

* Remove `kart show --output-format=geojson`. This doesn't work well
  with multiple-dataset commits and was added rather coincidentally
  between 0.10.0 and 0.10.1. We'll remove it for simplicity as it isn't
  very usable; if multiple datasets end up in a single commit, then this
  shows all features together in one FeatureCollection, even if their
  feature schemas are different.

## Related links:

Fixes #487

## Checklist:

### DIFF
--- a/kart/json_diff_writers.py
+++ b/kart/json_diff_writers.py
@@ -262,26 +262,29 @@ class GeojsonDiffWriter(BaseDiffWriter):
 
     @classmethod
     def _check_output_path(cls, repo, output_path):
+        if isinstance(output_path, Path):
+            if output_path.is_file():
+                raise click.BadParameter(
+                    "Output path should be a directory for geojson format.",
+                    param_hint="--output",
+                )
+            if not output_path.exists():
+                output_path.mkdir()
+            else:
+                for p in output_path.glob("*.geojson"):
+                    p.unlink()
+        else:
+            output_path = "-"
         return output_path
 
     def write_diff(self):
-        if isinstance(self.output_path, Path) and self.output_path.is_dir():
-            self.write_file_per_dataset()
-        else:
-            output_obj = {
-                "type": "FeatureCollection",
-                "features": self.filtered_repo_feature_deltas_as_geojson(),
-            }
-
-            dump_json_output(
-                output_obj,
-                self.output_path,
-                json_style=self.json_style,
-            )
-        self.write_warnings_footer()
-
-    def write_file_per_dataset(self):
         has_changes = False
+
+        if len(self.all_ds_paths) > 1 and not isinstance(self.output_path, Path):
+            raise click.BadParameter(
+                "Need to specify a directory for geojson with >1 dataset",
+                param_hint="--output",
+            )
         for ds_path in self.all_ds_paths:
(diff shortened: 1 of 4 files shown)


craigds:
@volaya does this fix the issue for you?

kart/json_diff_writers.py
```diff
@@ -262,26 +262,29 @@ class GeojsonDiffWriter(BaseDiffWriter):
...
-            }
+        has_changes = False
 
-            dump_json_output(
-                output_obj,
-                self.output_path,
-                json_style=self.json_style,
+        if len(self.all_ds_paths) > 1 and not isinstance(self.output_path, Path):
+            raise click.BadParameter(
+                "Need to specify a directory for geojson with >1 dataset",
```
rcoup, line 285:
```suggestion
                "Need to specify a directory via --output-path for GeoJSON with more than one dataset",
```

kart/json_diff_writers.py
```diff
@@ -262,26 +262,29 @@ class GeojsonDiffWriter(BaseDiffWriter):
 
     @classmethod
     def _check_output_path(cls, repo, output_path):
+        if isinstance(output_path, Path):
+            if output_path.is_file():
+                raise click.BadParameter(
+                    "Output path should be a directory for geojson format.",
+                    param_hint="--output",
+                )
```
rcoup, line 270:
```suggestion
                raise click.BadParameter(
                    "Output path should be a directory for GeoJSON format.",
                    param_hint="--output-path",
                )
```

kart/json_diff_writers.py
```diff
@@ -262,26 +262,29 @@ class GeojsonDiffWriter(BaseDiffWriter):
...
+            if not output_path.exists():
+                output_path.mkdir()
+            else:
```
rcoup, line 273:
```suggestion
            else:
                L.debug("Cleaning *.geojson files from %s ...", output_path)
```
(`L` isn't defined)

Issue #496, opened 2021-11-02
Add functionality to get the history of a given feature


volaya:
Just like now you can get a list of commits that affect a given layer, it should be possible to get the list of commits that affect a feature, given its id

craigds:
at the moment this _can_ be done by finding out the internal path of the feature (somehow!) and doing:

`kart log HEAD -- datasetname/.table-dataset/feature/A/A/A/R/kc0EQA==`

... however that's not a real solution. But i would _like_ it to continue working, if possible, as it's a pretty handy feature for developers I think.

A better solution might be `kart log HEAD -- datasetname:<primarykey>`.

Since normal paths above shouldn't ever contain colons (related: #497) the parsing is unambiguous here.

craigds:
This was fixed in #513 and will be in the 0.10.8 release.

Issue #493, opened 2021-10-25
Improve `kart log --help`


craigds:
`kart log` seems a pretty important command but the docs are lacking:

```
$ kart log --help
Usage: kart log [OPTIONS] [ARGS]...

  Show commit logs

Options:
  -o, --output-format [text|json|json-lines]
  --json-style [extracompact|compact|pretty]
                                  How to format the output. Only used with
                                  --output-format=json

  --with-feature-count [veryfast|fast|medium|good|exact]
                                  Adds a 'feature_count' (the number of
                                  features modified in this diff) to JSON
                                  output.If the value is 'exact', the feature
                                  count is exact (this may be slow.)
                                  Otherwise, the feature count will be
                                  approximated with varying levels of
                                  accuracy.

  --help                          Show this message and exit.
```

This is because it's a thin wrapper of `git log`, so we didn't implement any args really, we just pass whatever we're given on to `git log`.

We should probably either:
 - parse the args/options using click, just so they show up properly in the help text.
 - override the help text function so it displays meaningful args even though we're not parsing them

craigds:
Seems like the best option is to stop forwarding unknown arguments to `git log`, and instead parse the whole command line (with a subset of the options git supports), and throw usage errors for unknown options.

Many of the `git log` options make no sense in a kart context, and supporting them just means we can't stop forwarding to git later (which we'll almost certainly want to do.)

PR #498, opened 2021-11-04
Log argument parsing improvements

### LINKED ISSUE #494: `kart log` doesn't handle `--` the same way as git
`kart log` passes args off to `git log`, _except_ that it parses them using click first. Since click swallows the `--` arg, that means it never makes it to git.

So:

* `kart log {datasetname}` --> `git log {datasetname}`
* `kart log -- {datasetname}` invokes `git log {datasetname}`
* `kart log -- -- {datasetname}` invokes `git log -- {datasetname}`

We should fix this.

The only _obvious_ way I can see to fix it is to inspect `sys.argv` before click gets to it, and if there's a `--`, just insert _another_ `--`. But we should probably figure out if there's a click way to do it

craigds:
## Description

Improves parsing of arguments in `kart log`

## Related links:

- Fixes #494 
- This PR is prerequisite work for #496 and #493

## Checklist:

### DIFF
--- a/kart/log.py
+++ b/kart/log.py
@@ -4,6 +4,7 @@ import sys
 import warnings
 
 import click
+import pygit2
 
 from .cli_util import tool_environment
 from .exec import execvp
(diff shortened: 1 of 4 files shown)


kart/log.py
```diff
@@ -51,19 +85,62 @@
...
+    # Handle positional args.
+    # These could be refs, commits, trees, or paths.
+    #  * as soon as we encounter a path, we assume all remaining args are also paths.
+    #  * i.e. the paths must be given *last*.
+    #  * If it's ambiguous whether something is a path or not, we assume it's a commit-ish.
+    #  * If you want to be unambiguous, provide the `--` arg to separate the list of commit-ish-es and paths.
+    # This behaviour is consistent with git's behaviour.
+    try:
+        dash_index = args.index("--")
+    except ValueError:
```
olsen232, line 99: Maybe it's pythonic how it is but I don't love try-except-else for readability - I kinda prefer:

try:
  dash_index = ...
except ValueError:
  dash_index = None
  
  if dash_index is None:
    ...

kart/log.py
```diff
@@ -51,19 +85,62 @@
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def log(ctx, output_format, json_style, do_dataset_changes, with_feature_count, args):
     """ Show commit logs """
+    repo = ctx.obj.get_repo(allowed_states=KartRepoState.ALL_STATES)
+
+    # Handle positional args.
```
olsen232, line 90: Arg parsing could probably have it's own function. No harm in adding more functions, makes code easier to read.

olsen232 (approved):

kart/log.py
craigds: I prefer the `else:` myself.

But now that I'm thinking about it, I actually prefer this to both of those:

```python

    if "--" in args:
        dash_index = args.index("--")
        paths = args[dash_index + 1 :]
        other_args = args[:dash_index]
    else:
        ...
```

PR #501, opened 2021-11-09
dataset paths: Forbid certain chars & components

### LINKED ISSUE #497: Document and enforce what characters can be used in dataset names
It's currently possible to import a dataset with a colon (`:`) in its name.

This breaks at least one current feature of kart:

- `kart diff` accepts filters in the format `DATASET[:PRIMARY_KEY]`; since string PKs can contain colons the parsing is ambiguous

We should specify that dataset names don't contain colons, and reject them during import if they do.

craigds:
## Description

Background: There's existing ambiguity caused by the use of `:` in dataset names, when you use either:

* `kart import file.gpkg oldname:something:newname`
* `kart diff --filter=datasetname:something:primarykeyvalue`

This PR enforces a fairly conservative spec for dataset names to avoid problems with `:` as well as future databases, OSes and filesystems we might want to play nicely with.

## Related links:

Fixes #497

## Checklist:

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -1061,6 +1061,10 @@ def test_import_list_formats(data_archive_readonly, cli_runner):
         pytest.param(("",), False, id="empty-string"),
         pytest.param(("a:b",), False, id="special-ascii-chars"),
         pytest.param(("a\nb",), False, id="ascii-control-chars"),
+        pytest.param(("a/b/c",), True, id="slash-separated"),
+        pytest.param(("a/b/",), False, id="trailing-slash"),
+        pytest.param(("a//b",), False, id="empty-component"),
+        pytest.param((r"a\b\c",), True, id="backslash-separated"),
         pytest.param(("1a",), False, id="leading-numeral"),
         pytest.param(("a1",), True, id="trailing-numeral"),
         pytest.param((".a",), False, id="leading-dot"),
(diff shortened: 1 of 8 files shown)


tests/test_init.py
```diff
@@ -1051,3 +1053,60 @@ def test_import_list_formats(data_archive_readonly, cli_runner):
...
+        pytest.param(("a1",), True, id="trailing-numeral"),
+        pytest.param((".a",), False, id="leading-dot"),
+        pytest.param(("a.",), False, id="trailing-dot"),
+        pytest.param(("a.b",), True, id="contains-dot"),
+        pytest.param(("a ",), False, id="trailing-space"),
+        pytest.param(("a b",), True, id="contains-space"),
+        pytest.param(("COM1",), False, id="windows-reserved-filename"),
+        pytest.param(("com1",), False, id="windows-reserved-filename-ignoring-case"),
+        pytest.param(("ⅶ",), False, id="leading-numeral-unicode"),
+        pytest.param(("Ἢⅶ",), True, id="trailing-numeral-unicode"),
```
rcoup, line 1074: Why are trailing numbers a problem?

rcoup (commented):
Do we need to check for empty path parts or does that get normalised elsewhere?

tests/test_init.py
craigds: they're not, this just tests that that's fine (hence the `True` here)

rcoup: Ah, duh

craigds:
> Do we need to check for empty path parts or does that get normalised elsewhere?

I should probably add something to check that.

This reminded me of something - I have no idea what happens if you add backslashes into the path. Do they get split into directory components on Windows? 🤷

rcoup:
I have a vague feeling Git does some path normalisation too. Try it?

craigds:
git fast-import crashed and hung the process when I included empty path components. I've now forbidden that.

Using backslashes works fine in Git (which treats them as equivalent to forward slashes afaict) but crashes some of our code. I've added code to quietly convert them to forward slashes during import so Kart code doesn't have to deal with them.

rcoup:
LGTM, though should check behaviour in Windows works how we expect too.

Test case for leading `/` too? Or is that not a problem because it's not a letter?

craigds:
> Or is that not a problem because it's not a letter?
that

olsen232:
I was logged into windows already so I tested this on windows - tried importing datasets with user-specified names a/b/c and a\b\c. Both forms were normalised to a/b/c, or perhaps more accurately, both resulted in a dataset at tree / subtree / subtree in the git repo.

olsen232 (approved):

PR #502, opened 2021-11-10
Loosen dataset path restrictions

craigds:
## Description
The rules for dataset paths added in #501 were a bit too unwieldy and
hard to follow.
Since we didn't have any clear reason for forbidding initial characters
in dataset paths, we discussed and decided to relax the restriction, so
that now the first character of a dataset path has almost the same
restrictions as the second and later characters. (However, starting with
a `/` or `.` is still forbidden)

Additionally this change fixes an unintentional behaviour in `import`
where a `.` in a dataset path was replaced with a `/`. This was intended
to import schema-qualified database names in a hierarchy to avoid name
conflicts when importing with `--all-tables`.

However we felt that was a misfeature; the user probably doesn't want to
import schema-qualified names, and if they do they can do so explicitly.
So this change removes that behaviour also.
## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
   - `kart create-patch` now supports `--patch-type minimal`, which creates a much-smaller patch; relying on the patch recipient having the HEAD commit in their repository [#482](https://github.com/koordinates/kart/issues/482)
   - `kart apply` now applies both types of patch.
 * `kart log` now accepts a `--` marker to signal that all remaining arguments are dataset names. [#498](https://github.com/koordinates/kart/issues/498)
+* `import` from a Postgres or MSSQL source will no longer prepend the database schema name to the imported dataset path.
 * Bugfix: Diffing between an old commit and the current working copy no longer fails when datasets have been deleted in the intervening commits.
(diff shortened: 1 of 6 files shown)


olsen232:
The code that causes a fully-qualified[1] table name to sometimes arrive as the default dataset path is still intact - this happens whenever someone tries to import from a sqlalchemy datasource without specifying the the schema-name as part of the datasource, which means they have to specify a schema-name + table-name for each table they wish to import.
The code previously imported such tables to a dataset called `schema-name/table-name`, now it will import them to `schema-name.table-name`.
If you like, you can punt this for a bit - or I can take it, since I put together the sqlalchemy import source in the first place and so it's my fault this bit of it is a bit esoteric.
 
[1] By fully-qualified I mean things like `schema-name.table-name` - or for MySQL, `db-name.table-name`. Whether this is actually fully qualified is up for debate.

craigds:
> The code that causes a fully-qualified[1] table name to sometimes arrive as the default dataset path is still intact

isn't that [this change](https://github.com/koordinates/kart/pull/502/files#diff-30ee65974df04dc5209fa0c5d85cbabb846f636a57bb212608082b7766545397R154) ?

I confess I didn't actually _try it_ ;) but that's what that code seems to do

olsen232 (approved):

PR #503, opened 2021-11-12
Handle features that are missing due to spatial filters

olsen232:
Handles two different scenarios:
1. While trying to generate a WC diff from the tracking table,
  the base features of dirty features are missing from the dataset.
  In this scenario, immediately fetches all dirty features of that dataset.
2. While trying to stream a diff (feature blobs are loaded and output just
  in time) finds that some blobs are missing from the dataset. Kicks off
  a fetch of each missing blob, and comes back to them once all other
  diffs for that dataset have been output.

Together, this should mean that any diff can be generated, regardless
of which features are missing before hand.

Also added test data which is already spatially indexed: polygons-with-s2index.tgz

See:

Some spatial filter TODOs (see linked issue for more) -
- custom git, libgit2, pygit2 not yet built on windows, so, windows can't do a spatial filtered clone, or tell why features are missing
- kart merge doesn't support spatial filters yet - it will need similar code to handle missing features


kart/working_copy/base.py
```diff
@@ -694,6 +695,21 @@ def diff_db_to_tree_feature(
...
+            # There is no such feature. This is okay: it just means the user has inserted
+            # the feature into the working copy and it has not yet been committed.
+            return None
+        elif subcode == LibgitSubcode.EOBJECTPROMISED:
+            # There is such a feature, but we don't have it locally right now. Fetch it.
+            dataset.fetch_missing_dirty_features(self)
+            return dataset.get_feature(feature_pk)
+        else:
+            # Some other error has happened, or no subcode was found. Re-raise the error.
+            raise feature_error
```
craigds, line 711: This will cause the traceback to show `_fetch_if_promised` rather than `dataset.get_feature`, which might be confusing. Just use bare `raise` here to preserve the original traceback.

However, since it's confusing for a function to use bare `raise` when it doesn't appear to ahve a try/except, move that into the function too, and call it `get_feature_and_fetch_if_needed` or something.

craigds (approved):

kart/working_copy/base.py
olsen232: Done

kart/promisor_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        "--filter=blob:none",
+        "--stdin",
+    ]
+
+    p = subprocess.Popen(
+        cmd,
+        cwd=repo.path,
+        stdin=subprocess.PIPE,
+        env=tool_environment(),
+        bufsize=128 * 1024,
```
rcoup, line 61: do we want this? I mean, aren't we basically line-buffering wrt IO here?

Ala git writes to the object store, we're just telling it what to fetch?

kart/spatial_filters.py
```diff
@@ -333,6 +335,25 @@ def split_file(cls, contents):
...
+    # Doesn't match any spatial filter because it is null / None / non-existent:
+    IS_NONE = auto()
+    # Isn't present locally, but is promised. From this we infer it doesn't match the repo's current spatial filter:
+    IS_PROMISED = auto()
+    # Exists and is present and does NOT match the given spatial filter:
+    NON_MATCHING = auto()
+    # Exists and is present and DOES match the given spatial filter.
+    # This includes the case where a feature matches the "match-all" spatial-filter, which matches everything,
+    # and it includes the case where a feature has no geometry (and so matches all spatial filters).
+    MATCHING = auto()
```
rcoup, line 351: these are active voice, the promised is passive... maybe `NOT_MATCHED` & `MATCHED`?

Alternatively I guess non-matching could be is-outside/excluded/is-out?

kart/spatial_filters.py
```diff
@@ -333,6 +335,25 @@ def split_file(cls, contents):
         return [p.strip() for p in parts]
 
 
+class MatchResult(Enum):
+    """Stores a bit more detail about whether a feature matches or doesn't match the spatial filter."""
+
+    # The given feature...
+    # Doesn't match any spatial filter because it is null / None / non-existent:
+    IS_NONE = auto()
```
rcoup, line 343: None is a Python-ey term... should we be using null for data values? Not a big deal if it's something we're doing already

tests/test_spatial_filters.py
```diff
@@ -161,7 +186,98 @@ def test_init_with_invalid_spatial_filter(cli_runner, tmp_path):
     assert not repo_path.exists()
 
 
+@pytest.mark.skipif(is_windows, reason=SKIP_REASON)
```
rcoup, line 189: can we feature-detect libgit2 (& git) and skip based on that?

tests/test_spatial_filters.py
```diff
@@ -161,7 +186,98 @@ def test_init_with_invalid_spatial_filter(cli_runner, tmp_path):
...
+            assert H.row_count(sess, H.POLYGONS.LAYER) == 44
+
+        # However, the entire polygons layer was cloned.
+        # Spatial filters are currently only applied locally... all features are still present.
+        assert local_features(repo2.datasets()[H.POLYGONS.LAYER]) == H.POLYGONS.ROWCOUNT
+
+        # Unless you use an experimental environment variable:
+        # TODO: Always apply spatial filters during the clone, then clean up this test.
+        try:
+            os.environ["X_KART_SPATIAL_FILTERED_CLONE"] = "1"
```
rcoup, line 223: if this was outside the try then you wouldn't need an `if` in the finally...

tests/test_spatial_filters.py
```diff
@@ -161,7 +186,98 @@ def test_init_with_invalid_spatial_filter(cli_runner, tmp_path):
...
+                for pk in H.POLYGONS.SAMPLE_PKS:
+                    if not is_local_feature(ds, pk):
+                        insert(sess, with_pk=pk, commit=False)
+
+            r = cli_runner.invoke(["-C", repo3_path, "status"])
+            assert r.exit_code == 0, r.stderr
+            # TODO - label pk-conflicts as being different to updates in kart status
+            assert "6 updates" in r.stdout
+            # All of the 6 featues that are conflicts / were "updated" in the WC have been loaded:
+            assert local_features(ds) == 58
```
rcoup, line 255: is there a (sane) way to test the number of objects passed to git to fetch on demand?

rcoup (commented):
> custom git, libgit2, pygit2

can we get links to the matching changes?

olsen232:
Links to current koordinates fork of git / libgit2 / pygit2:
git: https://github.com/koordinates/git/commit/e1645ec76fb513918e35204a65d94524682b1445
libgit2: https://github.com/koordinates/libgit2/commit/bb69c8fa3d30c38e409771e388787d5ea38b7723
pygit2: https://github.com/koordinates/pygit2/commit/3443cd8fd2cead2de608f51ccb3cdee5167c1fd4

I was going to tag those commits at some point, but haven't yet - not sure if they'll change again at some point

PR #505, opened 2021-11-16
Split CI Jobs

rcoup:
## Description

Split the posix CI job into a macOS and a Linux one.

Loads of conditionals made it hard to read & reason about, and it also makes adding additional jobs (CMake/etc) more straightforward.

Attempt to fix some of the macOS vendor cache issues at the same time, primarily by adding the packages & versions installed from the vendor `Brewfile` into the vendor-dist cache key.

To run CI only on one platform is now: `[ci only linux]`; `[ci only macos]`; `[ci only windows]`

## Checklist:

- [x] ~Have you included test(s)?~ lots of existing tests
- [x] ~Have you updated the changelog?~ there's no user-visible changes here

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,11 +119,7 @@ jobs:
         if: steps.cache-dist.outputs.cache-hit != 'true'
         with:
           path: .ccache
-          key: vendor-ccache-Linux-${{ hashFiles('vendor/**/Makefile') }}-${{ hashFiles('vendor/build-manylinux.sh') }}-${{ github.sha }}
-          restore-keys: |
-            vendor-ccache-Linux-${{ hashFiles('vendor/**/Makefile') }}-${{ hashFiles('vendor/build-manylinux.sh') }}-${{ github.sha }}
-            vendor-ccache-Linux-${{ hashFiles('vendor/**/Makefile') }}-${{ hashFiles('vendor/build-manylinux.sh') }}-
-            vendor-ccache-Linux-
+          key: vendor-ccache-Linux
 
       - name: "vendor: build"
         if: steps.cache-dist.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -304,37 +210,28 @@ jobs:
 
       - name: "package: cleanup"
         run: |
-          sudo find vendor/dist -mindepth 1 -maxdepth 1 ! -name "vendor-*.tar.gz" -exec rm -rf {} \;
+          sudo find vendor/dist -mindepth 1 -maxdepth 1 ! -name "vendor-Linux.tar.gz" -exec rm -rf {} \;
```
olsen232, line 213: Is this the same as 
`rm -rf vendor/dist/vendor-Linux.tar.gz`

.github/workflows/build.yml
```diff
@@ -61,126 +25,110 @@ jobs:
...
-        run: |
-          mkdir -p .cache/pydist
-          wget -nv ${{ env.PY3_URL }} -O .cache/pydist/${{ env.PY3_PKG }}
-
       - name: "python: pip cache"
         uses: actions/cache@v1
         with:
           path: .cache/pip
-          key: pip-${{ runner.os }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements/*.txt') }}-${{ hashFiles('Makefile') }}
+          key: pip-Linux-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements/*.txt') }}-${{ hashFiles('Makefile') }}
```
olsen232, line 71: FYI hashFiles has been updated and now takes multiple glob arguments so that:
`${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements/*.txt') }}`
is equivalent in some sense to:
`${{ hashFiles('requirements.txt', 'requirements/*.txt') }}`
Lots more opportunities in this file to do this.
This could be as good a time as any to tidy that up, although note that it will effectively clear all caches.

olsen232 (approved):
I'm gonna miss the little 🍎 🐧

.github/workflows/build.yml
rcoup: yes it is 😁  Nice spotting

.github/workflows/build.yml
rcoup: ah, ace. Willdo.

.github/workflows/build.yml
rcoup: ah, it finds everything _other_ than the `vendor-{OS}.tar.gz` archive and deletes it.

PR #511, opened 2021-11-23
Development builds using CMake

rcoup:
## Description

Initial adoption of [CMake](https://cmake.org) for improved developer builds. Build instructions are in `CONTRIBUTING.md`.

* on macOS & Linux, CMake picks up system-installed dependencies from Homebrew or otherwise, via normal CMake methods (eg. `-DMyDependency_ROOT=...`)
* this includes newer versions of Python and other library dependencies, Visual Studio versions, etc.
* all Python wheels (including S2) are built on macOS & Linux
* Windows only uses a (manually) downloaded vendor archive from existing CI at the moment.
* macOS & Linux can do that as well to simplify things
* Kart now initialises paths using a `_kart_env` global module
* features are detected in tests a bit better (eg. spatial filter support in Git)
* builds run in CI, currently if the CMake jobs error the overall build will still pass
* for CI, tests are wrapped via CTest, which handles sharding/etc configuration (just invokes pytest though)
* out-of-tree builds are supported out of the box with CMake, and so we now support that too — the entire build process lives in the build tree (`./build/` by default)
* pre-commit adds linters/formatters for CMake and Github Actions

These are developer builds, so they're not relocatable, and currently there's no install step either (you need to link to or run  `./kart` in the build tree). Obviously most of the vendor dependencies aren't built, and there's no packaging yet. These are all follow up steps.

* the first 5x commits are the bulk of the CMake integration
* the xdist, proj, and conditional skipping fixes are generic
* and the remaining fixes are mostly getting builds and tests working under CI

## Checklist:


vendor/cmake/FindODBC.cmake
```diff
@@ -0,0 +1,261 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying file Copyright.txt or
+# https://cmake.org/licensing for details.
+
+# From: GDAL
+
+#[=======================================================================[.rst:
+FindODBC
```
craigds, line 7: Does this differ from https://cmake.org/cmake/help/latest/module/FindODBC.html ? At first glance it looks very similar

CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+    set(CMAKE_${lang}_COMPILER_LAUNCHER ${CMAKE_COMMAND} -E env ${ccacheEnv} ${CCACHE_PROGRAM})
+  endforeach()
+endif()
+
+# OSX: prefer Homebrew over Frameworks
+set(Python3_FIND_IMPLEMENTATIONS "CPython")
+find_package(Python3 3.7 REQUIRED COMPONENTS Interpreter)
+pythongetabiinfo()
+
+# version configure_file(${KART_CMAKE_TEMPLATE_DIR}/VERSION.in kart/VERSION @ONLY)
```
craigds, line 58: accidentally commented out?

CONTRIBUTING.md
```diff
@@ -2,7 +2,156 @@
...
+$ git clone https://github.com/koordinates/kart.git
+$ cd kart
+```
+
+### Installing dependencies on macOS
+
+If you're a Homebrew user, you can get all of those via:
+
+```console
+$ brew install --upgrade cmake gdal git openssl@1.1 libpq python@3.7 \
```
craigds, line 40:
```
Error: invalid option: --upgrade
```

I _think_ `install` always upgrades, so that option is probably redundant

CONTRIBUTING.md
```diff
@@ -2,7 +2,156 @@
...
+    spatialindex libspatialite sqlite3 swig unixodbc
+```
+
+Then configure Kart:
+```
+$ cmake -B build -S .
```
craigds, line 46: I think the MacOS instructions are missing the bit about installing a custom libgit2 (it's below in the Linux section, probably just copy/move it):

```
  Could NOT find LibGit2 (missing: LibGit2_LIBRARY LibGit2_INCLUDE_DIR)
  (Required is at least version "1.3.0")
```

If I additionally `brew install libgit2` and try again I get this:

```
-- Found LibGit2: /usr/local/lib/libgit2.dylib (found suitable version "1.3.0", minimum required is "1.3.0")
CMake Error at vendor/CMakeLists.txt:76 (message):
  Found LibGit2 v1.3.0, but git_error_subcode isn't defined.  Build & install
  from https://github.com/koordinates/libgit2/tree/kx-latest
```

craigds: actually I get that latter error _even if_ I follow the linux install instructions for our custom libgit2. So something else missing I guess

vendor/cmake/FindODBC.cmake
rcoup: Good spotting — GDALs targetting older CMake, so we can probably just use upstreams. Will check/try

CMakeLists.txt
rcoup: ah, I'll revisit it at some point. Will remove for now.

CONTRIBUTING.md
rcoup: > I think the MacOS instructions are missing the bit about installing a custom libgit2 (it's below in the Linux section, probably just copy/move it):

Ah, yeah, decided it was better not to use the Homebrew one in the end.

> If I additionally brew install libgit2 and try again I get this:

Which is expected

> actually I get that latter error even if I follow the linux install instructions for our custom libgit2. So something else missing I guess

1. Did you `brew unlink libgit2`?
2. Does deleting `build/CMakeCache.txt` help? If it found something CMake caches it, so it'll reuse the cache instead of finding it again

rcoup:
@olsen232 I have a conflict in [`test_git_spatial_filter_extension()`](https://github.com/koordinates/kart/pull/511/conflicts)...
* In general I think feature detection is better than `if_some_os:` tests if possible
* Git won't always have filter-extension support, since we might be using not-our Git c/- CMake
* I added feature-detection for filter-extension support in [`git_supports_filter_extensions()`](https://github.com/koordinates/kart/blob/cmake-2/tests/test_spatial_filters.py#L22)
* Wrapped in a [`@git_with_filter_extension_support` fixture](https://github.com/koordinates/kart/blob/cmake-2/tests/test_spatial_filters.py#L60)
* In CI, the fixture asserts if `$KART_EXPECT_GITNOFILTEREXTENSION` isn't set
* Outside CI it skips, which skips any tests that use it.

-> So maybe this test can just go away? Or should it change into a similar feature-detection + CI-check `@git_with_spatial_filter_extension` fixture?

olsen232:
I plan to look through this (or at least try it out) today - haven't got to it yet sorry.

This test - background:
When I wrote that test, it was testing if git was willing to apply `--filter=extension:spatial=1,2,3,4` to a repository. On macos and linux it will, on windows it won't since it only knows about things like `--filter=blob:none`.
The change that I've made that is conflicting with your change is that the windows now recognises filter extensions, but doesn't have any compiled in, since I haven't done the work to link in the filter extension along with libs2 and sqlite on windows. This is necessary and sufficient for windows to be able to do partial clones from remotes that support the extension.
So, now I suppose this test is actually two tests, and should probably be broken into two - on linux and macos, I'm testing that we have the spatial filter extension, and on windows I'm testing that git understands the concept of filter extensions (but doesn't have that one).
For now you can keep your change and revert mine - then it will be the test that tests if we have the filter extension, and should pass on linux and macos. One of us can write another similar test that tests if git understands filter extensions at all. That should pass on all platforms (but would fail, of course, if someone builds kart but keeps a standard git).

rcoup:
On any platform the relevant tests should skip if someone doesn't have a Git with filter extension support, or doesn't have a spatial filter extension... whatever the specific test needs.

Except in CI which we know the environment & can set `KART_EXPECT_*`

olsen232:
Sure I'm on board.
- No platform specific tests - instead, we just detect which features are working or are not working.
- Unless the tester has specifically said they expect this feature to be working using an environment variable.
- Which we will do on CI, since, we don't want our linux build to regress to the level of our windows build without it causing a red build.

The code you have now does something like that - so you can resolve the conflict just by accepting your change over mine, since your change is much closer to what we want.

Once this is merged, one or other of us can do a little bit more work on it:
We can add something similar to what you have here for detecting if the spatial filter filter extension is present, since there's now 3 levels of git we might have:
1. regular git
2. git which understands filter extensions, but doesn't have any built in
3. same as (2) but also has the spatial-filter-extension built in

I had another comment about way you've structured environment variables, but I've added that elsewhere

rcoup:
ok, I made the changes (I think) in 5c5e96c — there's two feature-detections now.

> I had another comment about way you've structured environment variables, but I've added that elsewhere

Can't see it?

olsen232:
I have pending comments that I failed to publish - maybe this comment will publish them, otherwise, I'll have to go find the right button

olsen232:
Yeah I'm not sure I can publish them anymore. I'll just paste them here:

Regarding `git_with_filter_extension_support:`
Here's another way of structuring these environment variables that makes more sense to me:

```
git_supports_filter_extensions = git_supports_filter_extensions()
if "KART_EXPECT_GITFILTEREXTENSION" in os.environ:
  assert git_supports_filter_extensions == bool(os.environ["KART_EXPECT_GITFILTEREXTENSION"])
if not git_supports_filter_extensions:
  raise pytest.skip("The in-use git doesn't support filter extensions")
```
Reasoning: one less variable to consult, easier to test it locally without having to pretend to be CI - if the user thinks that GITFILTEREXTENSION should work, then, let them signal that, and we'll test it. And, it lets you assert that it's not supported, if that's what you are expecting. (It would be strange to be wrong about that, but you'd probably want to know)

Regarding libgit2 and CONTRIBUTING.md:
Taking these instructions literally means you end up with a libgit2 subdirectory right under the kart repo root.
- is this where it should go, or should we recommend the user put it somewhere else?
- should we add it to the .gitignore?


I noted that I followed the instructions and that they worked for me, and tests pass.
But regarding testing: 
Tests all passed, but note that I got this message:
1: SKIPPED [3] tests/test_spatial_filters.py:67: The in-use git doesn't support filter extensions
which is fine if that's something you haven't go to yet, or maybe it's been fixed since?


Regarding `ctest -V`
I found you can still run the tests just by running:
./build/venv/bin/pytest from the repo root. This currently has the following advantages:
- coloured output (which I find is pretty useful for something like this with pages of text, maybe the odd failure amongst many successes)
- all the pytest options work. Not sure how to make them work for ctest yet.
Anyway just FYI - we should do some work on ctest at some point to remove these disadvantages, or just use pytest directly

rcoup:
> Regarding libgit2 and CONTRIBUTING.md:
> Taking these instructions literally means you end up with a libgit2 subdirectory right under the kart repo root.
> * is this where it should go, or should we recommend the user put it somewhere else?
> * should we add it to the .gitignore?

no to both

rcoup:
> I found you can still run the tests just by running: ./build/venv/bin/pytest from the repo root.

yeah, ctest literally runs pytest.

There's no advantage to using ctest for local dev, I'll reword it. Though we need to make the test suite robust on Windows without sharding stuff.

rcoup:
> I noted that I followed the instructions and that they worked for me, and tests pass.
> But regarding testing:
> Tests all passed, but note that I got this message:
> 1: SKIPPED [3] tests/test_spatial_filters.py:67: The in-use git doesn't support filter extensions
> which is fine if that's something you haven't go to yet, or maybe it's been fixed since?

Platform? What git did you build with/against?

rcoup:
> Here's another way of structuring these environment variables that makes more sense to me:

Sure, sounds good.

rcoup:
feature-flag stuff done in 5b799b3

PR #513, opened 2021-11-29
Add support for kart log of only certain features

olsen232:
Supported previously but only if you knew the exact path of the
feature you were interested in, which is generally hidden from the
end user. Now works with the same syntax that diff and commit accept:
`<dataset>:feature:<feature-primary-key>` or simply
`<dataset>:<feature-primary-key>`

https://github.com/koordinates/kart/issues/496

## Checklist:

### DIFF
--- a/kart/log.py
+++ b/kart/log.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timezone, timedelta
+from enum import Enum, auto
 import re
 import subprocess
 import sys
@@ -10,6 +11,7 @@ import pygit2
 from .cli_util import tool_environment
 from .exec import execvp
 from .exceptions import SubprocessError
+from .key_filters import RepoKeyFilter
 from .output_util import dump_json_output
 from .repo import KartRepoState
 from .timestamps import datetime_to_iso8601_utc, timedelta_to_iso8601_tz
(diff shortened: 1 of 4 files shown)


kart/log.py
```diff
@@ -72,65 +118,110 @@ def parse_extra_args(
...
+            capture_output=True,
+            env=tool_environment(),
+        )
+    except subprocess.CalledProcessError as e:
+        raise SubprocessError(
+            f"There was a problem with git log: {e}", called_process_error=e
+        )
+    if r.stdout:
+        try:
+            return repo.datasets(r.stdout.strip())[ds_path]
```
craigds, line 188: The log command above will find the last commit in which the given dataset was modified.

If that commit actually deleted the dataset, I think this will raise a KeyError (which is ignored here), and we'll return None, which is handled as if the dataset didn't exist in the entire log.

Instead we should probably use the dataset from the parent commit in that case.

craigds (changes requested):
> `<dataset>:feature:<feature-primary-key>` or simply
`<dataset>:<feature-primary-key>`

what if you have a string PK and one of the values is `feature:abc` and another is `abc`? I would favour removing one of these syntaxes to avoid this ambiguity.

rcoup:
> what if you have a string PK and one of the values is `feature:abc` and another is `abc`? I would favour removing one of these syntaxes to avoid this ambiguity.

As long as the ambiguity is resolve-able, is it that big a deal? 

```console
$ kart log mydataset:12345
$ kart log mydataset:abc
$ kart log mydataset:feature:feature:abc
$ kart log mydataset:AAAC_7211_0033
$ kart log mydataset:tile:AAAC_7211_0033
```

kart/log.py
olsen232: Handled this case and added a test.

olsen232:
I didn't fix the ambiguity where token "feature" may or may not occur in the pattern for a feature.
As Rob shows, you can still unambiguously refer to such features.

Removing the longer syntax and only having the shorter and easier syntax would actually make this worse: `dataset:meta:schema.json` is a valid pattern for a meta item right now, and `dataset:feature:meta:schema.json` is a confusing pattern for a feature with primary key `meta:schema.json` (may it never exist). But if we get rid of the `:feature:` token then there's no way to refer to this feature. So that leaves only the option of getting rid of the `dataset:123` syntax, which I think people will sometimes type by accident anyway so I prefer to keep supporting it.

craigds:
> you can still unambiguously refer to such features.

Are you saying that using the `<dataset>:feature:<pk>` syntax can always work unambiguously? But the `<dataset>:<pk>` syntax cannot:

* `kart log dataset:feature:123` matches PK=`123` but presumably _not_ PK=`feature:123`
* `kart log dataset:feature:feature:123` matches PK=`feature:123`
* `kart log dataset:123` matches PK=`123`

> Removing the longer syntax and only having the shorter and easier syntax would actually make this worse

yes, I'd go the other way and remove the shorter syntax instead.

rcoup:
> yes, I'd go the other way and remove the shorter syntax instead.

To me `mydataset:123` is obvious what it's doing and the extra `:feature:` is just noise 99.9% of the time. If you're interacting with Kart programmatically — sure, use the long format.

Issue #521, opened 2021-12-15
During import with --replace-existing, --replace-ids option doesn't work if primary key column is changing.


olsen232:
Specifically, the primary key column should be changing type - if the primary key column keeps the same type, this is treated as a rename, which is allowed. However if the primary key changes type altogether, then this means all existing features must be deleted, and all new features must be inserted using the new primary key type. In such a situation, it makes no sense to specify feature-ids, which limits which features are replaced by the new ones. 

Currently, specifying `--replace-ids` causes the repository to become corrupted, in that it allows old features to continue to exist even though they have no value for the new primary-key column. Trying to create a working copy fails since it tries to create features with NULL primary keys.

The fix is to not allow `--replace-ids` to be specified if the primary key type is changing.

PR #522, opened 2021-12-15
Fix for #521 - no replace-ids when PK changes type

olsen232:
See https://github.com/koordinates/kart/issues/521 for details

## Description

`--replace-ids` option isn't useful when a primary key type is being changed, and right now if you try to use it anyway, it corrupts your dataset. (If this has happened, you can put the dataset back to an earlier state using `kart reset`). See issue link for more details.
Lower impact than it sounds since very few users are using --replace-ids at this early stage...

## Related links:

https://github.com/koordinates/kart/issues/521

## Checklist:

### DIFF
--- a/kart/init.py
+++ b/kart/init.py
@@ -87,6 +87,13 @@ def get_default_num_processes():
     return max(1, int(math.ceil(num_processes)))
 
 
+def any_at_all(iterable):
+    # Returns True if anything exists in iterable - even falsey values like None or 0.
+    # Advances the iterable by one if non-empty, so is best used like so:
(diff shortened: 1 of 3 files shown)


kart/init.py
```diff
@@ -87,6 +87,13 @@ def get_default_num_processes():
     return max(1, int(math.ceil(num_processes)))
 
 
+def any_at_all(iterable):
```
rcoup, line 90: So this is a bit like:

```python
empty = object()
return (next(iterable, empty) is not empty)
```

Nice 😎

rcoup (approved):
LGTM 🚢

craigds (approved):

Issue #514, opened 2021-11-30
Add option to diff command to get the meta changes in all layers


volaya:
Now the meta changes for a given layer can be obtained using

`kart diff refa...refb datasetname:meta -ojson`

But there is no way of showing exisiting meta changes for all layers modified between the selected commits.

A bit of context for this: In the plugin, we are trying to detect whether there are meta changes or no before opening elements such as the diff viewer, which for now assume that the schema doesnt change

craigds:
Obvious way to implement this would be:

```
kart diff refa...refb *:meta -ojson
```

Since we [forbid the * character](https://github.com/koordinates/kart/blob/master/docs/DATASETS_v3.md#valid-dataset-names) in dataset names, this results in unambiguous syntax.

The only obvious problem is that `*` will be expanded by shells to a filename, if a file matching that pattern exists in the current directory. That seems pretty unlikely though and it can always be quoted, so probably not a big deal.

- If we do this we should make it work for `*:feature` too
- Also `*:<featureID>` - this seems easy enough to make work, and means for a single-layer repo, the user doesn't have to use the actual dataset name, which they may not remember.

Issue #219, opened 2020-08-20
PK filters only work for features, not for meta items


olsen232:
`sno diff datasetA` show all diffs in dataset A
`sno diff datasetA:123` shows the diff for a particular feature (plus any meta diffs for datasetA)
`sno commit datasetA` commit all diffs in dataset A
`sno commit datasetA:123` commit the diff for a particular feature (plus any meta diffs for datasetA)

So filters always include meta items. There's no way to use the filters to exclude meta items, or to contain exclusively meta items. If there was it would look something like this:

`sno X datasetA` matches the entire dataset
`sno X datasetA:meta` matches only meta items
`sno X datasetA:feature` matches only features
`sno X datasetA:meta:title` matches a particular meta item
`sno X datasetA:feature:123` or `datasetA:123` matches a particular feature item (doesn't match meta items)

There is one area of this that would need some extra care taken - committing a feature change while there are still uncommitted schema changes for that schema should probably not be allowed. Until the current schema is committed, committing a feature which uses that schema isn't guaranteed to store all of its fields, which is surprising and probably bad.

craigds:
I think my PR addresses all of this. @olsen232 concur?

PR #532, opened 2021-12-24
Support wildcards and meta items in diff/commit filters

craigds:
## Description

This PR adds support for the following filter styles:

* `kart diff HEAD^ datasetname:meta:title` - diff a meta item
* `kart diff HEAD^ datasetname:meta` - diff all meta items from datasetname
* `kart diff HEAD^ *:meta` - diff all meta items from all datasets
* `kart diff HEAD^ *:feature` - diff all features from all datasets
* `kart diff HEAD^ *abc*:feature` - diff all features from datasets whose names contain `abc`

In addition specific-feature filters no longer also show meta items.

### Todo:

- ~~[ ] Prevent user from committing a feature without an associated schema change (`kart commit '*:feature'` when a schema change is also present, as mentioned in #219 )~~
- [x] implement or disable wildcard filters for `log` (it's hard)

## Related links:

* #219 
* #514

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.10.8 (UNRELEASED)
 
+* More advanced filters for `log`, `diff` and `commit`: All of these now work:
+  - Wildcard (`*`) filters for dataset names, e.g. `kart diff -- *parcel*:meta:schema.json` will show only schema changes for all datasets with `parcel` in their names. `*` by itself matches all datasets. [#532](https://github.com/koordinates/kart/issues/532)
+  - You can now output the history of individual features: `kart log -- <dataset-name>:feature:<feature-primary-key>`. [#496](https://github.com/koordinates/kart/issues/496)
 * Simpler developer builds using CMake, see the [contributing notes](./CONTRIBUTING.md).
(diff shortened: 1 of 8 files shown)


olsen232 (approved):
Nice work!

craigds:
>  Prevent user from committing a feature without an associated schema change (kart commit '*:feature' when a schema change is also present

This in fact turns out to not be a problem since Kart actually commits your meta items whenever you commit any feature. This might be improved in future to be less implicit, but for now it seems okay and doesn't need to explicitly prevent you from committing a feature when you have a local schema change.

craigds:
I disabled the specific case of wildcard dataset name with feature IDs (e.g. `*:feature:123`), because it's quite tricky to implement. Since `kart log` delegates to `git log`, supporting this would mean we would need to figure out the full paths of features affected by the pathspecs in advance of calling git log. But we'd need to look at the associated schemas for those features to figure out the mapping from PK to feature path, and the mapping may change over time...

In short, it's very difficult unless we stop calling `git log` and implement log ourselves, so I just disabled it:

```
$ kart log -- *:feature:0008240571
Error: `kart log` doesn't currently support filters with both wildcards and feature IDs
```

I don't expect anyone will notice for a good while, in which time maybe we'll implement log ourselves

olsen232 (approved):

PR #543, opened 2022-01-16
`diff -o json-lines`: Add feature count estimate

craigds:
## Description

With new `--add-feature-count-estimate=<accuracy>` option, adds a line
to the output stating the number of features expected to be in
the diff. This should be useful for showing a progress bar when the diff is
large.

For small diffs it may never appear, as the estimate can easily take
longer to calculate than the whole diff anyway.

## Checklist:

### DIFF
--- a/kart/json_diff_writers.py
+++ b/kart/json_diff_writers.py
@@ -1,11 +1,13 @@
 from datetime import datetime, timezone, timedelta
 import json
 import logging
+import threading
 from pathlib import Path
 
 import click
 
 from .base_diff_writer import BaseDiffWriter
+from .diff_estimation import estimate_diff_feature_counts
 from .diff_structs import DatasetDiff
 from .feature_output import feature_as_json, feature_as_geojson
 from .log import commit_obj_to_json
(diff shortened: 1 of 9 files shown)


kart/json_diff_writers.py
```diff
@@ -226,9 +231,37 @@ def write_header(self):
                 "outputFormat": "JSONL+hexwkb",
             }
         )
+        if self._diff_estimate_accuracy is not None:
+            t = threading.Thread(
+                target=self._calculate_and_feature_count_estimate,
+                # don't block process exit.
+                daemon=True,
```
craigds, line 238: Not sure this is safe; e.g. it might cause the annotations database to get corrupted or locked. second opinion @olsen232 ? Using a non-daemon thread is safer but we'll need to build some kind of alerting to kill the thread on process exit :/

UberMouse (commented):
Assuming it works as advertised (the feature count message appears in stream) this looks like it will give me the info I need :+1:

kart/json_diff_writers.py
craigds: Removed this in favour of a `threading.Event` and the thread terminating itself.

craigds:
needs adding to the changelog but I see 0.11.0 RCs have been produced so perhaps we can just hold off merging this until 0.11.0 is released and then add it in after that

PR #548, opened 2022-01-20
Two fixes for GPKGs with more than 1 XML metadata

olsen232:
Firstly, output a proper error message instead of
TypeError: a bytes-like object is required, not 'list'

Secondly, automatically discard one piece of XML metadata in the case
that both pieces of XML metadata are functionally identical, but one
is original and the other wraps the original in the following manner:

```xml
<GDALMultiDomainMetadata>
  <Metadata>
    <MDI key="some_key">
	[ESCAPED_ORIGINAL_XML]
    </MDI>
  </Metadata>
</GDALMultiDomainMetadata>
```

## Checklist:

### DIFF
--- a/kart/sqlalchemy/adapter/gpkg.py
+++ b/kart/sqlalchemy/adapter/gpkg.py
@@ -512,6 +512,10 @@ class KartAdapter_GPKG(BaseKartAdapter, Db_GPKG):
         if len(xml_list) == 1:
             return xml_list[0]
 
+        sole_useful_xml = cls.find_sole_useful_xml(xml_list)
+        if sole_useful_xml is not None:
+            return sole_useful_xml
+
         # We can't actually commit a whole list of XML, but we need to return something that makes sense.
         # Simply throwing an error here stops dirty-detection working, and stops commands that would fix the situation
         # from working, like `kart reset --discard-changes` or `kart create-workingcopy --discard-changes`.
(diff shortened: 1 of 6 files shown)


kart/sqlalchemy/adapter/gpkg.py
```diff
@@ -529,6 +533,33 @@ def gpkg_to_xml_metadata(cls, gpkg_meta_items):
...
+        # Currently only the following heuristic is implemented:
+        # If there are two pieces of XML and the second piece of XML is the same as the first XML, but wrapped inside a
+        # `<GDALMultiDomainMetadata>` structure, then we just return the first piece of XML (or vice versa).
+        # In all other cases we have no opinion on which piece(s) of XML are useful or useless.
+        if len(xml_list) != 2:
+            return None
+        from html import unescape
+
+        for first, second in ((xml_list[0], xml_list[1]), (xml_list[1], xml_list[0])):
+            wrapper = unescape(second).split(first, maxsplit=1)
```
craigds, line 551: Wouldn't it be just as simple to use an XML parser? H̸̡̪̯ͨ͊̽̅̾̎Ȩ̬̩̾͛ͪ̈́̀́͘ ̶̧̨̱̹̭̯ͧ̾ͬC̷̙̲̝͖ͭ̏ͥͮ͟Oͮ͏̮̪̝͍M̲̖͊̒ͪͩͬ̚̚͜Ȇ̴̟̟͙̞ͩ͌͝S̨̥̫͎̭ͯ̿̔̀ͅ ([xml.etree in the stdlib](https://docs.python.org/3/library/xml.etree.elementtree.html), don't bother with the other stdlib ones)

craigds (commented):
This feels like a good start but I wonder if we could go a little further if we could figure out what GDALMultiDomainMetadata is _meant_ be used for. Every time I've seen it it's full of junk (or a dupe of another file, as seen in this PR), and I can't find any docs for it. Perhaps we could ignore it entirely wherever we see it show up. 

@rcoup might know better.

kart/sqlalchemy/adapter/gpkg.py
olsen232: Let's say I'm doing test driven development and I only write code if it helps the test to pass. We don't care enough about the content of this XML for me to write any more tests, and so the regex approach is probably the least amount of code I can write and should be faster than parsing a tree for this kind of check. I can always swap it out later if we start digging deeper into XML contents

olsen232:
I think GDALMultiDomainMetadata does the same as MIME multipart message - it means multiple separate XML documents can be embedded in a single XML document of type GDALMultiDomainMetadata. So, we could actually this construct to solve this problem in a different way - if we ever find two apparently useful XML attachments, A and B, we could import them combine them into a single GDALMultiDomainMetadata that contains both.

However that is not the right solution in this case, since B in this case is already GDALMultiDomainMetadata wrapping A - so if we did that here, we would just get a GDALMultiDomainMetadata that wraps two XML attachments, but the second attachment is just a GDALMultiDomainMetadata wrapping the first attachment:
```
<GDALMultiDomainMetadata>
  <MDI>A</MDI>
  <MDI><GDALMultiDomainMetadata><MDI>A</MDI></GDALMultiDomainMetadata></MDI>
</GDALMultiDomainMetadata>  
```

craigds (approved):
This PR is a step forward and, in the absence of feedback from @rcoup above we should still merge it

PR #559, opened 2022-02-23
Move table-specific code into kart/tabular

olsen232:
## Description

So far, Kart is used for managing table-datasets, but this won't always be true.
Lots of objects - for instance datasets - will need table in the name somewhere to distinguish them from other types of datsets. This is achieved here as simply as possible by moving a whole bunch of files into a folder called "tabular".

No update to CHANGELOG or tests since no code is changed apart from imports.


craigds (approved):
code looks fine but whatever tool you used has changed indentation and stuff... can you make sure you have pre-commit installed properly? it should have run `black` over all this during commit...

otherwise 👍

Issue #565, opened 2022-03-02
Add support for version-controlled point clouds


olsen232:
Tracking bug - will add PRs here, and make notes of functionality that is missing. Later when point clouds gets closer to launch, this can be used to file individual point cloud issues.

Right now all functionality is missing - point clouds can't be imported, checked out, diffed, modified, committed, etc.

olsen232:
If you use a recent build of Kart, and get PDAL and Git LFS installed, such that they are linked in properly with their existing dependencies such as sqlite, and if you set environment variable `X_KART_POINT_CLOUDS=1` then you should be able to do the following:
- import point clouds, including automatically converting to COPC
- push and pull
- show, diff and commit tile changes
As of #645, branch switches and resets and the working copy side of things is more robust.

Not done:
- meta changes - these are not detected and not supported
- merges + conflicts
- spatial filtering
- the build (PDAL and Git LFS)
- copy-on-write in order to save disk space (?)

olsen232:
### Meta

- Detect meta changes

Right now we don’t store LAZ version or COPC version except in the tiles themselves, and in the pointer files to those tiles (unlike the schema, which we store per-dataset in schema.json). This means that a dataset with no tiles (eg if you delete every tile it has) still has a schema, but no longer has a well-defined LAS version or COPC version.

- Commit meta changes
  o Commit tile changes and discard meta changes
  o Commit tile and meta changes: update dataset meta items to match new tile(s)

Make sure we support adding non-copc files to copc dataset, since, we auto convert non-copc to copc during import, and it will confuse users if they then can't commit additional non-copc files to the imported dataset.

### Merges/conflicts

Make sure repo is correctly moved into merge-state in the case of conflicts.
Add a way to resolve conflicts. 
(For vector, this is only possible by providing the resolution to the CLI using a stdin or a file. There is no way to provide the resolution using a working copy. For PC, it would probably make most sense to provide it using the working copy. We could - if we wanted - add support for providing vector resolutions at the working copy level too).

### Spatial filtering
Support spatial filtering.

### Build issues
PDAL 
Git LFS client

### Copy-on-write
Not sure what this will involve

### Odds and ends
Some more work on reset()
Naming bug - don't name things .copc.copc.laz
Rename detection? If you rename a tile, right now it is treated as a delete + insert.

olsen232:
### Meta changes - done ✅

- Detect and commit meta changes ✅
- Commit meta changes ✅
[...]

PR #570, opened 2022-03-09
Start of point cloud import

olsen232:
Imports individual point cloud tiles by writing pointer files and
putting the tiles in the LFS cache.
This is sufficient so that running `git push` or `kart push` will
cause the tiles to be written to the remote's LFS server (if LFS is
installed).

Lots more things to be done in subsequent changes:
- No point-cloud dataset metadata is written
- Point cloud working copy is not checked out
- Although we generate the necessary tile metadata, we don't yet write
  it to the pointer files since Git LFS doesn't currently allow this

### DIFF
--- a/kart/cli.py
+++ b/kart/cli.py
@@ -294,11 +294,26 @@ def gc(ctx, args):
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def git(ctx, args):
     """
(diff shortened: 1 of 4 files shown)


olsen232:

olsen232:
Will merge for now and work more from here

PR #577, opened 2022-03-13
PC: Convert LAS/LAZ files to COPC.LAZ on import

olsen232:
COPC is a stricter standard - a COPC file  is also a valid LAZ 1.4 file
(Much like a COG file is a type of GeoTiff)

Moved dataset-name validation out of tabular, since it's not table specific.

## Checklist:

### DIFF
--- a/kart/point_cloud/import_.py
+++ b/kart/point_cloud/import_.py
@@ -10,6 +10,7 @@ import click
 from osgeo import osr
 
 from kart.crs_util import make_crs
+from kart.dataset_util import validate_dataset_paths
 from kart.exceptions import (
     InvalidOperation,
     NotFound,
@@ -33,8 +34,11 @@ from kart.repo_version import (
 
 @click.command("point-cloud-import", hidden=True)
 @click.pass_context
+@click.option(
+    "--dataset-path", "ds_path", help="The dataset's path once imported", required=True
+)
 @click.argument("sources", metavar="SOURCES", nargs=-1, required=True)
-def point_cloud_import(ctx, sources):
+def point_cloud_import(ctx, ds_path, sources):
     """
     Experimental command for importing point cloud datasets. Work-in-progress.
     Will eventually be merged with the main `import` command.
@@ -44,6 +48,10 @@ def point_cloud_import(ctx, sources):
     import pdal
 
     repo = ctx.obj.repo
+
+    # TODO - improve path validation to make sure datasets of any type don't collide with each other
+    # or with attachments.
+    validate_dataset_paths([ds_path])
 
     for source in sources:
         if not (Path() / source).is_file():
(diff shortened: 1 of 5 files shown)


kart/point_cloud/import_.py
```diff
@@ -116,19 +124,14 @@ def point_cloud_import(ctx, sources):
             "crs84_envelope": crs84_envelope,
         }
 
-    click.echo()
+    click.echo(copc_version_set[0])
```
craigds, line 127: this will echo `1`, accidentally left in?

kart/point_cloud/import_.py
```diff
@@ -142,15 +145,14 @@ def point_cloud_import(ctx, sources):
...
-    # TODO: Don't accept all possible versions of everything / maybe convert everything to COPC-1.0
-    copc_version = copc_version_set[0]
-    if copc_version == 1:
-        kart_format = "pc:v1/copc-1.0"
-    elif copc_version != NOT_COPC:
-        kart_format = f"pc:DEV/copc-{copc_version}"
-    else:
-        filetype = "laz" if compressed_set[0] is True else "las"
-        kart_format = f"pc:DEV/{filetype}-{version_set[0]}"
+    is_copc = (
```
craigds, line 148: should we silently support future variants of COPC? I'd expect them to work just as well as current versions, so I'd tend not to downgrade to LAZ 1.4/COPC1.0 here if we encounter a newer version.

kart/point_cloud/import_.py
```diff
@@ -265,14 +267,56 @@ def _transform_3d_envelope(transform, envelope):
...
-    BUF_SIZE = 65536
+_BUF_SIZE = 65536
+
+
+def _convert_tile_to_copc_lfs_blob(source, dest):
+    """
+    Converts a LAS/LAZ file of some sort as source to a COPC.LAZ file at dest.
+    Returns the SHA256 and length of the COPC.LAZ file.
+    """
+    # Note that this requires a cutting edge PDAL - e.g. 63eb89ab3f504e8af7259bf60c8158363c99b6e3.
```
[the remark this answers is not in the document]
olsen232: Nope the build does not pull in PDAL yet

kart/point_cloud/import_.py
olsen232: Fixed

kart/point_cloud/import_.py
olsen232: Ah I guess? There is no COPC > 1.0 (and no Kart Point Clouds really either) so hard for me to guess if those two non-existent things are compatible. But at least allowing it makes no real difference right now... I'll allow it

craigds (approved):

kart/point_cloud/import_.py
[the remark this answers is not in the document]
rcoup: @craigds copc writing is in master now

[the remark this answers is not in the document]
olsen232: But it doesn't work properly on master - it needs this PR: https://github.com/PDAL/PDAL/pull/3703
I put the minimum explanation in this comment - will update it at some point.

rcoup:
The LAZ→COPC conversion needs to be optional, but opt-out (ala --no-translate or something) rather than opt-in initially... follow-up to this?

Issue #561, opened 2022-02-27
Spatial filtering: Working copy checkout is slow to skip missing features


olsen232:
This could depend on the size and complexity of features, but testing locally, Kart is currently 70% slower at skipping features that are missing+promised compared to finding and writing a feature that is present locally. This means that creating a spatially filtered clone may or may not be faster than a regular clone, depending on if the time saved by sending less features over the network is counteracted by more time spent trying to read those missing features.

The slowness could be due to any or all of the following:
- raising and catching more exceptions
- git ODB looking in more places for the missing features before finally accepting they are missing
- something else?

olsen232:
Leaning towards the second possibility, which unfortunately means there is no obvious fix - maybe we can delve deep enough into libgit2 to optimise it somehow.
I've been running the following git command to merely print the OIDs of all the objects and verify whether or not they are missing - I don't think it actually loads the data from each object. The pv bit at the end tells me how many lines git is printing per second.
(I'm using a git command directly since it avoids any inefficiencies of going between C and Python repeatedly, including inefficiencies I have added myself. No guarantee that git is doing it as efficiently as possible but it is the best indication we have).

`time git rev-list HEAD --objects --missing=allow-any | pv --line-mode --rate > /dev/null`

- If I run this on a Kart repo where all of the objects were fetched during clone, it runs at 200k to 310k objects per second.
- If I run this on a Kart repo where no objects were fetched during clone - all objects are missing+promised - it runs at 300 to 400 trees per second. Since each tree contains 50 - 60 missing objects in this repository, this is equivalent to running at around 20k objects per second - much slower than the first example.

This may not be a completely fair comparison since from within Kart, it's not enough know all the OIDs and whether they are missing or not - we need to know all the OIDs (without which we can't look up the contents), whether they are missing, *and* the actual contents of any non-missing ones. Looking up the contents would obviously slow down the first example but not the second, since the second has no blob content. However, a full working copy checkout in the first repo completes quicker than this command runs in the second anyway - about 30k features per second. 

In conclusion it seems difficult for us to make working copy checkout in the second repo go as fast as the first repo (30k FPS) when we can't even ask git to print all the non-missing feature OIDs for that fast in the second repo (only 20k FPS), and that knowing the OIDs and if they are missing is a subset of the problem of checking out the working copy.

The good news is that switching between commits often doesn't involve a full working copy checkout, and so will generally be much faster (relatively few features have changed, and relatively few features outside the spatial filter have changed).

[...]

PR #581, opened 2022-03-15
Fix slowness when checking out spatial-filter WC

olsen232:
Slowness is due to libgit2 repeatedly refreshing whenever it encounters
a missing object, in the hopes that the missing object is only due to
its stale view of the data on disk. The fix is to refresh once before
the WC checkout and then disable refreshing for the rest of the
checkout - this current-as-of-start-of-checkout view of the ODB is
good enough for us since we don't read and write concurrently.

https://github.com/koordinates/kart/issues/561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please note that compatibility for 0.x releases (software or repositories) isn't
 
 _When adding new entries to the changelog, please include issue/PR numbers wherever possible._
 
+## 0.11.1
+
+* Fix for the slowness when creating the working copy in a spatially filtered repository (this was previously slower than usual in a spatially filtered repository, now it is faster) [#561](https://github.com/koordinates/kart/issues/561)
+
 ## 0.11.0
 
 ### Major changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.1
 
-* Fix for the slowness when creating the working copy in a spatially filtered repository (this was previously slower than usual in a spatially filtered repository, now it is faster) [#561](https://github.com/koordinates/kart/issues/561)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Please note that compatibility for 0.x releases (software or repositories) isn't
 
 _When adding new entries to the changelog, please include issue/PR numbers wherever possible._
 
+## 0.11.1
+
+* Fix for the slowness when creating the working copy in a spatially filtered repository (this was previously slower than usual in a spatially filtered repository, now it is faster) [#561](https://github.com/koordinates/kart/issues/561)
```
craigds, line 9:
```suggestion
* Improve performance when creating a working copy in a spatially filtered repository (this was previously slower than in a non-filtered repository; now it is much faster) [#561](https://github.com/koordinates/kart/issues/561)
```

kart/working_copy/base.py
```diff
@@ -1444,3 +1445,10 @@ def _reset_dirty_rows(
             self._reset_tracking_table_for_table(
                 sess, base_ds.table_name, feature_filter
             )
+
+
+@contextlib.contextmanager
+def pause_refreshing(odb):
+    odb.set_lookup_flags(pygit2.GIT_ODB_LOOKUP_NO_REFRESH)
+    yield
+    odb.set_lookup_flags(0)
```
craigds, line 1454:
```suggestion
    old_flags = odb.lookup_flags()
    odb.set_lookup_flags(pygit2.GIT_ODB_LOOKUP_NO_REFRESH)
    try:
        yield
    finally:
        odb.set_lookup_flags(old_flags)
```

craigds (approved):
nice find

CHANGELOG.md
olsen232: Done

kart/working_copy/base.py
olsen232: Done

PR #584, opened 2022-03-16
Make schema changes fail gracefully

### LINKED ISSUE #515: Kart breaks when geojson output is used in conflicts command and conflict only affects schemas
To reproduce:

- Modify a dataset changing its schema
- Make a different schema change in a different branch
- Merge branches.

The merge operation will result in conflicts.

Run `kart conflicts -ogeojson`

The following error is raised:
```
  File "kart_cli.py", line 4, in <module>
  File "kart\cli.py", line 340, in entrypoint
  File "lib\site-packages\click\core.py", line 829, in __call__
  File "lib\site-packages\click\core.py", line 782, in main
  File "kart\cli.py", line 163, in invoke
  File "lib\site-packages\click\core.py", line 1259, in invoke
  File "lib\site-packages\click\core.py", line 1066, in invoke
  File "lib\site-packages\click\core.py", line 610, in invoke
  File "lib\site-packages\click\decorators.py", line 21, in new_func
  File "kart\conflicts.py", line 294, in conflicts
  File "kart\conflicts.py", line 86, in list_conflicts
  File "kart\conflicts.py", line 214, in conflicts_json_as_geojson
TypeError: list indices must be integers or slices, not str
[18628] Failed to execute script kart_cli
```

It doesn'tmake sense to use the geojson output format in this case, but I guess that the command should fail more gracefully.

amal-thundiyil:
## Description
Added logs for conflicts with `geojson` output when parsing `meta-items` which can't be shown as `geojson`. This gives a better error message on using `kart conflicts -ogeojson` for `meta-item` conflicts.

## Related links:
- Fixes #515 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.2 (UNRELEASED)
 
+- Bugfix: Better error message on using `kart conflicts -ogeojson` for `meta-item` conflicts. [#515](https://github.com/koordinates/kart/issues/515)
 - Removed the older `upgrade-to-tidy` and `upgrade-to-kart` features which were only relevant to Sno (predecessor of Kart). [[#585](https://github.com/koordinates/kart/issues/585]]
 - Added support for `--decorate` and `--no-decorate` in `kart log`. [[#586](https://github.com/koordinates/kart/issues/586]]
 
--- a/kart/conflicts.py
+++ b/kart/conflicts.py
@@ -1,6 +1,5 @@
 import logging
 import sys
-
 import click
 
 from .cli_util import OutputFormatType, parse_output_format
@@ -216,9 +215,19 @@ def conflicts_json_as_text(json_obj):
 def conflicts_json_as_geojson(json_obj):
     """Converts the JSON output of list_conflicts to geojson."""
     features = []
+    meta_item_keys = set()
     for key, feature in json_obj.items():
+        if type(feature) != dict or feature.get("type") != "Feature":
+            meta_item_keys.add(":".join(key.split(":")[:-1]))
+            continue
         feature["id"] = key
         features.append(feature)
+
+    if meta_item_keys:
+        L.warning(
+            f"There are meta-item conflicts that cannot be output as geojson: \n{meta_item_keys}"
+        )
+
     return {"type": "FeatureCollection", "features": features}
 
 
--- a/tests/test_conflicts.py
+++ b/tests/test_conflicts.py
@@ -336,3 +336,25 @@ def test_find_renames(data_working_copy, cli_runner):
         )
 
         assert not index.conflicts
+
+
+def test_json_conflicts_as_geojson(data_working_copy, cli_runner):
+    with data_working_copy("points") as (repo_path, wc):
+        repo = KartRepo(repo_path)
+
+        r = cli_runner.invoke(["checkout", "-b", "theirs_branch"])
+        r = cli_runner.invoke(["branch", "ours_branch"])
+        with repo.working_copy.session() as sess:
+            r = sess.execute(f"""ALTER TABLE {H.POINTS.LAYER} ADD COLUMN new_col1""")
+            assert r.rowcount == -1
+        r = cli_runner.invoke(["commit", "-m", "theirs_commit"])
+
+        r = cli_runner.invoke(["checkout", "ours_branch"])
+        with repo.working_copy.session() as sess:
+            r = sess.execute(f"""ALTER TABLE {H.POINTS.LAYER} ADD COLUMN new_col2""")
+            assert r.rowcount == -1
+        r = cli_runner.invoke(["commit", "-m", "ours_commit"])
+
+        r = cli_runner.invoke(["merge", "theirs_branch"])
+        r = cli_runner.invoke(["conflicts", "-o", "geojson"])
+        assert r.exit_code == 0, r


olsen232:
This could be an opportunity to familiarise yourself with Kart itself and with the codebase a little bit.
Have you been able to reproduce the issue yourself? It's admittedly a bit complicated. 
Start with the [Basic usage tutorial](https://github.com/koordinates/kart/wiki/Basic-Usage-Tutorial). Make sure you can do everything there.
Then, are you able to create two different branches that contain different changes, and merge them together? Looking up Git tutorials about branches and merging might help you here, since Kart behaviour is very similar.
Then, are you able to create two different branches that contain conflicting changes, and merge them together? `kart conflicts` and `kart resolve` are the two commands that help you to resolve the conflicts.
Then, are you able to create two different branches that contain conflicting schema changes, and try merge them together?
And can you then run `kart conflicts -o geojson` while there are schema conflicts, to reproduce the bug?

The bug is to do with the fact that all "features" - database table rows - are stored as dicts, and can always be represented as geojson. We also store some pieces of metadata about the table, which we mostly call "meta items" and they are not necessarily dicts and they cannot be shown as geojson - geojson is specifically for representing geographic features.

A fix for the bug should probably leave any meta item conflicts as they are, instead of trying to convert them to geojson. An alternative fix would be to not output them at all (but in that case we would have to output some kind of warning to indicate that meta item conflicts have been skipped).

amal-thundiyil:
@olsen232 Thank you for the elaborate description. I have made some changes according to it.

olsen232:
This is getting better but it still has the following issue:
- if a meta-item happens to be a dict, this code will still try to convert it into geojson. Meta-items are not geographic features and so we shouldn't try to convert them into geojson - we should be detecting specifically if it is a meta-item or a feature, not whether or not it is a dict. (Speaking of this, were you able to reproduce the issue locally?)

Probably prefer to use either click.echo or the logging package to warn about missing meta items - we don't really use `warnings.warn` so far and I don't want every class to use a different framework for output.

amal-thundiyil:
> Speaking of this, were you able to reproduce the issue locally?
Yes I could reproduce the issue locally.

Made changes using `logging` and checking for `Feature`.

kart/conflicts.py
```diff
@@ -217,6 +216,10 @@ def conflicts_json_as_geojson(json_obj):
     """Converts the JSON output of list_conflicts to geojson."""
     features = []
     for key, feature in json_obj.items():
+        if type(feature) != dict or feature.get("type") != "Feature":
+            L.warning(f"Failed to convert meta-item conflicts to geojson.")
```
olsen232, line 220: Since converting anything that is not a geographic feature to geojson doesn't make sense, these errors could be more to the point eg "There are meta-item conflicts that cannot be output as geojson." ("Failed to convert" makes it seem like we attempted to convert them to geojson, but we don't).
This is a bit too warny right now - it warns about three times per conflict that you have, and each new warning adds no additional information. I would prefer any of the following:
- Warn once per conflict, include the conflict's key (but not the version part eg "ancestor/ours/theirs")
- Or, Warn once per dataset, include the name of the dataset.
- Or, just warn once if there are dropped meta-items at all.
Doing any of these will make the fix a little bit more complicated. But you can choose whichever one seems easiest or best to you.

No need to log the dropped meta items - if the user needs to see them, they can run a command that shows them.

olsen232 (commented):
Nearly there! Thanks for working with me on this!

kart/conflicts.py
amal-thundiyil: Thanks for the suggestion.

kart/conflicts.py
```diff
@@ -216,9 +216,20 @@ def item_to_text(key, value, path, level):
 def conflicts_json_as_geojson(json_obj):
     """Converts the JSON output of list_conflicts to geojson."""
     features = []
+    meta_item_key = ""
```
olsen232, line 219: If you have an optional string, prefer to initialise it to None (or if there could be more than one, have an empty collection of strings). Using an empty string is more ambiguous - it might be representing the NULL value, as in, we don't have a string, or it might be a valid value for the string - it depends on the context. So, prefer to use None if you can.

kart/conflicts.py
```diff
@@ -1,6 +1,6 @@
+from curses import meta
```
olsen232, line 1: I think your IDE has imported this for you - doesn't look like you are using it

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 Please note that compatibility for 0.x releases (software or repositories) isn't guaranteed. Kart is evolving quickly and things will change. However, we aim to provide the means to upgrade existing repositories between 0.x versions and to 1.0.
 
 _When adding new entries to the changelog, please include issue/PR numbers wherever possible._
+## 0.11.2 (UNRELEASED)
```
olsen232, line 6: You'll need to merge or rebase this with your other change now which I've just merged to master for you

kart/conflicts.py
```diff
@@ -216,9 +216,20 @@ def item_to_text(key, value, path, level):
...
     features = []
+    meta_item_key = ""
     for key, feature in json_obj.items():
+        if type(feature) != dict or feature.get("type") != "Feature":
+            meta_item_key = key
+            continue
         feature["id"] = key
         features.append(feature)
+
+    if meta_item_key:
```
olsen232, line 227: This prints out the last encountered meta-item conflict, regardless of how many there are.
I recommend instead one of the following options: 
- print out a single warning (in this case, don't bother to explain what they are at all)
- print out a warning for each dataset with meta-item conflicts (in this case, give the path of the dataset that has a meta-item conflict)
- print out a warning for each meta-item conflict (in this case, print the entire key of the meta-item, but not the version part of it eg ancestor, ours, theirs)

kart/conflicts.py
amal-thundiyil: Sure, will keep this in mind.

kart/conflicts.py
amal-thundiyil: > print out a warning for each meta-item conflict (in this case, print the entire key of the meta-item, but not the version part of it eg ancestor, ours, theirs)

I have made some changes according to this. Let me know if it's what you meant.

olsen232 (approved):

PR #587, opened 2022-03-17
Import point-cloud metadata - schema and CRS

olsen232:
- import meta/schema.json and meta/crs/<CRS_ID>.wkt during point cloud
  import.
- also:
  = leave COPC as COPC
  = optionally convert LAZ to COPC (default is yes)
  = always convert LAS to COPC

### DIFF
--- a/kart/point_cloud/import_.py
+++ b/kart/point_cloud/import_.py
@@ -9,7 +9,7 @@ import sys
 import click
 from osgeo import osr
 
-from kart.crs_util import make_crs
+from kart.crs_util import make_crs, get_identifier_str, normalise_wkt
 from kart.dataset_util import validate_dataset_paths
 from kart.exceptions import (
     InvalidOperation,
@@ -24,7 +24,7 @@ from kart.fast_import import (
     write_blob_to_stream,
     write_blobs_to_stream,
 )
-from kart.serialise_util import hexhash
+from kart.serialise_util import hexhash, json_pack, ensure_bytes
 from kart.output_util import format_wkt_for_output
 from kart.repo_version import (
     SUPPORTED_REPO_VERSIONS,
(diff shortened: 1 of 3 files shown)


kart/point_cloud/import_.py
```diff
@@ -34,11 +34,12 @@
 
 @click.command("point-cloud-import", hidden=True)
 @click.pass_context
+@click.option("--convert-to-copc/--no-convert-to-copc", is_flag=True, default=True)
```
craigds, line 37: not sure why this is an option? I would have thought we'd always want everything in COPC. Why would a user [be able to] disable this?

If it's just for some testing use-case, should add hidden=True to avoid 'supporting' this.

If it's for performance reasons (don't want import to be slow) we might be shooting ourselves in the foot since it'll presumably make data consumption slower later?

kart/point_cloud/import_.py
```diff
@@ -124,8 +137,39 @@ def point_cloud_import(ctx, ds_path, sources):
             "crs84_envelope": crs84_envelope,
         }
 
+        if schema is None:
+            crs_name = get_identifier_str(crs_set[0])
```
craigds, line 141: minor thing but: these things are ListBasedSet instances - I understand the reason is so you can have a set that stores non-hashable things; ok fair enough

but then you're accessing them here quite unlike a set, relying on the underlying type as well as the ordering. I'd prefer `list(crs_set)[0]` or something I guess. Maybe you shouldn't be able to index a ListBasedSet at all since it's trying to be like a set

kart/point_cloud/import_.py
```diff
@@ -124,8 +137,39 @@ def point_cloud_import(ctx, ds_path, sources):
...
+            schema = metadata["filters.info"]["schema"]
+            schema["CRS"] = crs_name
+
     click.echo()
 
+    version = version_set[0]
```
craigds, line 147: and, for these ones where they _don't_ store dicts, maybe just use a normal `set()` ?

kart/point_cloud/import_.py
olsen232: This spec has been provided by Rob himself. The original plan was to only allow COPC since it's the cooler format that lets us do more stuff - but, newer plan (IIUC) is that sometimes you'll be able to turn a pile of LAZ files that you have in eg S3 into a Kart repo somehow.  I haven't got to that bit of the plan yet - focussing on local only still - but as a concession to it, I'm already allowing repos backed with non-COPC LAZ files (which are commonly used in the real world, but not LAS files, which mostly aren't).

kart/point_cloud/import_.py
olsen232: I made a function .only() that does the same as [0] except that it asserts that it's a single-element-set, and I removed the __getitem__ operator.
It's a well-known fact that the first element of a single-element-set is reliably the only element, and that you can get it with next(iter(myset)) but that's too wordy. So I've compromised with .only() which is 7 characters compared to [0] which is three.

kart/point_cloud/import_.py
olsen232: I could but there's no real upside - it doesn't even make the code simpler

craigds (approved):

PR #593, opened 2022-03-21
Check out point cloud tiles

olsen232:
Currently only performed immediately after a point cloud import.
Assumes all tiles are available in the local cache (doesn't try to
fetch them if they are not).

### DIFF
--- a/kart/lfs_util.py
+++ b/kart/lfs_util.py
@@ -6,16 +6,18 @@ import pygit2
 POINTER_PATTERN = re.compile(rb"^oid sha256:([0-9a-fA-F]{64})$", re.MULTILINE)
 
 
-def get_hash_from_pointer_file(pointer_file):
-    if isinstance(pointer_file, pygit2.Blob):
-        pointer_file = memoryview(pointer_file)
-    match = POINTER_PATTERN.search(pointer_file)
+def get_hash_from_pointer_file(pointer_file_bytes):
+    """Given a pointer-file Blob or bytes object, extracts the sha256 hash from it."""
+    if isinstance(pointer_file_bytes, pygit2.Blob):
+        pointer_file_bytes = memoryview(pointer_file_bytes)
+    match = POINTER_PATTERN.search(pointer_file_bytes)
     if match:
         return str(match.group(1), encoding="utf8")
     return None
 
 
 def get_local_path_from_lfs_hash(repo, lfs_hash):
+    """Given a sha256 LFS hash, finds where the object would be stored in the local LFS cache."""
     return (
         repo.gitdir_path / "lfs" / "objects" / lfs_hash[0:2] / lfs_hash[2:4] / lfs_hash
     )
--- a/kart/lfs_util.py
+++ b/kart/lfs_util.py
@@ -1,0 +1,21 @@
+import re
+
+import pygit2
+
+
+POINTER_PATTERN = re.compile(rb"^oid sha256:([0-9a-fA-F]{64})$", re.MULTILINE)
+
+
+def get_hash_from_pointer_file(pointer_file):
+    if isinstance(pointer_file, pygit2.Blob):
+        pointer_file = memoryview(pointer_file)
+    match = POINTER_PATTERN.search(pointer_file)
+    if match:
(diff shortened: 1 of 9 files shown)


kart/point_cloud/dataset1.py
```diff
@@ -24,4 +32,29 @@ def __init__(self, tree, path, dirname=None, repo=None):
...
+    @property
+    @functools.lru_cache(maxsize=1)
+    def tiles_tree(self):
+        """Returns the root of the tiles tree, or the empty tree if no tiles tree exists."""
+        if self.inner_tree:
+            try:
+                return self.inner_tree / self.TILES_PATH
+            except KeyError:
+                pass
+        return self.repo.empty_tree if self.repo else None
```
craigds, line 44: is `self.repo is None` a valid state? how does that happen? I'm not clear that returning None here is a sensible thing to do. Similarly unclear on whether returning an empty tree is helpful. Maybe this needs more comments explaining those states

kart/repo.py
```diff
@@ -657,8 +668,8 @@ def del_config(self, key):
 
     def invoke_git(self, *args):
         try:
-            args = ["git", "-C", self.path, *args]
-            subprocess.check_call(args, env=tool_environment())
+            args = ["git", *args]
+            subprocess.check_call(args, env=tool_environment(), cwd=self.workdir_path)
```
craigds, line 672: 👍

kart/tabular/dataset3.py
```diff
@@ -160,7 +144,7 @@ def crs_definitions(self):
         """Returns {identifier: definition} dict for all CRS definitions in this dataset."""
         if not self.inner_tree or self.CRS_PATH not in self.inner_tree:
             return
-        for blob in find_blobs_in_tree(self.inner_tree / self.CRS_PATH):
+        for blob in find_blobs_in_tree(self.inner_tree / self.CRS_PATH, max_depth=4):
```
craigds, line 147: why these extra kwargs? Feels unnecessary at best, and could introduce a nasty bug if we change the tree depth for some reason and don't realise

craigds (approved):

kart/lfs_util.py
```diff
@@ -0,0 +1,21 @@
+import re
+
+import pygit2
+
+
+POINTER_PATTERN = re.compile(rb"^oid sha256:([0-9a-fA-F]{64})$", re.MULTILINE)
+
+
+def get_hash_from_pointer_file(pointer_file):
```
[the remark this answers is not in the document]
olsen232: Done

kart/point_cloud/dataset1.py
olsen232: Yeah the dataset constructor interface is a little confused - a symptom of the long evolution of datasets up to this point, and the variety of different ways they are now used. Every dataset currently has the same constructor, which is used as an interface, and most constructors still handle every special case (partly because I think it will end up in a common-to-all-datasets base class soon).
This particular weirdness (repo maybe None but maybe isn't) I think is a symptom of the following causes:
- I had a bug "Fix for feature-KeyError in empty dataset" and I fixed it by making methods that return trees return empty-trees if they had no data (ie, dataset doesn't exist, dataset does exist but doesn't have a `feature` directory, etc). That means no dataset needs to be special cased - `.feature_tree` always returns a valid tree.
- Returning an empty-tree means wiring in the repo, which a dataset doesn't otherwise depend directly. I did this but made it optional (which was a bit messy and maybe the wrong call)...
- which means that unit tests on the Dataset3 class which don't even use a git repo still work - they just create some tree-like objects that support the `/` operator, but they don't create a repo-like object or a global empty-tree object.

Anyway you are right it needs cleaning up but I think for now I'll just make this particular dataset require a non-null repo, and I'll fix the rest of it when I figure out what the common base-class of all datasets should be.

kart/tabular/dataset3.py
olsen232: I just moved this function into core and this was the old default since features were only ever at a quite a low max-depth compared to the feature tree. But I'll just take it out max-depth completely since it's more likely to cause a bug than to prevent it.

PR #599, opened 2022-03-27
Codify meta-items so that they can be reused

olsen232:
Specifically, codify the following -
- each type of dataset has a certain set of possible meta-items
- some meta-items will be common to most datasets
- meta-items can be singular, at a path, or multiple with a pattern
- each meta-item has a type that informs how it is serialised and inflated
- some meta items are more user-facing, some are more hidden (see permissions)


kart/base_dataset.py
```diff
@@ -105,6 +227,18 @@ def __init__(self, tree, path, repo, dirname=None):
...
+    def _init_meta_item_definitions(cls):
+        if hasattr(cls, "PATH_META_ITEMS"):
+            return
+        cls.PATH_META_ITEMS = {}
+        cls.PATTERN_META_ITEMS = []
+        for definition in cls.META_ITEMS:
+            if definition.path:
+                cls.PATH_META_ITEMS[definition.path] = definition
+            else:
+                cls.PATTERN_META_ITEMS.append(definition)
```
craigds, line 240: super edge case, but this init pattern doesn't guard against a race condition where `cls.PATH_META_ITEMS` is set, an exception is thrown, the exception is caught and the dict/list are never populated.

Instead, populate a couple of local vars and then set them as class attributes at the end.

Also, maybe this should live in a metaclass so it doesn't have to happen during instance init

kart/tabular/v3.py
```diff
@@ -60,43 +82,68 @@ class TableV3(RichTableDataset):
...
+        SCHEMA_JSON,
+        BaseDataset.CRS_DEFINITIONS,
+        GENERATED_PKS,
+        PATH_STRUCTURE,
+        LEGEND,
+    )
+
+    # This meta-item is generally stored in the "attachment" area, alongside the dataset, rather than inside it.
+    # Storing it in this unusual location adds complexity without actually solving any problems, so newer datasets
+    # don't do this.
```
craigds, line 123: in table dataset v3 we still do this, so maybe clarify what you mean by 'newer datasets' (point clouds)

craigds (approved):

kart/base_dataset.py
```diff
@@ -1,28 +1,132 @@
...
+        parts = meta_item_path.rsplit(".", maxsplit=1)
+        if len(parts) == 2:
+            try:
+                return cls[parts[1].upper()]
+            except KeyError:
+                pass
+        return None
+
+    def decode_from_bytes(self, data):
+        if data is None:
```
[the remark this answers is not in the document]
olsen232: We discussed this offline and we agree that one way is more natural than the other but that both ways ultimately do the same thing. I'm still not convinced that == is the more natural way of expressing this but luckily it doesn't really matter so I've changed it to ==

kart/base_dataset.py
```diff
@@ -1,28 +1,132 @@
...
+            return definition.file_type
+        else:
+            return (
+                MetaItemFileType.get_from_suffix(meta_item_path)
+                or MetaItemFileType.TEXT
+            )
+
+
+@functools.total_ordering
+class MetaItemPermission(Enum):
```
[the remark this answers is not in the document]
olsen232: Done

kart/tabular/v3.py
olsen232: I've added "for example point clouds", so I don't have to maintain a list here of datasets newer than table.v3

kart/base_dataset.py
olsen232: Done

PR #601, opened 2022-03-31
Add commit<>commit diffs to point-cloud datasets

olsen232:
Moves the feature_diff code out of the table-based datasets and into
the base-dataset. This code runs a git-diff, then takes all the files
that have changed and optionally uses a method to decode the paths
(eg decode_to_primary_key) and a method to transform the blob contents
(eg get_feature_from_blob). Similar but more basic transforms now
exist for point-cloud tiles that make a summary of the tile just from
the information in the pointer file.

### DIFF
--- a/kart/dataset_mixins.py
+++ b/kart/dataset_mixins.py
@@ -1,5 +1,9 @@
-from kart.diff_structs import DatasetDiff, DeltaDiff
-from kart.key_filters import DatasetKeyFilter, MetaKeyFilter
+import functools
+
+import pygit2
+
+from kart.diff_structs import DatasetDiff, DeltaDiff, Delta
+from kart.key_filters import DatasetKeyFilter, MetaKeyFilter, UserStringKeyFilter
 
 
 class DatasetDiffMixin:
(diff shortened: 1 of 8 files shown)


kart/dataset_mixins.py
```diff
@@ -46,3 +50,130 @@ def diff_to_wc(self, repo, ds_filter=DatasetKeyFilter.MATCH_ALL):
...
+    _INSERT_UPDATE = (pygit2.GIT_DELTA_ADDED, pygit2.GIT_DELTA_MODIFIED)
+    _UPDATE_DELETE = (pygit2.GIT_DELTA_MODIFIED, pygit2.GIT_DELTA_DELETED)
+
+    def diff_subtree(
+        self,
+        other,
+        subtree_name,
+        key_filter=UserStringKeyFilter.MATCH_ALL,
+        key_decoder_method=None,
+        value_decoder_method=None,
```
craigds, line 89: these appear to be _required_ but default to None and aren't checked. I'd suggest not giving them a default value but putting them after a `*` so they're required keyword only args

kart/dataset_mixins.py
```diff
@@ -46,3 +50,130 @@ def diff_to_wc(self, repo, ds_filter=DatasetKeyFilter.MATCH_ALL):
...
+                if dataset is not None
+                else lambda key: key
+            )
+
+        old_key_decoder = get_decoder(old, key_decoder_method)
+        new_key_decoder = get_decoder(new, key_decoder_method)
+        old_value_decoder = get_decoder(old, value_decoder_method)
+        new_value_decoder = get_decoder(new, value_decoder_method)
+
+        subtree_path = subtree_name + "/"
```
craigds, line 122:
```suggestion
        subtree_path = subtree_name.rstrip("/") + "/"
```

kart/dataset_mixins.py
```diff
@@ -46,3 +50,130 @@ def diff_to_wc(self, repo, ds_filter=DatasetKeyFilter.MATCH_ALL):
...
+
+        for d in raw_diff.deltas:
+            self.L.debug(
+                "diff(): %s %s %s", d.status_char(), d.old_file.path, d.new_file.path
+            )
+
+            if d.status not in self._INSERT_UPDATE_DELETE:
+                # RENAMED, COPIED, IGNORED, TYPECHANGE, UNMODIFIED, UNREADABLE, UNTRACKED
+                raise NotImplementedError(f"Delta status: {d.status_char()}")
```
craigds, line 131: is this something we need to implement, or could some/all of these delta types simply be ignored? From the comment I expect only RENAMED is something we'll actually care about

kart/json_diff_writers.py
```diff
@@ -98,6 +107,11 @@ def default(self, obj):
                 result["feature"] = self.filtered_ds_feature_deltas_as_json(
                     ds_path, ds_diff
                 )
+            if "tiles" in ds_diff:
```
craigds, line 110: we have "meta" (ambiguous) and "feature" (singular), should we have "tile" singular too for consistency

kart/json_diff_writers.py
```diff
@@ -98,6 +107,11 @@ def default(self, obj):
...
+                result["tiles"] = [
+                    self._postprocess_tiles_delta(key, value)
+                    for key, value in ds_diff["tiles"].items()
+                ]
```
craigds, line 114: So until now `tiles` here is a dict/object:

```
{
    "mytilename1.laz": {"-": {...}, "+": {...}},
    ...
}
```

and this thing turns it into a list

```
[
    {"-": {"name": "mytilename1.laz", ...}, "+": {"name": "mytilename1.laz", ...}},
    ...
]
```

Could you add a comment explaining what's going on here? It's not very clear to me

kart/lfs_util.py
```diff
@@ -1,11 +1,30 @@
...
+    if isinstance(pointer_file_bytes, pygit2.Blob):
+        pointer_file_bytes = pointer_file_bytes.data
+    pointer_file_str = pointer_file_bytes.decode("utf8")
+    result = {}
+    for line in pointer_file_str.splitlines():
+        if not line:
+            continue
+        parts = line.split(" ", maxsplit=1)
+        if len(parts) < 2:
+            L.warn(f"Error parsing pointer file:\n{line}")
```
craigds, line 21: Seems like this should be an exception; it's an unexpected condition and there's no telling what effect this might have further up the stack

kart/point_cloud/v1.py
```diff
@@ -40,3 +47,52 @@ def tilenames_with_lfs_paths(self):
         """Returns a generator that yields every tilename along with the path where the tile content is stored locally."""
         for blob_name, lfs_hash in self.tilenames_with_lfs_hashes():
             yield blob_name, get_local_path_from_lfs_hash(self.repo, lfs_hash)
+
+    def decode_path(self, path):
+        rel_path = self.ensure_rel_path(path)
+        if rel_path.startswith("tiles/"):
```
craigds, line 53: Maybe(?) it's too late to change this, but since the feature tree is called `feature/` singular it feels like maybe this should be called `tile/` singular too. 🤷

kart/dataset_mixins.py
olsen232: Done

kart/dataset_mixins.py
olsen232: Done

kart/dataset_mixins.py
olsen232: This has been like this for as long as Sno / Kart has been a thing. You are right that RENAMED is probably the only one that this commit<>commit diff code could potentially handle which it right now doesn't.

Saying the rest "can be ignored" is a bit misleading - right now we don't "ignore" them, we raise an error if we encounter them, and in practise so far we have not encountered them - due to some mixture of the limited things that Kart does and the limited ways we are generating pygit2.Diff objects here (ie commit<>commit only, no rename detection)

I'll add a comment explaining we don't expect to see these here and why.

kart/json_diff_writers.py
olsen232: I'll do this in a separate commit.

kart/dataset_mixins.py
```diff
@@ -46,3 +50,130 @@ def diff_to_wc(self, repo, ds_filter=DatasetKeyFilter.MATCH_ALL):
...
         # Subclasses to override.
         pass
+
+    def get_raw_diff_for_subtree(self, other, subtree_name, reverse=False):
+        """
+        Get a pygit2.Diff of the diff between some subtree of this dataset, and the same subtree of another dataset
+        (generally the "same" dataset at a different revision).
+        """
+
+        flags = pygit2.GIT_DIFF_SKIP_BINARY_CHECK
```
rcoup, line 60: should we be using `GIT_DIFF_FORCE_BINARY` here?

> `GIT_DIFF_FORCE_BINARY`
> 
> Treat all files as binary, disabling text diffs
>
> `GIT_DIFF_SKIP_BINARY_CHECK`
> 
> Disable updating of the binary flag in delta records. This is useful when iterating over a diff if you don't need hunk and data callbacks and want to avoid having to load file completely.
>
> ([source](https://libgit2.org/libgit2/#HEAD/type/git_diff_option_t))

kart/dataset_mixins.py
rcoup: I think if we want other status' we need to pass additional options to libgit2? Which presumably has some performance penalty.

> I'll add a comment explaining we don't expect to see these here and why.

👍

kart/lfs_util.py
rcoup: should probably check for empty and repeated keys too?

kart/point_cloud/v1.py
```diff
@@ -40,3 +47,52 @@ def tilenames_with_lfs_paths(self):
...
+            other,
+            "tiles",
+            key_filter=tiles_filter,
+            key_decoder_method="decode_tile_path",
+            value_decoder_method="get_tile_summary_from_pointer_blob",
+            reverse=reverse,
+        )
+
+    def get_tile_summary_from_pointer_blob(self, tile_pointer_blob):
+        # For now just return the blob contents as a string - it's a reasonable summary.
```
rcoup, line 94: This is certainly a good starting point.

Conceptually `pdal info --summary` (and `gdalinfo -json` for raster tiles) are probably more useful to see high level changes: tile has a different extent, etc

Do we have a concept of more-verbose diffs? (ala `--schema`,`--metadata`,`--statistics`,`--all` args to `pdal info` might be useful, but too verbose usually).

rcoup (commented):
👍

PR #602, opened 2022-04-02
Make conflicts and diffs consistent

### LINKED ISSUE #135: Consistent geojson output in `kart diff` and `kart conflicts`
Background:
Geojson output must contain a flat featurecollection. 
JSON output that is quite structured from `sno diff` and `sno conflicts` must be flattened for geojson output, and they take different approaches.

`sno diff -o json` organises features into the following JSON structure:
`<dataset>` / `<feature|meta>` / `<pk>` / `<+|->` / `<feature>`

`sno diff -o geojson` can't do this, so it just puts the features into a flat array:
[`<feature>`, ...]
but each feature is given an ID which has the following structure:
`<+/-><U|I|D>::<pk>`
The dataset is not part of the ID. If multiple datasets have diffs, then a directory must be specified for output, and one file is output for each dataset with diffs.

`sno conflicts -o json` organises conflicts into the following JSON structure:
`<dataset>` / `<feature|meta>` / `<pk>` / `<ancestor|ours|theirs>` / `<feature>`

`sno conflicts -o geojson` can't do this, so it just puts the features into a flat array, but each feature is given an ID which exactly matches its path in the JSON output:
`<dataset>:<feature|meta>:<pk>:<ancestor|ours|theirs>`

Both `sno diff` and `sno conflicts` can limit the output to a single dataset if needed using a filter eg `sno diff HEAD datasetA`

To do:
The `sno conflicts` approach seems better in at least one respect: a feature's JSON path is the same as its geojson ID, which is the simplest / least surprising behaviour. `sno diff -o geojson` should probably be changed so that geojson IDs match with JSON paths. 

The other way in which the two commands differ - returning a single array with data from all dataset, or multiple arrays, one per dataset, in separate files - I don't know which is better. I would hope that if the consumer can correctly extract primary key or old/new version from the ID, they could also extract the dataset from the ID, but perhaps the consumer (ie QGIS) cannot actually break the ID into its parts - in which case, geojson diffs loaded into QGIS will be hard enough to understand without having multiple datasets in the same diff, breaking them into several files could help the user considerably.
Perhaps we could support both behaviours - if the user supplies a file or stream as output, we generate a file, but if they supply a directory, we write one file per dataset to that directory. Either way, we should do this consistently for `sno diff` and `sno conflicts`.

amal-thundiyil:
## Description

- Changed feature's GeoJSON id to be same as JSON path. Example: `dataset:feature:id:U-`
  - This is simpler or atleast less surprising than ids like `U+::123`.
- Added support for `--output` for JSON, text and GeoJSON formats in `kart conflicts`.
  - Added support for writing `kart conflicts` to separate directories according to datasets for GeoJSON output.
- Refactor `kart conflicts` to make use of polymorphism.

## Related links:

- Resolves #135 

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: check-added-large-files
       - id: check-json
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
   - repo: https://github.com/sirosen/check-jsonschema
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.2 (UNRELEASED)
 
+- Added support for `--output` option `kart conflicts`. [#135](https://github.com/koordinates/kart/issues/135)
 - Bugfix: Better error message on using `kart conflicts -ogeojson` for `meta-item` conflicts. [#515](https://github.com/koordinates/kart/issues/515)
 - Removed the older `upgrade-to-tidy` and `upgrade-to-kart` features which were only relevant to Sno (predecessor of Kart). [[#585](https://github.com/koordinates/kart/issues/585]]
 - Added support for `--decorate` and `--no-decorate` in `kart log`. [[#586](https://github.com/koordinates/kart/issues/586]]
--- a/kart/conflicts_writer.py
+++ b/kart/conflicts_writer.py
@@ -1,0 +1,365 @@
+from __future__ import annotations
+import logging
+import sys
+import click
+import pygit2
+
+from typing import Union, Type, List
+
+from .crs_util import CoordinateReferenceString
+from .merge_util import MergeContext, MergeIndex, RichConflict
+
+
+from .conflicts_util import (
+    _CONFLICT_PLACEHOLDER,
+    set_value_at_dict_path,
+    summarise_conflicts,
+    conflicts_json_as_text,
+)
+from .merge_util import MergeContext, MergeIndex, ensure_conflicts_ready, rich_conflicts
+from pathlib import Path
+from .output_util import dump_json_output, resolve_output_path
+from pathlib import Path
+from .key_filters import RepoKeyFilter
+from . import diff_util
+
+L = logging.getLogger("kart.conflicts_writer")
+
+
+class BaseConflictsWriter:
+    """
+    A base class for writing conflicts output.
+    """
+
+    def __init__(
+        self,
+        repo: pygit2.Repository,
+        user_key_filters: tuple = (),
+        output_path: str = "-",
+        summarise: int = 0,
+        flat: bool = False,
+        *,
+        json_style: str = "pretty",
+        target_crs: CoordinateReferenceString = None,
+        merge_index: MergeIndex = None,
+        merge_context: MergeContext = None,
+    ):
+        self.repo = repo
+        self.target_crs = target_crs
+        self.flat = flat
+        self.summarise = summarise
+        self.merge_context = merge_context
+        self.merge_index = merge_index
+        self.repo_key_filter = RepoKeyFilter.build_from_user_patterns(user_key_filters)
+        self.json_style = json_style
+        self.output_path = self._check_output_path(
+            repo, self._normalize_output_path(output_path)
+        )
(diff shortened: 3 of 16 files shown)


.github/workflows/build.yml
```diff
@@ -263,7 +263,7 @@ jobs:
   #
   windows:
     name: Windows
-    runs-on: windows-2016
+    runs-on: windows-2019
```
olsen232, line 266: Take this commit out of your pull-request - a PR should really just focus on one thing at a time. Here is an example of why: the windows change you've made isn't enough to fix our windows problem. We do need to use a later version of windows, but we also need to use a later version of Visual Studio, and a couple of other things will need to be changed too to make it all work. Nobody has figured out exactly how to do this yet - if you want to work on it, you can do it in a separate PR. For now, leave it out of this PR.

.pre-commit-config.yaml
```diff
@@ -11,7 +11,7 @@ repos:
       - id: check-added-large-files
       - id: check-json
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
```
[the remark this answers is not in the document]
olsen232: So, if you rebase or merge with master, this particular diff should disappear.

tests/test_conflicts.py
```diff
@@ -71,23 +72,23 @@ def test_summarise_conflicts(data_archive, cli_runner):
             "",
         ]
 
-        r = cli_runner.invoke(["conflicts", "-s", "-o", "json"])
+        r = cli_runner.invoke(["conflicts", "-s", "-o", "json", "--output=-"])
```
olsen232, line 75: Keep the default behaviour for output working - if you don't specify output, it should default to stdout.

tests/test_conflicts.py
```diff
@@ -182,57 +188,63 @@ def test_list_conflicts(data_archive, cli_runner):
         )
 
         expected_geojson = {
+            "type": "FeatureCollection",
```
olsen232, line 191: You've rearranged this expected_geojson a bit but it's not obvious to me what's changed.
Has anything changed?
Is one of these orders the "right" order? I guess it would be good to have it in the order that the code actually prints it... would make the test easier to read...

olsen232 (commented):
You've moved a good chunk of code around for someone new to the project, and I think it's made the conflicts code better. Good job!

olsen232:
Related info: this commit is as far as we have got with upgrading the windows version that we build against:
https://github.com/koordinates/kart/commit/ce42d055adddaf0a50ade18e3f8c2f913cac73e7

.github/workflows/build.yml
amal-thundiyil: Sure. Will keep it in mind.

tests/test_conflicts.py
amal-thundiyil: > I guess it would be good to have it in the order that the code actually prints it

Rearranged according to this.

kart/conflicts_writer.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+class GeojsonConflictsWriter(BaseConflictsWriter):
+    """Writes all feature conflicts as a single GeoJSON FeatureCollection of GeoJSON features.
+
+    The id of each feature in the collection indicates which branch it belongs to.
+
+    Example:
+
+        dataset:feature:123:ancestor  - conflicts of feature 123 on ancestor branch
```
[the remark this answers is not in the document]
amal-thundiyil: Done.

tests/test_conflicts.py
olsen232: Did anything change?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.2 (UNRELEASED)
 
+- Added support for `--output` option `kart conflicts`. [#135](https://github.com/koordinates/kart/issues/135)
```
olsen232, line 9: We should probably also mention that the feature IDs have changed in GEOJSON diff output

olsen232 (approved):

CHANGELOG.md
amal-thundiyil: Thanks for adding [it](https://github.com/koordinates/kart/commit/e168985271eca136aa12490d89753d177798cccb).

PR #604, opened 2022-04-06
Add basic WC tile diffs for point cloud

olsen232:
This uses an index called "worktree-index" to keep track of what has been written to the user's working copy.
Diffs so far are basic and only show the SHA and size of the file itself - the PC content of the file is not yet inspected to generate richer diffs.


kart/diff_util.py
```diff
@@ -1,12 +1,77 @@
...
+        return index.diff_to_workdir(pygit2.GIT_DIFF_INCLUDE_UNTRACKED)
+
+    @functools.lru_cache(maxsize=1)
+    def workdir_deltas_by_ds_path(self):
+        """Returns all the deltas from self.workdir_diff() but grouped by dataset path."""
+        with_and_without_slash = [
+            (p.rstrip("/") + "/", p.rstrip("/")) for p in self.all_ds_paths
+        ]
+
+        def find_ds_path(delta):
```
craigds, line 61: Could this function be reduced to basically `return path.split('/.')[0]`, or more explicitly, `re.split(r'/\.(table|point-cloud)-dataset', path)[0]` ?

kart/point_cloud/checkout.py
```diff
@@ -29,6 +47,12 @@ def reset_wc_if_needed(repo):
                 continue
             shutil.copy(lfs_path, wc_tiles_dir / tilename)
 
+            try:
+                args = ["git", "add", dataset.path]
```
craigds, line 51: this is adding the dataset path to the index for each _tile_ right? should it be un-indented by one level? Or possibly run it once with all point-cloud dataset paths as args.

should this have `stdout=subprocess.DEVNULL` and probably stderr too?

kart/point_cloud/v1.py
```diff
@@ -84,13 +102,41 @@ def diff_tile(self, other, tile_filter=FeatureKeyFilter.MATCH_ALL, reverse=False
...
+        if not deltas:
+            return
+
+        def encode_tile_path_relative(tilename):
+            return self.encode_tile_path(tilename, relative=True)
+
+        yield from self.decode_raw_deltas(
+            deltas,
+            tile_filter,
+            old_path_decoder=encode_tile_path_relative,
```
craigds, line 132: might be a terminology conflict here since you're passing an `encode_` function to kwarg called `_decoder`, if that's correct them maybe it  needs a comment? Or rename something :)

kart/point_cloud/v1.py
```diff
@@ -84,13 +102,41 @@ def diff_tile(self, other, tile_filter=FeatureKeyFilter.MATCH_ALL, reverse=False
...
+
+    def diff_tile_to_wc(self, wc_diff_context, tile_filter):
+        deltas = wc_diff_context.workdir_deltas_by_ds_path().get(self.path)
+        if not deltas:
+            return
+
+        def encode_tile_path_relative(tilename):
+            return self.encode_tile_path(tilename, relative=True)
+
+        yield from self.decode_raw_deltas(
```
craigds, line 129: this call could use some explaining overall I think. Maybe a docstring for diff_tile_to_wc or a comment here explaining why the various decoder things are the right things.

I don't love the idea of having 5-6 decoder callbacks as kwargs here, especially because it seems like a PC dataset is likely to use the same 5-6 callbacks every time (is it?) I can't quite follow it

Would a less callback-focused approach would be better, e.g. a DatasetCheckoutStrategy class with some default methods overridden by a PointCloudV1CheckoutStrategy class. Just a suggestion

kart/base_diff_writer.py
```diff
@@ -178,11 +179,11 @@ def _get_common_ancestor(cls, repo, rs1, rs2):
         return repo.structure(ancestor_id)
 
     @classmethod
-    def _all_feature_keys(cls, old_feature, new_feature):
+    def _all_dict_keys(cls, old_dict, new_dict):
```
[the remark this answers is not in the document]
olsen232: The proposed change doesn't preserve the ordering of fields in diff text-output, which we should probably preserve.
I'm open to making this code better somehow but it's existing code, not relevant to point clouds, so will leave as-is for now

kart/diff_util.py
olsen232: No - the ODB has helpful marker directories that tell you when you are entering a dataset, which have dots in them meaning "I am hidden".
This code deals with the user-visible file-based working-copy (aka workdir aka worktree, we probably need to standardise a bit). That doesn't have the markers, or any hidden folders - so the way we can tell which files belong in which datasets is by seeing is by getting a list of dataset paths from the ODB side of things, and then comparing delta paths to those dataset paths to see which dataset each one is inside, if any.

kart/point_cloud/checkout.py
olsen232: Unindented by one level, silenced stdout. Left stderr intact.

kart/point_cloud/v1.py
olsen232: Yeah this got a bit weird. Will tidy up some names and add some comments.

(Vector datasets have paths which *encode* to primary keys, and *decode* back to paths again. Vector encoder / decoder doesn't deal with the working copy, which doesn't encode primary keys, and is in a separate part of Kart.
Whereas: for point cloud datasets, tiles have complicated paths (for reasons of limiting tree-size)... which... decode(?) to their names? And encode back to their complicated paths? But also, in the working copy, tiles have a different path... the tidy encode / decode system started breaking down a bit)

kart/point_cloud/v1.py
olsen232: Again, I will tidy this up a bit now.

There is a pattern for generating diffs which lots of datasets will use:
- take some specific subtree of myself and of a similar dataset (ie self.inner_tree / "meta", other.inner_tree "meta")
- run self_subtree.diff_to_tree(other_subtree) to get a pygit2.Diff
- go through all the resulting (insert, update, delete) deltas
- fix up the paths to be proper relative paths again (ie, prepend "meta/" onto them all
- run some transform on the paths to decide what to call each object (eg decode primary key)
- run some transform on the path to get the content of each object (eg, read and decode feature)
- yield as kart deltas

(I think the above looks like good-ish documentation - I might check it in).

So, that's what diff_subtrees does - it takes the key and value transforms that it needs to apply as callbacks. It could use a different callback strategy eg polymorphism but I'm not feeling inspired - I won't change it right now.

That code is quite similar to the sort of thing we need to do to generate a working copy diff for file-based datasets:
- run some magic index-diffing code to get a pygit2.Diff of changes that the user made to the working copy folder
- go through all the resulting (insert, update, delete) deltas
- fix up the paths, maybe. the paths will all be working-copy paths - we might want the old-paths to be fixed up so that they are dataset paths instead. the dataset paths can and do have a different structure to the wc-paths 
- run some transform on all the paths to decide what to call each object (eg just show its filename)
- run some transform on the path to get the content of each object (eg get its hash/size, for now)
- yield as kart deltas

So, I've reused the same code for both by setting lots of callbacks. But I think I can maybe make it clearer if I make a working-copy-diff specific method that gives the callbacks better names, maybe takes less callbacks.
Again, this could be polymorphic and maybe it should be but it's not yet obvious to me how I can make it polymorphic and also definitely better

PR #605, opened 2022-04-07
Basic support for committing point cloud tile diffs

olsen232:
For each tile diff, commits it by
- copying the tile to the local LFS cache
- writing a pointer-file to the ODB
- updating the worktree-index so the diff disappears

Also makes sure that only tiles/*.laz are committable -
- QGIS likes to create random files next to tiles, that we don't want to commit

### DIFF
--- a/kart/lfs_util.py
+++ b/kart/lfs_util.py
@@ -22,16 +22,45 @@ def get_hash_and_size_of_file(path):
 
     size = path.stat().st_size
     sha256 = hashlib.sha256()
-    with open(str(path), "rb") as input:
+    with open(str(path), "rb") as src:
         while True:
-            data = input.read(_BUF_SIZE)
+            data = src.read(_BUF_SIZE)
             if not data:
                 break
             sha256.update(data)
     return sha256.hexdigest(), size
 
 
-def pointer_file_to_json(pointer_file_bytes, result=None):
+def get_hash_and_size_of_file_while_copying(src_path, dest_path, allow_overwrite=False):
+    """
+    Given a path to a file, calculates and returns its SHA256 hash and length in bytes,
+    while copying it to the given destination.
+    """
+    if not isinstance(src_path, Path):
+        src_path = Path(src_path)
+    assert src_path.is_file()
+
+    if not isinstance(dest_path, Path):
+        dest_path = Path(dest_path)
+
+    if allow_overwrite:
+        assert not dest_path.is_dir()
+    else:
+        assert not dest_path.exists()
+
+    size = src_path.stat().st_size
+    sha256 = hashlib.sha256()
+    with open(str(src_path), "rb"), open(str(dest_path)) as (src, dest):
+        while True:
+            data = input.read(_BUF_SIZE)
+            if not data:
+                break
+            sha256.update(data)
+            dest.write(data)
+    return sha256.hexdigest(), size
+
+
+def pointer_file_bytes_to_dict(pointer_file_bytes, result=None):
     if isinstance(pointer_file_bytes, pygit2.Blob):
         pointer_file_bytes = pointer_file_bytes.data
     pointer_file_str = pointer_file_bytes.decode("utf8")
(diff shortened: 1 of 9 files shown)


kart/lfs_util.py
```diff
@@ -49,6 +78,26 @@ def pointer_file_to_json(pointer_file_bytes, result=None):
...
+    for key, value in sorted(pointer_dict.items(), key=sort_key):
+        # TODO - LFS doesn't support our fancy pointer files yet. Hopefully fix this in LFS.
+        if key not in ("version", "oid", "size"):
+            continue
+        blob += key.encode("utf8")
+        blob += b" "
+        blob += str(value).encode("utf8")
+        blob += b"\n"
+
+    return blob
```
rcoup, line 98:
```suggestion
    blob = bytearray()
    for key, value in sorted(pointer_dict.items(), key=lambda i: (i[0] != "version", i)):
        # TODO - LFS doesn't support our fancy pointer files yet. Hopefully fix this in LFS.
        if key not in ("version", "oid", "size"):
            continue
        blob += f"{key} {value}\n".encode("utf8")

    return blob
```

rcoup: > `if key not in ("version", "oid", "size"):`

This ignores all the `ext-*` ones, but I guess for now it's ok to ignore them? Nothing else will be writing Kart repos

kart/lfs_util.py
```diff
@@ -61,6 +110,8 @@ def get_hash_from_pointer_file(pointer_file_bytes):
 
 def get_local_path_from_lfs_hash(repo, lfs_hash):
     """Given a sha256 LFS hash, finds where the object would be stored in the local LFS cache."""
+    if lfs_hash.startswith("sha256:"):
+        lfs_hash = lfs_hash[len("sha256:") :]
```
rcoup, line 114:
```suggestion
        lfs_hash = lfs_hash[len("sha256:"):]
```

Or:

```suggestion
        lfs_hash = lfs_hash[7:]  # len("sha256:")
```

rcoup (approved):
looks good 👍

kart/lfs_util.py
```diff
@@ -22,16 +22,45 @@ def get_hash_and_size_of_file(path):
...
+                break
+            sha256.update(data)
+    return sha256.hexdigest(), size
+
+
+def get_hash_and_size_of_file_while_copying(src_path, dest_path, allow_overwrite=False):
+    """
+    Given a path to a file, calculates and returns its SHA256 hash and length in bytes,
+    while copying it to the given destination.
+    """
```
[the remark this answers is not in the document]
olsen232: Yeah to be honest I haven't profiled anything so this is probably a premature optimisation. Changed it to 1MB

kart/lfs_util.py
olsen232: Yeah - this is a temporary workaround anyway. Plenty of time to fix it if we start writing extensions into pointer files

kart/lfs_util.py
olsen232: Went with the second. Black doesn't like the first one

olsen232:
The diffs (as they are currently) are visible in test_point_cloud, here they are:
```
--- auckland:tile:auckland_1_1.copc.laz
+++ auckland:tile:auckland_1_1.copc.laz
-                                      oid = sha256:1d9934b50ebd057d893281813fda8deffb0ad03b3c354ec1c5d7557134c40be1
+                                      oid = sha256:dafd2ed5671190433ca1e7cea364a94d9e00c11f0a7b3927ce93554df5b1cd5c
-                                     size = 23570
+                                     size = 68665
--- auckland:tile:auckland_3_3.copc.laz
-                                     name = auckland_3_3.copc.laz
-                                      oid = sha256:522ef2ff7f66b51516021cde1fa7b9f301acde6713772958d6f1303fdac40c25
-                                     size = 1334
+++ auckland:tile:auckland_4_4.copc.laz
+                                     name = auckland_4_4.copc.laz
+                                      oid = sha256:522ef2ff7f66b51516021cde1fa7b9f301acde6713772958d6f1303fdac40c25
+                                     size = 1334
```

These diffs are somewhat limited -
- you can tell which tiles have been inserted, updated, deleted
- you can even find renames if you look at the hashes, but the code doesn't do it for you yet
- you *can't* tell anything else about the tiles from the diffs
  - this is partly because I haven't yet written the code to use PDAL to check what's in the files that the user is trying to add
  - and partly because I am currently not storing any tile metadata in pointer files, due it not being LFS compatible and I haven't yet implemented a workaround
  
More work still to be done of course

Issue #609, opened 2022-04-17
FR: A way to list tables that are in the working copy area but are not tracked by kart


olsen232:
It's possible for the user to create tables inside the working copy that Kart uses, but if they do so, Kart will never take any notice of them, and there is currently no correct way to add them to the Kart repository (you could try importing such a table, but unless you rename it during import, the imported data will collide with the original table, which will cause the command to fail and add to the user's confusion). See https://github.com/koordinates/kart/issues/249

It is possible to get a list of the tables that Kart is currently managing (with `kart data ls`) but there is no way for the user to get a list of any tables in the WC that Kart is not tracking.
This would be a useful first step to implementing https://github.com/koordinates/kart/issues/249
Perhaps `kart status --list-untracked-tables`

There is code for listing tables for each of the four currently-supported types of working copy - see:
https://github.com/koordinates/kart/search?q=list_tables&type=code
It should be possible to make this change with very little or no code that is specific to a certain working copy type, if you use this interface that already exists.

Zumitify:
Thanks @olsen232, for creating this new enhancement issue. 

> to list untracked files you might just want to add something to the various working copy classes to directly list tables from the database, and then filter out ones that are already datasets.

The approach mentioned in this [comment](https://github.com/koordinates/kart/issues/249#issuecomment-1100939800) by @craigds seems like the most straightforward way to tackle this.
Thanks.

PR #620, opened 2022-05-10
Added option list-untracked-tables in status

Zumitify:
## Description
- Added option `--list-untracked-tables` under `kart status` command to list the tables/datasets that are added to the repository but are not tracked by Kart.
- This works for Geopackage Woking Copy format

## Related links:
- Partially resolves: https://github.com/koordinates/kart/issues/609

## Unit testing:
- Init a kart repo with a table.
- Added another package layer using QGIS-> Database Tools.
- `kart data ls` does not show all the newly added tables, because they are untracked.
```
E:\self_study\GSoC22\KART\kart_tutorial\buildings-project>kart data ls
wellington_building_outline
```
- Output of `kart status --list-untracked-tables`
```
E:\self_study\GSoC22\KART\kart_tutorial\buildings-project>kart status --list-untracked-tables
code_point_open
layer_styles
```
## Checklist:

## To-do:
- [ ] Add support for MySQL, PostGIS and SQL Server working copy formats.


kart/status.py
```diff
@@ -45,7 +52,23 @@ def status(ctx, output_format):
     if output_format == "json":
         dump_json_output({"kart.status/v1": jdict}, sys.stdout)
     else:
-        click.echo(status_to_text(jdict))
+        if list_untracked_tables:
+            working_copy = repo.working_copy
+            location = repo.workingcopy_location
+            wc_type = str(WorkingCopyType.from_location(location)).lstrip("WorkingCopyType.")
```
olsen232, line 58: There is a simpler way - some of this code can be replaced with the following:
```
with working_copy.session() as sess:
    ret_val = working_copy.adapter.list_tables(sess)
```

kart/status.py
```diff
@@ -22,7 +23,13 @@
     type=click.Choice(["text", "json"]),
     default="text",
 )
-def status(ctx, output_format):
+
+@click.option(
+    "--list-untracked-tables",
+    is_flag=True,
+    help="Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.",
```
olsen232, line 30: This help text is wrong

kart/status.py
```diff
@@ -22,7 +23,13 @@
     type=click.Choice(["text", "json"]),
     default="text",
 )
-def status(ctx, output_format):
+
+@click.option(
+    "--list-untracked-tables",
```
olsen232, line 28: This option has slightly weird behaviour - 
- `kart status` gathers the status of the kart repo and prints it as text
- `kart status --list-untracked-tables` gathers the status of the kart repo, then discards it, and lists the untracked tables
- `kart status --list-untracked-tables -o json` gathers the status of the kart repo and prints it as json, the list-untracked-tables option has no effect

Ideally, since this is an option and not a new command, we would probably want to *add* to the output, both for text and JSON output - ie, running `kart status --list-untracked-tables` would show you the kart status plus the untracked tables. You might feel that this is beyond the scope of this PR though - for now I will be happy if the list_untracked_tables option is checked first, so that we don't do the unnecessary work of generating the rest of the status and then discarding it

PR #621, opened 2022-05-10
Upgrade GDAL to 3.4.0

olsen232:
Pull GDAL 3.4.0 into Kart instead of 3.3.2

Currently makes no discernible difference - may fix some bugs, see release notes [here](https://github.com/OSGeo/gdal/releases/tag/v3.4.0) - but will be required later for point-cloud support via PDAL.

## Related links:

### DIFF
--- a/vendor/gdal/Makefile
+++ b/vendor/gdal/Makefile
@@ -162,7 +162,9 @@ $(configure-gdal): | src $(libs) $(ccache-deps)
 		--without-jasper \
 		--without-jpeg \
 		--without-jpeg12 \
+		--without-jxl \
 		--without-lerc \
+		--without-lz4 \
 		--without-netcdf \
 		--without-odbc \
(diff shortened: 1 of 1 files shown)


vendor/gdal/Makefile
```diff
@@ -42,6 +42,8 @@ ifeq ($(PLATFORM),Darwin)
 SQLITE_PREFIX = /usr/local/opt/sqlite
 JSONC_PREFIX = /usr/local/opt/json-c
 PG_PREFIX = /usr/local/opt/libpq
+JXL_PREFIX = /usr/local/opt/jpeg-xl
+LZ4_PREFIX = /usr/local/opt/lz4
```
rcoup, line 46: Can we drop this if we're doing `--without-jxl`/`--without-lz4`?

rcoup (approved):
👍

PR #623, opened 2022-05-13
add ext-run command for extension invocation

rcoup:
It can be useful to run scripts inside Kart's execution environment for prototyping or extensions. Add a new command, `kart ext-run PY_SCRIPT [ARGS...]` to do this.

It uses importlib to import and validate (as best we can) the script before calling a `main(ctx, args)` function.

example script:

```console
$ cat >plugin.py <<EOF
import kart

def main(ctx, args):
  print("hi from a plugin!", args)
  print("am i frozen?", kart.is_frozen)
EOF

$ kart -C /path/to/repo ext-run plugin.py arg1 arg2
hi from a plugin! ('arg1', 'arg2')
am i frozen? None
```

Obviously Kart doesn't have a public API, so it's all on the user to keep any extensions working.

## Checklist:

### DIFF
--- a/kart/cli.py
+++ b/kart/cli.py
@@ -1,10 +1,14 @@
 #!/usr/bin/env python3
 import importlib
+import importlib.util
+import inspect
 import logging
 import os
+import pathlib
 import re
 import subprocess
 import sys
+import traceback
 
 import click
 import pygit2
@@ -217,7 +221,7 @@ def cli(ctx, repo_dir, verbose, post_mortem):
 )
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def push(ctx, do_progress, args):
-    """ Update remote refs along with associated objects """
+    """Update remote refs along with associated objects"""
     ctx.invoke(
         git,
         args=[
@@ -239,7 +243,7 @@ def push(ctx, do_progress, args):
 )
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def fetch(ctx, do_progress, args):
-    """ Download objects and refs from another repository """
+    """Download objects and refs from another repository"""
     ctx.invoke(
         git,
         args=[
@@ -254,7 +258,7 @@ def fetch(ctx, do_progress, args):
 @click.pass_context
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def remote(ctx, args):
-    """ Manage set of tracked repositories """
+    """Manage set of tracked repositories"""
     ctx.invoke(git, args=["remote", *args])
 
 
@@ -262,7 +266,7 @@ def remote(ctx, args):
 @click.pass_context
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def tag(ctx, args):
-    """ Create, list, delete or verify a tag object signed with GPG """
+    """Create, list, delete or verify a tag object signed with GPG"""
     ctx.invoke(git, args=["tag", *args])
 
 
@@ -270,7 +274,7 @@ def tag(ctx, args):
 @click.pass_context
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def reflog(ctx, args):
-    """ Manage reflog information """
+    """Manage reflog information"""
     ctx.invoke(git, args=["reflog", *args])
 
 
@@ -278,7 +282,7 @@ def reflog(ctx, args):
 @click.pass_context
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
 def config(ctx, args):
-    """ Get and set repository or global options """
+    """Get and set repository or global options"""
     ctx.invoke(git, args=["config", *args])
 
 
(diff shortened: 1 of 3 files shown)


pfw:
Would it be worth adding the directory of the ext Python file to sys path so it can import other things?

kart/cli.py
```diff
@@ -318,6 +322,63 @@ def lfs(ctx, args):
...
+        f_sig = inspect.signature(module.main)
+        f_sig.bind(ctx=ctx, args=args)
+    except TypeError:
+        click.secho(
+            f"Invalid function definition for {script}:main()",
+            fg="red",
+            file=sys.stderr,
+        )
+        _print_ext_tb()
+        raise click.Abort()
```
craigds, line 377: is this the same as `ctx.abort()` ? the Abort exception is doc'd as internal-only

kart/cli.py
```diff
@@ -318,6 +322,63 @@ def lfs(ctx, args):
...
+
+    return module.main(ctx=ctx, args=args)
```
craigds, line 379: I think we should enable `main` to be a `click.command`; atm this doesn't work:

```python
import click


@click.command()
@click.pass_context
def main(ctx, args):
    print(args)
```


results in:

```python
$ kart ext-run x.py
...
  File "/Users/cdestigter/checkout/kart/kart/cli.py", line 379, in ext_run
    return module.main(ctx=ctx, args=args)
  File "/Users/cdestigter/checkout/kart/venv/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/cdestigter/checkout/kart/venv/lib/python3.7/site-packages/click/core.py", line 781, in main
    with self.make_context(prog_name, args, **extra) as ctx:
  File "/Users/cdestigter/checkout/kart/venv/lib/python3.7/site-packages/click/core.py", line 698, in make_context
    ctx = Context(self, info_name=info_name, parent=parent, **extra)
TypeError: __init__() got an unexpected keyword argument 'ctx'
```

kart/cli.py
```diff
@@ -318,6 +322,63 @@ def lfs(ctx, args):
...
+    try:
+        f_sig = inspect.signature(module.main)
+        f_sig.bind(ctx=ctx, args=args)
+    except TypeError:
+        click.secho(
+            f"Invalid function definition for {script}:main()",
+            fg="red",
+            file=sys.stderr,
+        )
+        _print_ext_tb()
```
craigds, line 376: Output is a bit weird here:

```
$ kart ext-run x.py
Invalid function definition for x.py:main()
TypeError: got an unexpected keyword argument 'args'
Aborted!
```
Maybe just show the full traceback? Or at least the ext_run frame itself

craigds (changes requested):

craigds:
> Would it be worth adding the directory of the ext Python file to sys path so it can import other things?

I'm tempted to say the script we're calling can just do it itself, but that means it has to be _aware_ it might be run this way. So maybe an explicit `--add-pythonpath=/path/` option? Or the caller could just set PYTHONPATH, assuming that makes it through the pyinstaller bootstrapping (does it?)

pfw:
I believe PYTHONPATH won't make it through the pyinstaller bootstrap. Since the code being called must already know a lot about the internals to run knowing that you need to append to sys.path if you have extra things to import probably is fine. That fits with giving this code flexibility to do whatever it wants as long as it knows the internals.

pfw:
Thinking further... if we do a helper speed up to run this we could allow for specifying expensive imports on the command line so they happen once even for these extensions. Something to keep in mind when doing further investigation on that.

kart/cli.py
pfw: This would require inspecting to see if `main` was decorated as a click command in the `ext-run` code as extensions could add `@click.option` and `@click.argument` and so get named or positional parameters to the main function, correct?

It would be nice to be able to do that and potentially have a group of commands but I think that would require 'registering' the extension somehow so it is imported and becomes part of the click hierarchy - something like `kart --ext x.py mycommand --frob==wongle`, where `--ext x.py` just imports that file and doesn't run anything directly itself. A natural path from that is to allow multiple `--ext` options and actually fits better with the forking model for speed ups - it would come for free.

rcoup: I think there's a difference between a plugin (`kart myplugin`) which needs registering into click in advance for help/arg handling; and an extension, which just invokes a script. This is the latter, the former is something we might want in future.

If the extension wants to setup click/argparse/whatever, then good on it.

pfw: Yep. The distinction is tricky there as either way the code has to fit a particular model - to support click all code must use it etc. and likewise for no click support it must do it itself, plug-ins give more scope to make opinionated choices on capability.

rcoup:
I've added it for now, I don't see a _huge_ number of downsides, and it makes it simpler for actually writing extensions.

kart/cli.py
rcoup: I passed the click context purely because some stuff that might have already been parsed globally could be useful: verbosity, `-c git.config.var=x`, whatever. Wasn't really for _actual_ use 😄

pfw (approved):

kart/cli.py
craigds: yep, nbd, I guess all that's needed is a tiny wrapper like:

```python

def main(*, ctx, **kw):
    ctx.invoke(actualmain, **kw)
```

PR #627, opened 2022-05-20
Store PC metadata in LFS pointer files

olsen232:
Allows for better tile diffs including commit<>commit diffs,
since we can turn pointer-file change into a human-readable tile-change,
regardless of whether we actually have the PC tile cached at present.
See improved diff output at test_point_clouds.py

Currently squashes metadata into an LFS extension - this can change if
we find a better solution / modify git LFS.

The actual metadata that we store is subject to change, but this
gets the pieces in place so that we can do so.

Another tweak:
Always store the CRS in "crs.wkt" - this is simpler than the current
system, which stores it by name, which was needed for vectors but adds
unnecessary complexity here.


kart/point_cloud/import_.py
```diff
@@ -258,16 +176,15 @@ def _format_array(array):
 
 
 def get_copc_version(info):
-    vlr_0 = info.get("vlr_0")
-    if vlr_0:
-        user_id = vlr_0.get("user_id")
-        if user_id == "copc":
-            return vlr_0.get("record_id")
-    return NOT_COPC
+    # TODO - find out how to extract the COPC version number using the current PDAL.
```
rcoup, line 179: https://copc.io/

The first (E)VLR in the file (COPC info VLR) should have UserID=`copc` & RecordID of `1` (it's a little endian unsigned short : 2 bytes)

Which should be the same as:
*  verify that the bytes at offsets 393 and 394 contain the values 1 and 0, respectively (this is the COPC version number, 1).

kart/point_cloud/import_.py
```diff
@@ -332,32 +280,75 @@ def _convert_tile_to_copc(source, dest):
...
+
+def pc_tile_metadata_to_pointer_metadata(metadata):
+    """
+    Given all the tile-metadata, returns thats which should be written to an LFS pointer file,
+    in the appropriate format.
+    """
+    return {
+        # TODO - available.<URL-IDX> <URL>
+        # TODO - find the real extent.crs84
+        "extent.crs84": _format_array(metadata["native-extent"]),
```
rcoup, line 343: should probably just not return `extent.crs84` until implemented?

`available` is a bit orthogonal to point-clouds, not sure whether we handle it somewhere else?

rcoup (changes requested):
Looking good!

kart/point_cloud/import_.py
olsen232: Not done - will do in a separate change. Returning "1.0" for COPC will do for now (especially since there is only 1 COPC version right now)

I was previously using PDAL to extract this VLR, but that stopped working, but now PDAL is exposing whether or not a PC tile is COPC at a high level. I need to get a hex editor and dig into why the low level way stopped working with a hex editor - is PDAL hiding VLRs from me if it has already handled them at a lower level? Or have I made some mistake somewhere

kart/point_cloud/import_.py
olsen232: Removed extent.crs84, removed the available TODO

kart/lfs_util.py
```diff
@@ -68,36 +72,68 @@ def get_hash_and_size_of_file_while_copying(src_path, dest_path, allow_overwrite
...
         if not line:
             continue
         parts = line.split(" ", maxsplit=1)
         if len(parts) < 2:
             L.warn(f"Error parsing pointer file:\n{line}")
-            continue
         key, value = parts
-        result[key] = value
+        if key.startswith("ext-0-kart-encoded="):
+            result.update(_decode_extra_values(key[len("ext-0-kart-encoded=") :]))
```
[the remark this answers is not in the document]
olsen232: Done

kart/lfs_util.py
```diff
@@ -68,36 +72,68 @@ def get_hash_and_size_of_file_while_copying(src_path, dest_path, allow_overwrite
...
+        return _dict_to_pointer_file_bytes_simple(pointer_dict)
+
+    extra_values = dict(
+        (k, v) for k, v in sorted(pointer_dict.items()) if k not in _STANDARD_LFS_KEYS
+    )
+    encoded_extra_values = _encode_extra_values(extra_values)
+
+    result = (
+        f"version {pointer_dict['version']}\n"
+        f"ext-0-kart-encoded={encoded_extra_values} {_EMPTY_SHA256}\n"
```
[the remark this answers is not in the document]
olsen232: Done

kart/lfs_util.py
```diff
@@ -68,36 +72,68 @@ def get_hash_and_size_of_file_while_copying(src_path, dest_path, allow_overwrite
...
+        f"oid {pointer_dict['oid']}\n"
+        f"size {pointer_dict['size']}\n"
+    )
+    return result.encode("utf8")
+
+
+def _encode_extra_values(extra_values):
+    return b64encode_str(msg_pack(extra_values))
```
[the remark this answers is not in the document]
olsen232: Done

kart/lfs_util.py
```diff
@@ -68,36 +72,68 @@ def get_hash_and_size_of_file_while_copying(src_path, dest_path, allow_overwrite
...
+
+    if not only_standard_keys or not (pointer_dict.keys() - _STANDARD_LFS_KEYS):
+        return _dict_to_pointer_file_bytes_simple(pointer_dict)
+
+    extra_values = dict(
+        (k, v) for k, v in sorted(pointer_dict.items()) if k not in _STANDARD_LFS_KEYS
+    )
+    encoded_extra_values = _encode_extra_values(extra_values)
+
+    result = (
```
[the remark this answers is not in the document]
olsen232: Done

kart/point_cloud/import_.py
olsen232: PDAL now hides the COPC VLRs from me - see
https://github.com/PDAL/PDAL/blob/3e33800d85d48f726dcd0931cefe062c4af2b573/io/private/las/Vlr.cpp#L53
But does expose whether or not a tile "is COPC" - as the code now checks. Just not the COPC version number (which may one day be useful, but is low priority right now)

rcoup: I think because COPC VLRs are content-dependent in general it's bad for them to flow through PDAL pipelines and that's been disabled. Not sure whether there's another option for querying.

Yeah, is low priority

PR #629, opened 2022-05-24
Add CRS 84 extents (EPSG:4326) to pointer file metadata

olsen232:
Uses PDAL filter.stats to do the conversion.
Remove code that corrects native extents using scale and offset - tests
show that this is not necessary, or no longer necessary, since PDAL
is (now?) doing this for us.


kart/point_cloud/import_.py
```diff
@@ -291,7 +294,7 @@ def extract_pc_tile_metadata(pc_tile_path, extract_schema=False):
         {
             "type": "readers.las",
             "filename": str(pc_tile_path),
-            "count": 0,  # Don't read any individual points.
+            "count": 10,  # Don't read any individual points.
```
rcoup, line 297: magic number has a meaning?

rcoup (approved):
👍

kart/point_cloud/import_.py
olsen232: Oops nope - was testing what the stats filter did (ie does it read the extent from the header, or calculate an extent from the points read). This should still be zero. Reverted.

Issue #632, opened 2022-05-27
Point-cloud dataset schemas are inconsistent with vector/table schemas


craigds:
`kart meta get {dataset} schema.json` in a vector/table repo looks like this:

```json
[
  {
    "id": "e97b4015-2765-3a33-b174-2ece5c33343b",
    "name": "fid",
    "dataType": "integer",
    "primaryKeyIndex": 0,
    "size": 64
  },
  {
    "id": "f488ae9b-6e15-1fe3-0bda-e0d5d38ea69e",
    "name": "geom",
    "dataType": "geometry",
    "geometryType": "POINT",
    "geometryCRS": "EPSG:4326"
  }
]
```

For a point-cloud dataset, it's subtly different:

```json
{
  "dimensions": [
    {
      "name": "X",
      "size": 8,
      "type": "floating"
    },
    {
      "name": "Y",
      "size": 8,
      "type": "floating"
    },
  ]
}
```

The differences are:

* no `id` per field (probably this is expected & okay)
* no `primaryKeyIndex` (also expected & okay)
* top-level object with `dimensions` key instead of a top-level list
* `size` is in bytes, rather than bits
* `type` key instead of `dataType`
* the types aren't the same as [the vector types](https://docs.kartproject.org/en/latest/pages/development/datasets_v3.html#data-types)

Except for the first two, the other differences seem quite avoidable and worth fixing

PR #633, opened 2022-05-27
Convert schema JSON that PDAL provides to Kart format

olsen232:
PDAL exposes the point-data-record-format in terms of a list of fields - the fields are described with their names and data types. We already have this concept in kart - the "schema.json" meta item - but in a slightly different format. This massages the data PDAL gives us to be in the Kart format.

Fixes https://github.com/koordinates/kart/issues/632


craigds (approved):
thanks! LGTM

PR #635, opened 2022-05-29
Allow missing WC parts.

olsen232:
Work on making Kart more lenient if some part of the WC are missing when
it tries to do a WC diff, since it will soon be common for only
part of the WC to be present (ie a repo with only tabular data
generally has no PC WC, and a repo with only PC data may have
no tabular WC).

Currently has no real effect since there is only one part that is
implemented - the tabular part - but is part of the tidy up /
refactor for implementing the PC part too.

### DIFF
--- a/kart/diff_util.py
+++ b/kart/diff_util.py
@@ -1,6 +1,7 @@
 import functools
 import logging
 
+import click
 import pygit2
 
 from .diff_structs import DatasetDiff, RepoDiff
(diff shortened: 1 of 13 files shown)


kart/diff_util.py
```diff
@@ -30,9 +31,13 @@ def __init__(self, repo, all_ds_paths=None):
...
 
     @property
-    def table_working_copy(self):
+    def working_copy(self):
         assert not self.is_bare
-        return self.repo.working_copy.tabular
+        return self.repo.working_copy
+
+    @property
+    def wc(self):
```
craigds, line 39: IMO we don't need these aliases. Just use `working_copy`. Adding extra aliases isn't actually helping with readability

kart/fsck.py
```diff
@@ -111,7 +112,7 @@ def fsck(ctx, reset_datasets, fsck_args):
             )
             click.echo(f"{track_count} rows marked as changed in working-copy")
 
-            wc_diff = table_wc.diff_db_to_tree(dataset)
+            wc_diff = table_wc.diff_ds_to_wc(dataset)
```
craigds, line 115: same here imo: `diff_dataset_to_working_copy`

craigds (approved):

PR #642, opened 2022-06-02
Formalise post-commit WC operations

olsen232:
After committing we need to do any/all of the following:
- update the WC to be tracking the latest commit-tree
- clear some PKs from the tracking table so they are no longer marked
  as dirty
- drop some features from the WC if they are outside the spatial filter

This adds that compound-operation to the WC, and tidies up the names
etc of some of the sub-operations (eg mark_as_clean is clearer than
reset_tracking_table). It also switches these methods to take key_filters
instead of ad-hoc structures like dicts or lists, which makes them
a little more self-documenting.

### DIFF
--- a/kart/working_copy.py
+++ b/kart/working_copy.py
@@ -242,7 +242,7 @@ class WorkingCopy:
                 rewrite_full=rewrite_full,
             )
 
-    def set_state_post_commit(
+    def soft_reset_after_commit(
         self,
         commit_or_tree,
         *,
@@ -273,7 +273,7 @@ class WorkingCopy:
             )
 
         for p in self.parts():
-            p.set_state_post_commit(
+            p.soft_reset_after_commit(
                 commit_or_tree,
                 mark_as_clean=mark_as_clean,
                 now_outside_spatial_filter=now_outside_spatial_filter,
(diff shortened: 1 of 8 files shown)


kart/working_copy.py
```diff
@@ -240,6 +242,43 @@ def reset(
                 rewrite_full=rewrite_full,
             )
 
+    def set_state_post_commit(
```
craigds, line 245: minor: I'd prefer `set_state_after_commit` because the word 'post' might mean some kind of action. However... is this the same as a 'soft reset' in git terminology? If so, what about we just call it that?

craigds (approved):

kart/working_copy.py
olsen232: Changed to `soft_reset_after_commit`.
(I don't love git's name for things - hard, soft and mixed resets are simply the name for those type of resets. What they mean simply to be learned since they don't, for instance, reset the hard/soft/mixed area of your working copy. Probably those resets should have been called things like `unstage`, `checkout --discard-changes`, `rebase-changes` or... I don't know. But I guess "set_state" also means nothing so soft_reset is at least step up from that. I've left in the "after_commit" since this is specifically the type of soft_reset we do after a commit - that's why we're clearing from the tracking table and remove features that don't match spatial filters. There's nothing in Git that's an exact correspondence)

PR #646, opened 2022-06-09
Fix primary key issues for shapefile import

olsen232:
- Don't use FID as a primary key, don't do anything with FID
- Use any other field as the primary key if specified, otherwise,
  generate a new primary key.
- Make tests check both of these behaviours
- Delete tests that convert GPKG to SHP files to use as test data -
  checked in equivalent test data. The auto-converted test data
  was bad anyway: it had lost the primary key columns.

Fix the PkGeneratingTableImportSource - it shouldn't modify the
schema of the delegate import-source that it wraps. (Modifying
the schema of an import-source in place is allowed, as long
as you only change the IDs of the columns - these IDs are
Kart-internal and so are arbitrary for an import-source until
and unless we align them to an existing table. But, other
modifications to the import-source schema - such as adding
a new `auto_pk` column - can cause the import-source to crash,
so this must be done at the outer wrapper class only - in
the PkGeneratingTableImportSource).

## Checklist:


kart/tabular/ogr_import_source.py
```diff
@@ -316,20 +315,23 @@ def has_geometry(self):
...
 
+            for i in range(ld.GetFieldCount()):
+                field = ld.GetFieldDefn(i)
+                if primary_key_name == field.GetName():
+                    return primary_key_name
+
             if primary_key_name == self.ogrlayer.GetFIDColumn():
                 # OGR automatically turns 'ogc_fid' column in postgres into an FID,
                 # and removes it from the list of fields below.
+                self.use_ogc_fid_as_pk = True
```
craigds, line 332: Probably it shouldn't? We could set the [`PGSQL_OGR_FID` config option](https://gdal.org/drivers/vector/pg.html#configuration-options) to some random UUID to prevent it doing this. Then we could just treat an `ogc_fid` field as a regular field

craigds (approved):

kart/tabular/ogr_import_source.py
olsen232: ... yeah we could do that. But we're not using this except for shapefiles right now so the comment is out of date and the behaviour for non-shapefiles isn't very important. I think best not to change this code too much unless it fixes a bug or we start using it for another OGR import type.

Issue #643, opened 2022-06-06
Tab Completion for Kart


amal-thundiyil:
Kart could do better to support CLI users as they interact with the commands and the data in their repositories. Adding tab-completion so it works smoothly and consistently for Kart commands and their options would make this better. Then expanding that further so datasets, branches, tags, files, metadata, features, and other context-relevant information is also presented where appropriate.

PR #648, opened 2022-06-09
Add `--install-tab-completion` option for installing shell completion

amal-thundiyil:
## Description

The user gets four options to choose from bash, zsh, fish and auto (auto-detects shell using `shellingham`). To activate shell completion run:

```sh
kart config --install-completion auto
```

Replace `auto` with desired shell eg:

```sh
kart config --install-completion zsh
```
Completion will take effect once you restart the terminal

## Related links:

- #643 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.4 (UNRELEASED)
 
+- Add `--install-tab-completion` option for installing shell completion. [#643](https://github.com/koordinates/kart/issues/643)
 - Add tab completion for kart commands and user-specific data. [#643](https://github.com/koordinates/kart/issues/643)
 - Changed format of feature IDs in GeoJSON output to be more informative and consistent. [#135](https://github.com/koordinates/kart/issues/135)
 - Fixed primary key issues for shapefile import - now generates an `auto_pk` by default, but uses an existing field if specified (doesn't use the FID). [#646](https://github.com/koordinates/kart/pull/646)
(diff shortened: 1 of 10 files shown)


kart/completion.py
```diff
@@ -0,0 +1,153 @@
...
+    parent_dir: Path = path_obj.parent
+    parent_dir.mkdir(parents=True, exist_ok=True)
+    script_content = get_completion_script(
+        prog_name=prog_name, complete_var=complete_var, shell=shell
+    )
+    path_obj.write_text(f"{script_content}\n")
+    return path_obj
+
+
+def install(
```
olsen232, line 112: I see there's an install method in click_completion that maybe does something like this and that you don't seem to be calling - but it looks like maybe it's not as good as your one, ie, it doesn't check if it's already installed. Can you confirm?

olsen232 (commented):
Looking good! Here's my comments, nothing major

kart/completion.py
amal-thundiyil: Yes, there's code in [`click-contrib/click-completion`](https://github.com/click-contrib/click-completion/) which does something similar to this (ref [code](https://github.com/click-contrib/click-completion/blob/a9b705d4f4df4579a96851b2b14b0c59765ef6ce/click_completion/core.py#L319-L381)). I have decided to not use [`click-contrib/click-completion`](https://github.com/click-contrib/click-completion/) because it seems to predate [Click's shell compeletion](https://click.palletsprojects.com/en/8.1.x/shell-completion/) capabilites (see [click-contrib/click-completion#41 (issue)](https://github.com/click-contrib/click-completion/issues/41)).

CHANGELOG.md
```diff
@@ -6,9 +6,10 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.4 (UNRELEASED)
 
+- Add `--install-completion` option for installing shell completion. [#643](https://github.com/koordinates/kart/issues/643)
```
olsen232, line 9: This is out of date now that it's called `--install-tab-completion`

olsen232 (approved):

requirements/licenses.ini
```diff
@@ -9,6 +9,7 @@ authorized_licenses:
     mozilla public license 2.0 (mpl 2.0)
     PSF
     Python Software Foundation
+    ISC License (ISCL)
```
craigds, line 12: @rcoup this looks ok to me but you asked to be in the loop on license changes. 👍 / 👎  ?

rcoup: Thanks. Yes, all good 👍

PR #649, opened 2022-06-10
Add --with-dataset-types to `kart meta get`

olsen232:
Simply adds two fields to each dataset's meta output:
"datasetType" and "version"
datasetType is always a string, and version is always an integer. 

Existing datasets will mostly be datasetType="table", version=3
Experimental point cloud datasets are datasetType="point-cloud" version=1


olsen232:
cc @hamishcampbell @UberMouse

PR #653, opened 2022-06-16
Start of PC meta changes - store fileInfo

olsen232:
If we want the file type to be a constrained part of the dataset,
we should store what it is and show diffs when it changes.
This standardises the different flavours of file types as a "meta item",
which looks like so:
{
  "compression": "laz",
  "lasVersion": "1.4",
  "optimization": "copc",
  "optimizationVersion": "1.0",
  "pointDataRecordFormat": 7,
  "pointDataRecordLength": 36
}

Also: since all of this metadata - whether we store in in JSON files,
or encode it into pointer files - forms part of our API since we
expose it in diffs, I've updated the names to be JSON-like everywhere.
This is consistent with our existing diffs - schema.json has a field
"dataType" not "data-type" or "data.type".

- We could use different names internally eg "extent.crs84" internally
and then switch it to "crs84Extent" externally, but there's no point
(and like Git we tend to leak our internals by exposing lower-level
commands for hacking them if needed).
- We could use differnt names for text based diffs eg "crs84Extent"
for JSON diffs and "CRS-84 Extent" for text diffs, but again there's
no point teaching user's two different names for the same thing.


olsen232:
This is superceded by #655

PR #654, opened 2022-06-22
Support tab completion in powershell for kart

amal-thundiyil:
Added tab completion support for Powershell (or `pwsh`). Run the following command to install:
```sh
kart config --install-tab-completion pwsh # (powershell or auto works too)
```

## Related links:
- #643 

## Checklist:

### DIFF
--- a/kart/completion.py
+++ b/kart/completion.py
@@ -1,9 +1,10 @@
 import os
 import re
 import sys
+import subprocess
 from enum import Enum
 from pathlib import Path
-from typing import Optional, Tuple, Any, Optional
+from typing import Optional, Tuple, Optional, Any, Dict, List
 
 import click
 from click.shell_completion import _SOURCE_BASH, _SOURCE_ZSH, _SOURCE_FISH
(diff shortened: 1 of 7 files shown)


olsen232 (approved):
Looks good overall. Just tidy up the encoding code - it would be nice if it would be correct and compatible (ie, have it work properly for a windows machine using any native encoding, even when the user's homedir isn't pure ASCII) - but whether it's fully compatible or not, it shouldn't decode-by-guessing like it does now

kart/completion.py
```diff
@@ -114,6 +141,39 @@ def install_fish(*, prog_name: str, complete_var: str, shell: str) -> Path:
...
+            "-Command",
+            "Set-ExecutionPolicy",
+            "Unrestricted",
+            "-Scope",
+            "CurrentUser",
+        ]
+    )
+    result = subprocess.run(
+        [shell, "-NoProfile", "-Command", "echo", "$profile"],
+        check=True,
```
olsen232, line 157: check=True means that subprocess.run raises an error if the exit-code is non-zero - either check the return code yourself, or use check=True, but there's no point doing both at once

PR #661, opened 2022-07-05
Kart helper

pfw:
## Description

Add `helper` command to reduce overhead of python startup. Includes a custom PyInstaller bootloader and upgrades to v5.1

Supported on Linux and MacOS, no changes on Windows but will be enabled in the future.

## Checklist:


pfw:
This uses [https://github.com/koordinates/pyinstaller/tree/v5.1-kart]() and I feel that it should be maintained on a branch there, there is no chance of this going back to PyInstaller proper and maintaining a fork over time whilst merging to the main branch is messy. That branch is definitely worth reviewing as it includes all of the client side behaviour.

kart/cli.py
```diff
@@ -233,6 +246,210 @@ def push(ctx, do_progress, args):
     )
 
 
+@cli.command(context_settings=dict(ignore_unknown_options=True))
```
olsen232, line 249: Consider moving all the helper stuff to a separate file. cli.py is getting to be a mix of things - the implementations of any commands that are too small to have their own file, plus the kart-helper, plus some other miscellany, but these things are unrelated to each other.

kart/cli.py
```diff
@@ -233,6 +246,210 @@ def push(ctx, do_progress, args):
...
+
+    # stash the required environment for used libs
+    # TODO - this should be checked to ensure it is all that is needed
+    #  is there anything beyond these which is needed, eg. PWD, USER, etc.
+    required_environment = {
+        k: v
+        for k, v in os.environ.items()
+        if k
+        in [
+            "PATH",
```
olsen232, line 282: Is there a reason we don't copy the entire environment - is it too big?

docs/pages/development/helper.rst
```diff
@@ -0,0 +1,75 @@
+CLI Helper
```
olsen232, line 1: Since this is well-documented (as opposed to a temporary hack that we don't want to advertise), we should probably put something about it in the changelog. You can link to this PR

kart/cli.py
```diff
@@ -233,6 +246,210 @@ def push(ctx, do_progress, args):
...
+    os.umask(0o077)
+    try:
+        # TODO - this will take over a socket from an existing running helper process
+        #  it should check to see if another helper is already running and bail out
+        if os.path.exists(socket_filename):
+            os.unlink(socket_filename)
+
+        sock.bind(str(socket_filename))
+    except OSError as e:
+        print(dir(e))
```
olsen232, line 312: Looks like some debug code left in here

kart/cli.py
```diff
@@ -233,6 +246,210 @@ def push(ctx, do_progress, args):
...
+    import pygments.styles.default
+
+    while True:
+        # The helper will exit if no command received within timeout
+        sock.settimeout(timeout)
+        try:
+            client, info = sock.accept()
+            if os.fork() == 0:
+                payload, fds = recv_fds(client, 8164, 4)
+                # print("kart helper: handlng request...", payload)
```
olsen232, line 342: No need to check in commented out debug-code

kart/cli.py
```diff
@@ -233,6 +246,210 @@ def push(ctx, do_progress, args):
...
+                        # TODO - do we ever see negative exit codes from cli (git etc)?
+                        libc.semctl(
+                            semid, 0, SETVAL, struct_semun(val=system_exit.code + 1000)
+                        )
+
+                    except Exception as e:
+                        traceback.print_exc(file=sys.stdout)
+                        print(
+                            f"kart helper: unhandled exception [{e}]"
+                        )  # TODO - should ext-run capture/handle this?
```
olsen232, line 434: Do you need both of these ways of printing the exception?
Should it be to stdout or to stderr?

olsen232 (approved):
To be honest a lot of it is beyond me - can't say if you have bugs in your C structs or your sockets - but you're running the e2e test and that's working, so I'd say overall it's good

kart/cli.py
pfw: Sure thing.

kart/cli.py
pfw: The thinking there is to ensure that the environment that the helper is run in after it forks matches the environment for the 'required' values that it had when the original instance was run - library locations etc. I'd be happy to just pass everything from the client but we need combine that with those required libraries which are set in __init__.py and __init__.py isn't run in the forked helper.

kart/cli.py
pfw: Correct.

docs/pages/development/helper.rst
pfw: I thought I added one for that, will have a look.

kart/cli.py
pfw: Yep.

kart/cli.py
pfw: Will drop the second here. This is more something that I was wanting some feedback on with the ext-run command, whether it should be doing something with the command and capture an error as a different result code. I'll create an issue around that.

olsen232:
I just read the bit about pyinstaller and went to check out the pyinstaller fork. From a code-organizational standpoint it's not the best I feel - that the sender and the receiver of these messages are in two different projects, which are otherwise not related at all. If we had two projects "koordinates/kart-helper-server" and "koordinates/kart-helper-client" that would be one thing, but instead the client code ended up buried inside a seemingly random part of our toolchain that we don't otherwise maintain.

In general I would prefer - although I accept that it may be too much work - if pyinstaller remained a part of a toolchain, agnostic as to exactly what it is installing, and if the changes we made to pyinstaller were generally useful ones that could in theory be upstreamed (even if that never happens in practice). The kart-specific code would be moved inside `koordinates/kart`, but pyinstaller would have to have some mechanism for linking in the kart-specific code, and for running it immediately before pyi_main. For example, pyinstaller could be modified so that it has a concept called something like C_STARTUP_HOOKS  - static libraries that pyinstaller will link into the final product and will run before it runs pyi_main, and these startup hooks can exercise the option of aborting the program with an exit-code of their choice.

(This sort of thing is actually what we did in order to get spatial-filters into Git - we (Rob) invented a concept of a third-party filter extension that is linked into Git as a static library, and we implemented Kart's spatial filter as a particular instance of the same. We hoped to get it upstreamed into Git - this never happened unfortunately, but it does mean that the spatial filter code is stored in koordinates/kart and not in koordinates/git - but again, maybe it's too much work to be worth it.)

@rcoup any feelings about this either way? Did we realise any benefits by injecting our custom code into our generic dependency, rather than editing it directly?

pfw:
> I just read the bit about pyinstaller and went to check out the pyinstaller fork...

I tend to agree that this isn't ideal, straightforward but not ideal. I think building out into a generic extension mechanism could be great but as with git and spatial filters the risk is it goes nowhere (quite likely) so I'd skip that path.

A couple of ways forward.

- don't use a git submodule, bring in the code directly as a 'vendored' and set up the build process to include the updated parts somehow. This is complicated by PyInstaller using a fairly customised build toolchain.

- keep the client side as a separate binary within the kart tree. This is relatively simple but has some complications around finding the correct helper to run and makes bigger changes to packaging, not insurmountable. This binary can be build as primary 'kart' executable along with the PyInstaller build as kart-helper. This is probably the one to go with it we don't want to have two repos as part of one final project and the code that is within PyInstaller is not specific in any way to it so can be extracted fairly easily.

pfw:
I will add that some of the complications around a separate binary came from which process got called for `kart`, I'm more comfortable now that we can make a `kart` binary that is safe to put in place given the experience running through the pyinstaller bootloader, ie. the work in the pyinstaller boot loader got things working reliably and we can extract that whilst dealing with the binary finding issue.

Thoughts?

olsen232:
I'd be okay with any of the ways described above - whatever seems easiest to you basically

In the meantime, feel free to merge

pfw:
This should be better now as there is no custom pyinstaller. There are now two executables but that is unavoidable if we don't customise pyinstaller bootloaders.

If code is calling `kart` via whatever it finds on the PATH it will work without change - where code goes looking for a particular executable (eg. QGIS Plugin) it will need to add `kart_cli_helper` to its search order so that it picks up the correct executable to enable helper mode.

craigds (approved):

craigds:
I've been running this in a few environments with the env var set, and haven't come across any issues 👍

pfw:
The only thing that I'd be cautious off in a merge would be when it runs without KART_USE_HELPER on... if that is the case it should operate as usual, kart_cli_helper should 'just' exec kart_cli with the original args. If everyone is comfortable with that mode (and overall code) I think it's worth merging.

olsen232 (approved):

PR #663, opened 2022-07-05
Don't allow LAS files to be committed.

olsen232:
- Adds an InvalidNewValue, which works just like a ListOfConflicts.
- Adds laz-autzen to test data - like las-autzen, but compressed.
- Makes the RewriteMetadata flags more straightforward - each flag now
  simply describes what it does. Comments explaining why we use each
  flag in a certain situation are now moved to the calling location.


kart/point_cloud/schema_util.py
```diff
@@ -23,11 +23,18 @@
 ]
 
 PDRF_TO_SCHEMA = {
+    # Currently only hard-coding schemas for PDRFs 6, 7, and 8 into Kart - haven't needed the others.
```
craigds, line 26: Can you add a comment which links to some docs on what the PDRF schemas are?

craigds (approved):
LGTM

PR #665, opened 2022-07-06
Add check for `.bash_profile` for MacOS

amal-thundiyil:
## Description

`.bash_profile` seems to ship as a default in some MacOS systems. If the user hasn't sourced the `.bashrc` file in the `.bash_profle` the tab completion scripts my not be activated in the current session. To remedy that we can append to `~/.bash_profile` on Darwin systems instead.

## Related links:
- #643 

## Checklist:

### DIFF
--- a/kart/completion.py
+++ b/kart/completion.py
@@ -1,6 +1,7 @@
 import os
 import re
 import sys
+import platform
 import subprocess
 from enum import Enum
 from pathlib import Path
@@ -100,6 +101,9 @@ def install_bash(*, prog_name: str, complete_var: str, shell: str) -> Path:
     # But installing in the locations from the docs doesn't seem to have effect
     completion_path = Path.home() / f".bash_completions/{prog_name}.sh"
     bashrc_path = Path.home() / ".bashrc"
+    bash_profile = Path.home() / ".bash_profile"
+    if platform.system() == "Darwin" and bash_profile.exists():
+        bashrc_path = bash_profile
     completion_init_lines = [f"source {completion_path}"]
     return install_helper(
         prog_name,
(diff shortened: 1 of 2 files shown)


rcoup:
I don't think this is the right approach — on macOS running with bash maybe we should append to `.bash_profile` instead of `.bashrc`?

We _definitely_ shouldn't be sourcing `.bashrc` into `.bash_profile`: that's overreaching beyond our "configure tab completion" mandate.

kart/completion.py
```diff
@@ -100,6 +101,8 @@ def install_bash(*, prog_name: str, complete_var: str, shell: str) -> Path:
     # But installing in the locations from the docs doesn't seem to have effect
     completion_path = Path.home() / f".bash_completions/{prog_name}.sh"
     bashrc_path = Path.home() / ".bashrc"
+    if platform.system() == "Darwin":
```
olsen232, line 104: Maybe we should be checking what exists - if user has one and not the other, probably it's good to append it to the one they have?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.4 (UNRELEASED)
 
+- Bugfix: Add check for `.bash_profile` for MacOS. [#643](https://github.com/koordinates/kart/issues/643)
 - Support tab completion in powershell for kart. [#643](https://github.com/koordinates/kart/issues/643)
```
craigds, line 10: I'd actually just remove these two lines; the whole tab completion feature is new in this release so we don't need to list all the bugfixes we've applied to it

craigds (approved):

Issue #658, opened 2022-07-04
man-style documentation for kart


amal-thundiyil:
Kart currently doesn't have a means of building or exposing man-style documentation (eg: via `kart x --help` or `man kart x`). This should be cross-platform as much as possible, supporting bash/zsh/fish/etc as well as PowerShell on Windows. 

A brief description of the command could be sourced from docstrings and click's help text in order to make it easier to maintain. But the larger detailed description (like for `kart checkout`) should include a variation of that with added details and should follow a "simple" format like reStructuredText or Markdown for editing.

PR #682, opened 2022-07-24
Add man-style help page renderers for Kart

amal-thundiyil:
## Description

Add man-style pagers instead of click's default help page when calling `--help`. Run `kart init --help` or `kart clone --help`  to see man pages with additional docs sourced from `docs/pages/commands`. The paging methods used are:

- For Windows - `click.echo_via_pager()`
- For POSIX -  `man <path_to_page>`

While experimenting with `click-man`, I also tried using `sphinx-click` for generating man pages but couldn't get much out of it (see [stackoverflow question](https://stackoverflow.com/questions/73104016/is-there-a-programmatic-way-to-get-man-pages-from-sphinx-builder-as-strings)). 

### Design

Blank diagram (3)

## Related links:
- #658 

## Checklist:


kart/help.py
```diff
@@ -0,0 +1,115 @@
...
+    def _send_output_to_pager(self, output):
+        cmdline = self.get_pager_cmdline()
+        L.debug("Running command: %s", cmdline)
+        p = self._popen(cmdline, stdin=PIPE)
+        p.communicate(input=output)
+
+    def _popen(self, *args, **kwargs):
+        return subprocess.Popen(*args, **kwargs)
+
+    def _convert_doc_content(self, ctx, contents):
```
olsen232, line 73: Maybe add a bit of documentation to the methods that are here so that subclasses can override them

kart/docs.py
```diff
@@ -0,0 +1,183 @@
...
+        if self.description:
+            lines.append("{0} DESCRIPTION".format(self.SECTION_HEADING_KEYWORD))
+            lines.append(self.replace_blank_lines(self.description))
+
+        # write the options
+        if self.options:
+            lines.append("{0} OPTIONS".format(self.SECTION_HEADING_KEYWORD))
+            for option, description in self.options:
+                lines.append(self.INDENT_KEYWORDD)
+                option_unpacked = option.replace("-", r"\-").split()
```
olsen232, line 150: Maybe something like this for readability:
```
parts = option.replace("-", r"\-").split(maxsplit=1)
first_part, rest = parts if len(parts) == 2 else (parts[0], "")
lines.append(rf"\fB{name}\fP {desc}")
```
Feel free to give these variables a better name than "parts" or "unpacked" - I'm not sure exactly what they are so I don't know why the first one should be bold

docs/pages/commands/init.rst
```diff
@@ -0,0 +1,29 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``kart init --import <source> [<repository>]``
+
+This command creates a new repository and imports all tables from the
+given database. (For more fine grained control, use ``kart init`` to
+create an empty repository, and then use ``kart import``.)
+
+-  ``<source>``: Path to the
+   `GeoPackage <https:/google.com>`_, or `PostgreSQL Connection URI <https:/google.com>`_
```
olsen232, line 19: What's going on with the <https:/google.com> here?

kart/docs.py
```diff
@@ -0,0 +1,183 @@
+"""
+This module provides functionality to
+write a man page from some given information
+about a CLI application and is inspired from
+click-contrib/click-man.
+
+:copyright: (c) 2016 by Timo Furrer.
```
olsen232, line 7: Maybe add your name to the copyright? I don't actually know what to do here
@rcoup probably knows about licenses

kart/cli_util.py
```diff
@@ -12,6 +13,22 @@
...
 
+from kart.help import get_renderer
+
+
+class KartCommand(click.Command):
+    def format_help(self, ctx, formatter):
+        file = Path(os.path.dirname(sys.prefix)) / os.path.join(
+            "docs", "pages", "commands", f"{ctx.command.name}.rst"
+        )
+        with file.open("a") as f:
```
olsen232, line 24: Not sure what this is for? Looks like you are appending the empty string to a file

olsen232: Also if you are using pathlib.Path, you can mostly use `write_text` etc instead of open() - a bit simpler to use

olsen232 (commented):
Found lots of little things to comment on. Keen to see it in action though!

kart/cli_util.py
amal-thundiyil: AFAIK, `write_text` overwrites the file contents. Since we're reading from the docs here to produce the man pages, appending would be better.

kart/help.py
```diff
@@ -0,0 +1,115 @@
...
+        if not self._exists_on_path(cmdline[0]):
+            L.debug("Pager '%s' not found in PATH, printing raw help." % cmdline[0])
+            self.output_stream.write(output.decode("utf-8") + "\n")
+            self.output_stream.flush()
+            return
+        L.debug("Running command: %s", cmdline)
+        p = self._popen(cmdline, stdin=PIPE)
+        p.communicate(input=output)
+
+    def _exists_on_path(self, name):
```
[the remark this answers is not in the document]
amal-thundiyil: `shutil.which("groff")` seems to check the PATH of the Python runtime which may be different from the shell PATH.

kart/docs.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __str__(self):
+        """
+        Generate and return the string representation
+        of this man page.
+        """
+        lines = []
+
+        # write title and footer
+        lines.append(
```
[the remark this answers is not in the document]
amal-thundiyil: > Also you have "{1}" "1" in your format string. Is that a bug?

I don't think it's a bug. Man pages are divided into [sections](https://en.wikipedia.org/wiki/Man_page#Manual_sections). This is to show that the command is an executable program or shell command.

kart/help.py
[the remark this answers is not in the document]
olsen232: Have you seen this give different results to checking os.environ.get("PATH")? I didn't know that there could be more than one PATH variable in the environment at once. Since you're running "groff" directly - not via any shell - I would expect shutil.which to do a good job of finding the groff that will be run

kart/docs.py
craigds: if this file is copied from an online source, include the URL in the docstring here (ideally a github blob url with an OID in it so it's clear what exact version it was copied from). No need to adjust the copyright unless you've modified it substantially, in which case perhaps you need to say so here.

Either way the license for this file remains MIT not GPL. The MIT license actually requires the license text be included in distributed works, so at minimum it should be copied into this docstring I think.

kart/help.py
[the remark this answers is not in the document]
amal-thundiyil: Ah, checked again with a fresh build, works perfectly. Don't know what went wrong before.

kart/cli_util.py
amal-thundiyil: Updated.

rcoup:
@amal-thundiyil can you sketch out (a drawing) of the process(es) happening here? ie: what formats are used, and the tools/conversions that happen (during builds, rendering, etc).

kart/docs.py
craigds: @rcoup can you double-check this? (license text has now been added to the docstring)

kart/docs.py
```diff
@@ -0,0 +1,289 @@
+"""
+This module provides functionality to
+write a man page from some given information
+about a CLI application and is inspired from
+click-contrib/click-man.
+
+https://github.com/click-contrib/click-man/blob/3a0684c9d70f94a2abef08fb3f88449dab6aea98/LICENSE
```
rcoup, line 7:
```suggestion
This module provides functionality to write a man page from some given
information about a CLI application and is modified from click-man.

https://github.com/click-contrib/click-man/tree/3a0684c9d70f94a2abef08fb3f88449dab6aea98

The original click-man source is licensed under the MIT License. Changes are
licensed under the Kart license (GPL).
```

(repeated for any other files that use non-trivial click-man code)

scripts/doc_gen.py
```diff
@@ -0,0 +1,375 @@
...
+from docutils.core import publish_string
+from docutils.writers import manpage
+
+try:
+    from importlib.metadata import distribution
+except ImportError:
+    try:
+        from importlib_metadata import distribution
+    except ImportError:
+        from pkg_resources import load_entry_point
```
craigds, line 30: This should never happen since we're depending on `importlib-metadata` in requirements.txt. Maybe move that into requirements.in to make it explicit? With a comment noting it's only needed for python 3.7

scripts/doc_gen.py
```diff
@@ -0,0 +1,375 @@
...
+    input_dir = sys.argv[1]
+    output_dir = Path(sys.argv[2])
+    output_dir.mkdir(parents=True, exist_ok=True)
+    entry_point = load_entry_point("kart", "console_scripts", "kart")
+    ep_module = import_module(entry_point.value.split(":")[0])
+    ep_module._load_all_commands()
+    ep_members = getmembers(ep_module, lambda x: isinstance(x, click.Command))
+    _, cli = ep_members[0]
+
+    generate_help_pages(name, input_dir, cli, target_dir=output_dir)
```
craigds, line 358: I'm wondering what the advantage is of using the entrypoints mechanism here. Why not just 

```suggestion
    from kart import cli
    cli._load_all_commands()
    generate_help_pages(name, input_dir, cli.cli, target_dir=output_dir)
```

I'd suggest removing the leading underscore from `_load_all_commands` since it's called outside the module.

craigds (changes requested):
Looking good

I think our [`help` subcommand](https://github.com/amal-thundiyil/kart/blob/060ea1196028c173f3c3f44337dff992e609b714/kart/cli_util.py#L150) is a bit broken, unless I'm building it wrong:

```
$ kart diff --help
(works)
```

```
$ kart help diff
/Users/cdestigter/code/kart/build/venv/help/kart-help.1: No such file or directory
No manual entry for /Users/cdestigter/code/kart/build/venv/help/kart-help.1
```

Nearly there though. Sorry this review was a while coming

scripts/doc_gen.py
amal-thundiyil: Certainly much better. Thanks :+1:

craigds (changes requested):
* Strange CI failure appears unrelated (it's also happening on master) 👍 
* Maybe something more required for second-level commands. e.g. `kart meta get --help` doesn't produce the manpage-style output, whereas `kart meta help get` does

Otherwise LGTM 👍

craigds (approved):

craigds:
Thanks for your persistence! 🎉

PR #683, opened 2022-07-25
Support conflict resolution using the working copy

olsen232:
Adds --with=workingcopy to `kart resolve` in which case,
Kart will take the feature or tile from the WC as the resolution
to the specified conflict.

Note that the WC is not currently not kept in sync with the
merge - it continues to track the branch HEAD labelled "ours".
Nor is there currently a way to actually see the merge-conflicts
of PC tiles. This means the following is still needed in further PRs:

- keep the WC up to date with the merge as conflicts are resolved
- require that meta conflicts are resolved before other types of conflicts
- write all versions of conflicting tiles to the working copy, remove
them as they are resolved
- clean up any merge-state that remains - files or folders containing
conflicts - when the merge state is exited using `--continue` or `--abort`

### DIFF
--- a/kart/resolve.py
+++ b/kart/resolve.py
@@ -171,11 +171,7 @@ def _load_workingcopy_resolve_for_tile(rich_conflict):
     return _load_file_resolve_for_tile(rich_conflict, matching_files[0])
 
 
-CHOICE_ALIASES = {"working-copy": "workingcopy"}
-CONTEXT_SETTINGS = dict(token_normalize_func=lambda x: CHOICE_ALIASES.get(x, x))
-
-
-@click.command(context_settings=CONTEXT_SETTINGS)
+@click.command()
 @click.pass_context
 @click.option(
     "--with",
--- a/kart/resolve.py
+++ b/kart/resolve.py
@@ -1,5 +1,7 @@
 import copy
 import json
+import os
+import shutil
 from pathlib import Path
 
 import click
@@ -27,7 +29,7 @@ def ungeojson_file(file_path, dataset):
     Given a file containing multiple geojson features belonging to dataset,
     returns the features as dicts containing gpkg geometries.
     """
-    features = json.load(Path(file_path).open())["features"]
+    features = json.load(file_path.open())["features"]
     return [ungeojson_feature(f, dataset) for f in features]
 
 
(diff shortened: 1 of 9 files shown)


kart/resolve.py
```diff
@@ -42,41 +44,148 @@ def write_feature_to_dataset_entry(feature, dataset, repo):
...
+            click.echo(
+                os.path.relpath(file.resolve(), repo.workdir_path.resolve()), err=True
+            )
+        raise InvalidOperation("Couldn't resolve conflict using working copy")
+    return _load_file_resolve_for_tile(rich_conflict, matching_files[0])
 
-@click.command()
+
+CHOICE_ALIASES = {"working-copy": "workingcopy"}
+CONTEXT_SETTINGS = dict(token_normalize_func=lambda x: CHOICE_ALIASES.get(x, x))
```
craigds, line 175: 🙈 
does this thing normalise tokens before they're parsed? doesn't that mean that it could 'normalise' tokens that belong to a different option altogether? There's got to be a better way

kart/tabular/working_copy/base.py
```diff
@@ -733,6 +742,21 @@ def get_dirty_pks(self, dataset):
             )
             return (row[0] for row in r)
 
+    def get_feature(self, dataset, pk, allow_schema_diff=True):
+        """Gets the features for the given dataset."""
```
craigds, line 746:
```suggestion
        """
        Returns the feature that has the given primary key value.
        If the primary key value does not exist in the dataset, returns None.
        """
```

kart/tabular/working_copy/base.py
```diff
@@ -733,6 +742,21 @@ def get_dirty_pks(self, dataset):
...
+        meta_diff = self.diff_dataset_to_working_copy_meta(dataset)
+        if not allow_schema_diff and meta_diff.recursive_get(["meta", "schema.json"]):
```
craigds, line 748: `if not x and y` might _technically_ have a well-defined meaning but it's hard to mentally parse. Add some parentheses

craigds (changes requested):

kart/tabular/working_copy/base.py
olsen232: Done

kart/resolve.py
olsen232: Removed it for now. Not sure what the better way is

kart/tabular/working_copy/base.py
olsen232: Done

craigds (approved):

PR #684, opened 2022-07-25
fix: install requirements from `docs.txt` on build

amal-thundiyil:
## Description

Added installation for dependencies from `requirements/docs.txt` as part of the build process.


rcoup (approved):
LGTM, thanks

PR #685, opened 2022-07-26
ci: change "Date" to work with user's locale

amal-thundiyil:
## Description
The `Date` specified via `time.strftime("%c")` seems to be different in POSIX and Windows systems (see [build](https://github.com/koordinates/kart/runs/7523956891#step:14:1167)). 

Previous diff on Windows:
```diff
-  'Date:   Thu Jun 20 15:28:33 2019 +0100',
+  'Date:   6/20/2019 3:28:33 PM +0100',
```

### DIFF
--- a/kart/text_diff_writer.py
+++ b/kart/text_diff_writer.py
@@ -49,7 +49,8 @@ class TextDiffWriter(BaseDiffWriter):
         click.secho(f"commit {self.commit.hex}", fg="yellow", **self.pecho)
         click.secho(f"Author: {author.name} <{author.email}>", **self.pecho)
         click.secho(
-            f'Date:   {author_time_in_author_timezone.strftime("%c %z")}', **self.pecho
+            f'Date:   {author_time_in_author_timezone.strftime("%c %z")}',
+            **self.pecho,
         )
         click.secho(**self.pecho)
         for line in self.commit.message.splitlines():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,7 @@ import tarfile
 import time
 import uuid
 from urllib.parse import urlsplit, urlunsplit
+from datetime import datetime
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


craigds:
not sure about this. `%c` varies depending on the user's locale. If they're different on windows/linux by default that just means that those platforms have adopted different locale settings.

I believe git does also uses `%c`, so perhaps it also has different output on windows vs linux (?)

I would suggest modifying the failed test so that it doesn't depend on the caller's locale (perhaps by just not asserting anything about the timestamp line), rather than switching to a hardcoded formatstring here which may not be appropriate for all locales.

@olsen232 thoughts?

olsen232:
> not sure about this. `%c` varies depending on the user's locale. If they're different on windows/linux by default that just means that those platforms have adopted different locale settings.
> 
> I believe git does also uses `%c`, so perhaps it also has different output on windows vs linux (?)
> 
> I would suggest modifying the failed test so that it doesn't depend on the caller's locale (perhaps by just not asserting anything about the timestamp line), rather than switching to a hardcoded formatstring here which may not be appropriate for all locales.
> 
> @olsen232 thoughts?

Yeah this logic makes sense. I would say any date formatting we do should be one of the following - 
not locale dependent at all (eg ISO8601: `2022-07-27 14:14:01`)
or, using the user's locale properly eg strftime("%c")
or, following some other logical standard (eg we could find out what Git does and copy it exactly, but I would really only want to emulate Git if it's roughly following one of the other two options)

craigds:
> not locale dependent at all (eg ISO8601

we do this in [`kart show -o json`](https://github.com/koordinates/kart/blob/9db5856182ea04f63fa674b9d898d7057b23c7f6/kart/log.py#L473)

which I think makes sense; json output is meant to be machine readable and shouldn't be locale-dependent

craigds (approved):

PR #691, opened 2022-08-10
Don't ship duplicate libraries in packages

pfw:
## Description

Pyinstaller will copy any file (shared library) in the source virtual env rather than honouring symlinks - the end result is that versioned symlinks for shared libraries get copied.

Additionally strip shared libraries on Linux to reduce their size.

## Checklist:


craigds (approved):
LGTM (I've tested  the smaller packages briefly, no issues found). Nice job

PR #694, opened 2022-08-10
tests: fix Date for commands passed to git

amal-thundiyil:
## Description

`git log` has a different way of parsing `%c` in dates. This PR gets dates from `git show`  and asserts the new date in tests for `kart log` and `kart upgrade`.  

## Related links:

- #685

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -714,6 +714,12 @@ class TestHelpers:
         subprocess.check_call(cmd + list(paths))
 
     @classmethod
+    def get_git_datetime(cls, sha: str) -> str:
+        cmd = ["git", "show", "-s", "--format=%cd", sha]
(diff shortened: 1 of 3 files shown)


olsen232 (commented):
Here's an idea that might make this a little nicer to read or perhaps a little more self explanatory:

rather than having to break the assertion into lots of smaller assertions, just insert the expected time directly into the expected-output. It could look something like this:
```
expected_output = [
                f"commit {H.POINTS.HEAD_SHA}",
                "Author: Robert Coup <robert@coup.net.nz>",
                f"Date:   {localised_datetime(H.POINTS.DATE_TIME)}",
```
or even like this:
```
    f"Date:   {H.POINTS.LOCALISED_DATETIME()}",
 ```

olsen232 (approved):

PR #695, opened 2022-08-11
Re-implement git-lfs pre-push in Kart.

olsen232:
It turned out not to be very complicated, since we can still delegate
to git-lfs to do the push itself.
The re-implementation only scans for PC tiles, so, there's no chance
of it encountering missing+promised spatial filtered features - so this
fixes the issue where X_KART_POINT_CLOUDS breaks spatial filtered repos.
(However, it means if we add more dataset types that use LFS, we'll need
to modify what is scanned).

Still TODO in further PRs:
Better chunking and progress while pushing.
Re-implement git-lfs fetch in Kart too, including spatial filtering.

### DIFF
--- a/kart/lfs_commands/__init__.py
+++ b/kart/lfs_commands/__init__.py
@@ -1,0 +1,86 @@
+import os
+import subprocess
+import sys
+
+import click
+
+from kart.cli_util import add_help_subcommand
+from kart.lfs_util import get_hash_from_pointer_file
+from kart.rev_list_objects import rev_list_tile_pointer_files
+from kart.repo import KartRepoState
+
+EMPTY_SHA = "0" * 40
+
+
+@add_help_subcommand
+@click.group("lfs+", hidden=True)
+@click.pass_context
+def lfs_plus(ctx, **kwargs):
+    """Git-LFS commands re-implemented in Kart to allow for spatial filtering."""
+
+
+@lfs_plus.command()
+@click.pass_context
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    help="Don't push anything, just show what would be pushed",
+)
+@click.argument("remote_name", required=False)
+@click.argument("remote_url", required=False)
+def pre_push(ctx, remote_name, remote_url, dry_run):
+    """
+    Re-implementation of git-lfs pre-push - but, only searches for pointer blobs at **/.point-cloud-dataset.v?/tile/**
+    This means it won't encounter any features that are missing due to spatial filtering, which git-lfs stumbles over.
+    """
+
+    if not remote_url and not remote_url:
+        raise RuntimeError(
+            "kart lfs+ pre-push should be run through Kart's pre-push hook."
+        )
+
+    if os.environ.get("GIT_LFS_SKIP_PUSH", False):
+        return
+
+    dry_run = dry_run or os.environ.get("GIT_LFS_DRY_RUN_PUSH", False)
+
+    repo = ctx.obj.get_repo(allowed_states=KartRepoState.ALL_STATES)
+
+    start_commits, stop_commits = get_start_and_stop_commits(sys.stdin)
+
+    lfs_oids = set()
+    for (commit_id, path_match_result, pointer_blob) in rev_list_tile_pointer_files(
+        repo, start_commits, stop_commits
+    ):
+        lfs_oids.add(get_hash_from_pointer_file(pointer_blob))
+
+    if dry_run:
+        click.echo(
+            f"Running pre-push with --dry-run: pushing {len(lfs_oids)} LFS blobs"
+        )
+        for lfs_oid in lfs_oids:
+            click.echo(lfs_oid)
+        return
+
+    if lfs_oids:
+        # TODO - chunk lfs_oids so that we don't overflow the maximum argument size.
+        # TODO - capture chunk progress and report our own overall progress
+        subprocess.check_call(
+            ["git-lfs", "push", remote_name, "--object-id", *lfs_oids]
+        )
+
+
+def get_start_and_stop_commits(input_iter):
+    start_commits = set()
+    stop_commits = set()
+    for line in input_iter:
+        if not line:
+            continue
+        local_ref, local_sha, remote_ref, remote_sha = line.split()
+        start_commits.add(local_sha)
+        stop_commits.add(remote_sha)
+
+    start_commits.discard(EMPTY_SHA)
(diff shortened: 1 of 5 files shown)


kart/lfs_commands/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+)
+@click.argument("remote_name", required=False)
+@click.argument("remote_url", required=False)
+def pre_push(ctx, remote_name, remote_url, dry_run):
+    """
+    Re-implementation of git-lfs pre-push - but, only searches for pointer blobs at **/.point-cloud-dataset.v?/tile/**
+    This means it won't encounter any features that are missing due to spatial filtering, which git-lfs stumbles over.
+    """
+
+    if not remote_url and not remote_url:
```
craigds, line 37: dupe expression

kart/lfs_commands/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+            f"Running pre-push with --dry-run: pushing {len(lfs_oids)} LFS blobs"
+        )
+        for lfs_oid in lfs_oids:
+            click.echo(lfs_oid)
+        return
+
+    if lfs_oids:
+        # TODO - chunk lfs_oids so that we don't overflow the maximum argument size.
+        # TODO - capture chunk progress and report our own overall progress
+        subprocess.check_call(
```
craigds, line 68: Maybe use `kart.exec.run_and_wait` for this, it tries to mimic `os.execvp` and adds the tool environment stuff in

kart/lfs_util.py
```diff
@@ -21,10 +21,15 @@
 _EMPTY_SHA256 = "sha256:" + ("0" * 64)
 
 
+PRE_PUSH_HOOK = "\n".join(["#!/bin/sh", 'kart lfs+ pre-push "$@"', ""])
```
craigds, line 24:
```suggestion
PRE_PUSH_HOOK = "\n".join(["#!/bin/sh", 'exec kart lfs+ pre-push "$@"', ""])
```

the shell doesn't need to fork before running kart, it can just exec kart directly. Not sure if/how this runs on windows though

kart/lfs_commands/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+from kart.repo import KartRepoState
+
+EMPTY_SHA = "0" * 40
+
+
+@add_help_subcommand
+@click.group("lfs+", hidden=True)
+@click.pass_context
+def lfs_plus(ctx, **kwargs):
+    """Git-LFS commands re-implemented in Kart to allow for spatial filtering."""
```
craigds, line 19: can/should we fallthru other `git lfs` commands to invoke `git lfs <command>`?

craigds (approved):

kart/lfs_commands/__init__.py
olsen232: Fixed

kart/lfs_commands/__init__.py
olsen232: Not done - calling subprocess directly works much the same but gives a bit more control

kart/lfs_util.py
olsen232: Not done - I checked what the LFS hooks looked like on windows and macos and in both cases they looked like this... not really too concerned if the shell forks or not, but aiming to have a hook is at least as cross-platform as the original

kart/lfs_commands/__init__.py
olsen232: We could... if we wanted to make a full suite of git-lfs commands, where some are overridden and some still have the default behaviour. However, it's a bit out of scope for right now
- the original commands are still there if you want them - just run git-lfs
- most of them will just be for diagnostic purposes - there's only one or two that we actually need to work in Kart, ie, fetching and pushing - so it's okay if they're a little bit hidden away
- I haven't looked through them to see which ones we expect to work perfectly, and which ones will be a little weird or completely broken in Kart repos due to us disabling the git-worktree in a slightly hacky way. So, until and unless I do this, not much point exposing them again from within Kart, it would just give the expectation that they can should work properly

kart/lfs_commands/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    if dry_run:
+        click.echo(
+            f"Running pre-push with --dry-run: pushing {len(lfs_oids)} LFS blobs"
+        )
+        for lfs_oid in lfs_oids:
+            click.echo(lfs_oid)
+        return
+
+    if lfs_oids:
+        # TODO - chunk lfs_oids so that we don't overflow the maximum argument size.
```
rcoup: LINZ is intending to use 5-figure tile counts pretty early, and on Windows command line arg length is somewhere between 8KB & 32KB...

Have submitted https://github.com/git-lfs/git-lfs/pull/5086 upstream to allow `git lfs push <remote> --object-id --stdin`

kart/rev_list_objects.py
```diff
@@ -96,6 +96,21 @@ def rev_list_feature_blobs(repo, start_commits, stop_commits):
     )
 
 
+TILE_POINTER_FILES_PATTERN = re.compile(r"(.+)/\.point-cloud-dataset[^/]*/tile/.+")
+
+
+def rev_list_tile_pointer_files(repo, start_commits, stop_commits):
+    """
+    Yield all the blobs with a path identifying them as features (or rows) of a "table-dataset".
```
rcoup, line 104: > Yield all the blobs with a path identifying them as features (or rows) of a **"table-dataset"**.

point cloud dataset?

rcoup: Worth adding a statement along these lines?

> Unlike git-lfs, we don't scan blob bodies to identify LFS pointer files, since we know exactly what paths will be LFS objects for a dataset.

rcoup (commented):
👍

kart/lfs_commands/__init__.py
olsen232: Nice one - godspeed to your PR

kart/rev_list_objects.py
olsen232: This is not necessarily relevant to this python file, which concerns itself only with running rev-list --objects commands, and, using them to list all the whatsits in a thingamy-dataset.

I've added a bit more documentation to the kart lfs+ push command to explain this

kart/lfs_commands/__init__.py
rcoup: that PR has been merged upstream now 🥳

Issue #701, opened 2022-08-23
config: tab completion setup issues


rcoup:
**Describe the bug**

@amal-thundiyil 

A couple of minor annoyances setting up tab completion:

```console
$ kart config
usage: git config [<options>]
...
```

Expected to see the same as `kart config --help`, or some combination of the two — maybe the key arguments from `git config`, and an explanation that anything you can pass to `git config` will also work for `kart config`?

```console
kart config --help
Usage: kart_cli config [OPTIONS] [ARGS]...

  Get and set repository or global options

Options:
  --install-tab-completion [bash|zsh|fish|powershell|pwsh|auto]
                                  Install tab completion for the specific or
                                  current shell
  --help                          Show this message and exit.
```

Then:
```console
$ kart config --install-tab-completion
Error: Option '--install-tab-completion' requires an argument.
```
I think the default should be `auto`

**Version Info **
 - OS: macOS
 - Version: v0.11.6.dev0

craigds:
fixed by #703

PR #703, opened 2022-08-23
fix(config): move `--install-tab-completion` to its own command

amal-thundiyil:
## Description

The problem with the earlier approach is that the `--install-tab-completion` which should be an option is also trying to play the role of a command (also see [koordinates/kart#648 (review)](https://github.com/koordinates/kart/pull/648#discussion_r897457027)). This PR adds a new command instead of adding it as an option in a subcommand. 
```sh
kart install tab-completion # to auto-detect shell and install tab completion
```

## Related links:

- #701 
- #643 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## 0.11.6 (UNRELEASED)
 
+- Install tab completion using `kart install tab-completion` instead of `kart config --install-tab-completion`. [#701](https://github.com/koordinates/kart/issues/701)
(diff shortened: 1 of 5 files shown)


rcoup:
I feel like we might end up with other setup commands which are only useful once.

(I think it was my suggestion to move it under config, sorry)

Thoughts @craigds @olsen232 ?

olsen232:
Hard to argue with the premise - it is a command, not an option, so if we make any distinction, that's where it should go.

It could be a subcommand if we don't want it at the top-level - if this is too much prominence for a one-off command - or if we expect to have more one-off commands like this.

It could maybe be `kart config install-tab-completion` (not sure if we can subcommand `kart config`, or if it has to be a special "group" command) or it could be `kart install tab-completion`

I think probably - perhaps more importantly than exactly what we call our commands - we need to one day customise the `kart --help` text so that it can give prominence to some commands ahead of others, maybe add some headings eg "common commands", "advanced", "setup commands"

craigds:
LGTM :shipit:

PR #704, opened 2022-08-24
Support shell detection in helper mode

pfw:
## Description

Shell detection in for command completion installation goes through a few hoops to determine the correct shell. If it can't, in this case due to double forking and detaching in the background of the helper if will fail.

Additionally shellingham.posix was not being picked up by the pyinstaller build.

Dump traceback for unhandled errors when running under helper

## Related links:
https://github.com/sarugaku/shellingham#notes-for-application-developers

## Checklist:

### DIFF
--- a/kart.spec
+++ b/kart.spec
@@ -115,6 +115,7 @@ pyi_analysis = Analysis(
         '_cffi_backend',
         # via a cython module ???
         'csv',
+        'shellingham.posix',
         *collect_submodules('kart'),
         *collect_submodules('sqlalchemy'),
     ],
--- a/kart.spec
+++ b/kart.spec
@@ -116,6 +116,7 @@ pyi_analysis = Analysis(
         # via a cython module ???
         'csv',
         'shellingham.posix',
+        'shellingham.nt',
         *collect_submodules('kart'),
         *collect_submodules('sqlalchemy'),
     ],
--- a/kart/completion.py
+++ b/kart/completion.py
@@ -17,7 +17,7 @@ except ImportError:
 
 def provide_default_shell():
     if os.name == 'posix':
-        return os.environ['SHELL']
+        return os.environ.get('SHELL', '')
     raise NotImplementedError(f'OS {os.name!r} support not available')
 
 class Shells(str, Enum):
--- a/kart/completion.py
+++ b/kart/completion.py
@@ -18,6 +18,8 @@ except ImportError:
 def provide_default_shell():
     if os.name == 'posix':
         return os.environ.get('SHELL', '')
+    elif os.name == 'nt':
+        return os.environ.get('COMSPEC', '')
     raise NotImplementedError(f'OS {os.name!r} support not available')
 
 class Shells(str, Enum):
--- a/kart/completion.py
+++ b/kart/completion.py
@@ -15,6 +15,10 @@ try:
 except ImportError:
     shellingham = None
 
+def provide_default_shell():
+    if os.name == 'posix':
+        return os.environ['SHELL']
+    raise NotImplementedError(f'OS {os.name!r} support not available')
 
 class Shells(str, Enum):
     bash = "bash"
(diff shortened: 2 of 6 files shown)


kart/completion.py
```diff
@@ -15,6 +15,10 @@
 except ImportError:
     shellingham = None
 
+def provide_default_shell():
+    if os.name == 'posix':
+        return os.environ['SHELL']
```
rcoup, line 20: from https://github.com/sarugaku/shellingham but safeguarded against KeyErrors. The caller already maps '' to a clean user error. 

```suggestion
    if os.name == 'posix':
        return os.environ.get('SHELL', '')
    elif os.name == 'nt':
        return os.environ.get('COMSPEC', '')
```

kart/helper.py
```diff
@@ -203,9 +204,11 @@ class struct_semun(ctypes.Union):
                         )
 
                     except Exception as e:
+                        print(os.environ)
```
rcoup, line 207: Secrets could be in the environment, we shouldn't be printing values at all — maybe just keys?

Ideally we'd gate this env info, and the full traceback, behind a verbosity flag.

(And `print()` doesn't go to stderr by default)

kart/helper.py
```diff
@@ -203,9 +204,11 @@ class struct_semun(ctypes.Union):
...
                         print(
-                            f"kart helper: unhandled exception [{e}]"
+                            f"kart helper: unhandled exception"
                         )  # TODO - should ext-run capture/handle this?
+                        traceback.print_exc(file=sys.stdout)
```
rcoup, line 211: should go to stderr

kart.spec
```diff
@@ -115,6 +115,7 @@ pyi_analysis = Analysis(
         '_cffi_backend',
         # via a cython module ???
         'csv',
+        'shellingham.posix',
```
rcoup, line 118: is there a `shellingham.nt` module we should be adding on windows? 
On windows we may be running in git-bash/mingw-sh, so I guess it needs both?

rcoup (changes requested):

kart/helper.py
pfw: Yep, that shouldn't be there, debug.

kart.spec
pfw: Not sure on that one - does pyinstaller include things in the module but not submodules? e.g. shellingham.nt is not a submodule. I'm not clear on that.

kart/completion.py
pfw: Sounds good, I had left of the nt version since we don't currently support helper mode there and so shouldn't have to fall back to a default way of finding the shell.

pfw: Not sure how we'd get to test that.

pfw:
Updating for merge of tab completion install being its own command.

kart.spec
rcoup: i was just guessing, since pyinstaller's autodetection code clearly didn't find the posix module

rcoup: the auto-detection finds import statements, but doesn't deal well with clever conditional imports inside functions.

kart/completion.py
rcoup: > I had left of the nt version since we don't currently support helper mode there and so shouldn't have to fall back to a default way of finding the shell.

Ah, ok. TBC, tab completion _does_ work on Windows

kart.spec
pfw: I've added 'shellingham.nt' as it fails when I added an e2e powershell test.

```
ModuleNotFoundError: No module named 'shellingham.nt'
```

kart/completion.py
pfw: As below I've added the test in the powershell e2e test and that picked up the need for an additional hidden import though it sounds like the env var will always be present on windows so that COMSPEC check might be safe.

rcoup: so, we should add the fallback to COMSPEC? is gone in the latest patch

pfw: Yep. Will do this evening.

Issue #706, opened 2022-08-24
diff, show (and others) need improved understanding of commits vs paths


olsen232:
Git can understand all of the following forms:
`git diff PATH PATH`
`git diff COMMIT PATH PATH`
`git diff COMMIT COMMIT PATH PATH`
`git diff COMMIT...COMMIT PATH PATH`
`git diff -- PATH PATH`
`git diff COMMIT -- PATH PATH`
`git diff COMMIT COMMIT -- PATH PATH`
`git diff COMMIT...COMMIT -- PATH PATH`

log, show, and commit have similar (but not identical) patterns.

Since most words eg "foo" could potentially be a reference to a commit, or a path, many of these forms are ambiguous - does the user mean `git diff COMMIT COMMIT` or `git diff PATH PATH`? 
Git can tell them apart using the `--` marker if it is present, or failing that, checking to see if "foo" is a reference to a commit. If "foo" is both a reference to a commit and a path, it warns the user.

Kart does not have this logic, for the most part. `kart log` is the exception - it works like Git - and `kart diff COMMIT PATH PATH` at least checks if the first PATH should actually be promoted to COMMIT - but in general, Kart commands don't understand the `--` at all, and in general Kart commands assume that the first argument, if present, is a commit, and all subsequent commands, if present, are paths.

The logic for distinguishing commits and paths should be extracted from `kart log` and generalised for use for all of these commands

olsen232:
I reused the `kart log` logic for show and diff as described above, and it works great as long as you only do sensible things. However, I realised much later that it falls apart at the edges, mostly by doing unhelpful operations rather than showing helpful error messages.

The major issue is that we don't check which paths exist when parsing filter arguments. Git does this and so the equivalent Git commands don't do things unless they are sure what you mean: `git log A B C D` will check if A, B, C and D are commits to be logged, or paths to filter the log by. If they are both / neither, it will error and ask you to be more specific.

Since Kart doesn't do this, if you say `kart log A B C D` it will assume any of A B C D that it doesn't recognize as a ref / commit, must be a filter. If it was supposed to be a commit but was mistyped, then it will become a useless filter, filtering the results down to nothing.

This bug is worse lately due to spreading to show and diff, where this behaviour is even more surprising - eg `kart show A` where A is a mistyped commit will instead show HEAD but filtered down to nothing. 

We have a lot of options for how to fix this:
[...]

PR #707, opened 2022-08-25
Use commit+filter parsing from log in diff, show

olsen232:
kart log understands the `--` separates commits from filters.
kart diff and kart show don't undestand that, and also they don't allow the first argument be a filter.

This moves the log argument parsing into a parse_args.py and reuses it for kart diff and kart show. 

## Related links:

https://github.com/koordinates/kart/issues/706

## Checklist:

### DIFF
--- a/kart/parse_args.py
+++ b/kart/parse_args.py
@@ -148,16 +148,12 @@ def parse_commits_and_filters(
     if kwargs is not None and allow_options:
         for option_name, option_val in kwargs.items():
             option_name = option_name.replace("_", "-", 1)
-            mapping = {
-                "int": lambda: options.append(f"--{option_name}={option_val}"),
-                "str": lambda: options.append(f"--{option_name}={option_val}"),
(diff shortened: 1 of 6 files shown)


kart/parse_args.py
```diff
@@ -0,0 +1,163 @@
...
+            allow_filters=allow_filters,
+        )
+        lists_by_type[arg_type].append(arg)
+        if arg_type == ArgType.FILTER:
+            allow_commits = False
+
+    if kwargs is not None and allow_options:
+        for option_name, option_val in kwargs.items():
+            option_name = option_name.replace("_", "-", 1)
+            mapping = {
```
craigds, line 151: this thing seems like it should just be a if/elif/else, and ditch the lambdas

craigds (approved):

kart/parse_args.py
olsen232: Done

PR #711, opened 2022-09-09
Improve understanding of revisions vs filters [#706]

olsen232:
Don't assume things are filters just because they seem not to be revisions - they may be a revision with a typo in them. Instead, check to make sure filters are present at HEAD, and fail with an error message if we can't decide if something is a revision or a filter.

Also fixes up some naming.

Fixes https://github.com/koordinates/kart/issues/706

Changelog doesn't need updating: already refers to this issue for the next release.

### DIFF
--- a/kart/parse_args.py
+++ b/kart/parse_args.py
@@ -72,37 +72,44 @@ def _append_kwargs_to_options(options, kwargs, allow_options):
 
     for option_name, option_val in kwargs.items():
         option_name = option_name.replace("_", "-", 1)
-        t = type(option_val)
-        if t == int or t == str:
+        if isinstance(option_val, bool):
+            if option_val:
+                options.append(f"--{option_name}")
+        elif isinstance(option_val, (int, str)):
             options.append(f"--{option_name}={option_val}")
-        elif t == tuple:
-            options.extend([f"--{option_name}={o}" for o in option_val]),
-        elif t == bool and option_val:
-            options.append(f"--{option_name}")
+        elif isinstance(option_val, tuple):
+            options.extend([f"--{option_name}={o}" for o in option_val])
 
 
-HEAD_PATTERN = re.compile(r"^HEAD\b", re.IGNORECASE)
+HEAD_PATTERN = re.compile(r"^HEAD\b")
+RANGE_PATTERN = re.compile(r"[^/]\.{2,}[^/]")
 
 HINT = "Use '--' to separate paths from revisions, like this:\n'kart <command> [<revision>...] -- [<filter>...]'"
 SILENCING_HINT = "To silence this warning, use '--' to separate paths from revisions, like this:\n'kart <command> [<revision>...] -- [<filter>...]'\n"
 
 
-def _is_revision(repo, arg, warning_set):
+def _is_revision(repo, arg, dedupe_warnings):
     # These things *could* be a path, but in that case the user should add a `--` before this arg to
     # disambiguate, and they haven't done that here.
-    if arg == "[EMPTY]" or ".." in arg or HEAD_PATTERN.match(arg) or arg.endswith("^?"):
+    if (
+        arg == "[EMPTY]"
+        or arg.endswith("^?")
+        or RANGE_PATTERN.search(arg)
+        or HEAD_PATTERN.search(arg)
+    ):
         return True
 
     filter_path = arg.split(":", maxsplit=1)[0]
     head_tree = repo.head_tree
     is_useful_filter_at_head = head_tree and filter_path in head_tree
 
-    if ":" in arg and not is_useful_filter_at_head:
-        click.echo(
-            f"Assuming '{arg}' is a filter argument (that doesn't match anything at HEAD)",
-            err=True,
-        )
-        warning_set.add(SILENCING_HINT)
+    if ":" in arg:
+        if not is_useful_filter_at_head:
+            click.echo(
+                f"Assuming '{arg}' is a filter argument (that doesn't match anything at HEAD)",
+                err=True,
+            )
+            dedupe_warnings.add(SILENCING_HINT)
         return False
 
     try:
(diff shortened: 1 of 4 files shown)


kart/parse_args.py
```diff
@@ -75,29 +61,89 @@ def get_arg_type(repo, arg, allow_options=True, allow_commits=True, allow_filter
...
+    else:
+        raise NotFound(
+            f"Ambiguous argument '{arg}' - doesn't appear to be either a revision or a filter\n{HINT}"
+        )
+
+
+def _disambiguate_revisions_and_filters(repo, args):
+    revisions = []
+    filters = []
+    warning_set = set()
```
craigds, line 131: not sure this needs to be a set; it looks like you're only adding SILENCING_HINT, so there should be other ways to dedupe that. Making it a set means the output will be nondeterministically ordered if you ever add a second thing to the set.

Perhaps we should configure the [warnings](https://docs.python.org/3/library/warnings.html) module? It will de-dupe warnings automatically if configured correctly

kart/show.py
```diff
@@ -88,10 +88,12 @@ def show(
     args,
 ):
     """
-    Show the given commit, or HEAD
+    Shows the given REVISION, or HEAD if none in specified.
```
craigds, line 91:
```suggestion
    Shows the given REVISION, or HEAD if none is specified.
```

kart/parse_args.py
```diff
@@ -75,29 +61,89 @@ def get_arg_type(repo, arg, allow_options=True, allow_commits=True, allow_filter
...
+def _is_revision(repo, arg, warning_set):
+    # These things *could* be a path, but in that case the user should add a `--` before this arg to
+    # disambiguate, and they haven't done that here.
+    if arg == "[EMPTY]" or ".." in arg or HEAD_PATTERN.match(arg) or arg.endswith("^?"):
+        return True
 
-def parse_commits_and_filters(
+    filter_path = arg.split(":", maxsplit=1)[0]
+    head_tree = repo.head_tree
+    is_useful_filter_at_head = head_tree and filter_path in head_tree
```
craigds, line 98: how's `filter_path in head_tree` work? It looks like `filter_path` is something like `["<datasetname>", "<pk_value>"]` and you're passing it to a `pygit2.Tree.__contains__`, which doesn't know about our path encodings. Or am I missing something?

kart/parse_args.py
```diff
@@ -75,29 +61,89 @@ def get_arg_type(repo, arg, allow_options=True, allow_commits=True, allow_filter
...
+
+    for option_name, option_val in kwargs.items():
+        option_name = option_name.replace("_", "-", 1)
+        t = type(option_val)
+        if t == int or t == str:
+            options.append(f"--{option_name}={option_val}")
+        elif t == tuple:
+            options.extend([f"--{option_name}={o}" for o in option_val]),
+        elif t == bool and option_val:
+            options.append(f"--{option_name}")
```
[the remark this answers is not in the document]
olsen232: 👍 good call but I find it amusing how you recommend style changes to your own code

kart/parse_args.py
```diff
@@ -75,29 +61,89 @@ def get_arg_type(repo, arg, allow_options=True, allow_commits=True, allow_filter
...
-    if allow_commits:
-        if arg == "[EMPTY]" or ".." in arg:
-            return ArgType.COMMIT
 
-        try:
-            repo.resolve_refish(arg)
-            return ArgType.COMMIT
-        except (KeyError, ValueError, pygit2.InvalidSpecError):
-            pass
+HEAD_PATTERN = re.compile(r"^HEAD\b", re.IGNORECASE)
```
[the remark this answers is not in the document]
olsen232: Hmm then we're probably inconsistent depending on whether we delegate to git or pygit2 or handle it ourselves... but happy to require capital HEAD here

kart/parse_args.py
```diff
@@ -75,29 +61,89 @@ def get_arg_type(repo, arg, allow_options=True, allow_commits=True, allow_filter
...
+SILENCING_HINT = "To silence this warning, use '--' to separate paths from revisions, like this:\n'kart <command> [<revision>...] -- [<filter>...]'\n"
 
-    raise click.UsageError(
-        f"Argument not recognised as a valid commit, ref, or range: {arg}"
-    )
 
+def _is_revision(repo, arg, warning_set):
+    # These things *could* be a path, but in that case the user should add a `--` before this arg to
+    # disambiguate, and they haven't done that here.
+    if arg == "[EMPTY]" or ".." in arg or HEAD_PATTERN.match(arg) or arg.endswith("^?"):
```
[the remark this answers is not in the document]
olsen232: Done

kart/parse_args.py
olsen232: ... the warnings module looks great for deprecating things in modules that will be imported by third parties, etc etc. But I am just trying to print something once: I don't need auto line numbers + module names, or configuration options, and I especially don't want it to remember which warnings it has already printed for any longer than this function call (as could happen with our testing framework / kart helper). It's not beautiful how it is but I've just spent 10 minutes thinking about this and I think it's not really worth spending any longer on it - there's actually nothing measurably wrong with it right now

kart/show.py
olsen232: Done

kart/parse_args.py
olsen232: I only pass the first part of a filter to it - everything before the ":" - which is a path with no particular encoding. Everything after that is ignored. So if you have an arg like "points:feature:25" and a dataset called "points" and no revision called "points", we're happy to assume your arg is a filter - even without encoding the feature PK to see if it exists, it's enough that the dataset exists.

kart/parse_args.py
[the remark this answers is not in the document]
craigds: yep. not sure what i was thinking when i wrote it, that code doesn't seem like something I would write normally. Maybe I wrote it before I had a ☕

kart/parse_args.py
craigds: > as could happen with our testing framework

AFAIK pytest resets the warnings state between each test, by default. It also has really good support for the warnings module if you want to check for warnings in your tests: https://docs.pytest.org/en/latest/how-to/capture-warnings.html

> kart helper

yep, that could use some work. Looks like there's no simple way to reset the warning registry globally: https://bugs.python.org/issue21724 - which we'd need to make this work well for kart helper

So in that case, I guess leave this as it is 👍

kart/parse_args.py
craigds: good point, I missed L102

craigds (approved):

Issue #583, opened 2022-03-16
Attachment support for Kart


olsen232:
### Google Summer of Code summary
Kart enables version control of vector or tabular datasets, just as Git enables version control of files and folders. If Kart also supported version control of files and folders, then any files that are relevant to these datasets - perhaps README files, images and thumbnails, documentation PDFs, licenses, metadata XML - could be stored in the same version controlled repository alongside the datasets they refer to. Kart is already capable of storing version controlled files, using a Git object database, but every Kart operation needs to be modified so that it works simultaneously on tabular datasets and on files and folders (ie, the command to commit changes should simultaneously commit tabular changes from a database and file changes from the filesystem). And we need a user experience that doesn't lead to datasets accidentally being committed as attachments!

### GSOC stats
* Project Size: Medium (175h)
* Mentors: [@olsen232](https://github.com/olsen232)
* Skills needed: Python, Git

### What's working so far
Kart repos are actually Git repos, which of course are version-controlled folder hierarchies of files. All of these files are hidden away inside the Git Object Database, which is in `.kart/objects` but which can't be read without the proper Git tools. Embedded inside these version controlled hierarchies of files are "datasets" - folders which contain lots of files in a particular format, which represent a database table (optionally with geometry data in one of the columns). At present, Kart only exposes that dataset part of the folder hierarchy - any other files in the hierarchy are only visible if you know the right Git tools to see them. But, they are there - repositories can already contain these "attachments", some repositories do, and it is possible to add more attachments using a hidden command called `kart commit-files`.

### View existing attachments
Try running `git ls-tree -t -r HEAD | grep -v dataset | grep blob` to see if there are any attachments at HEAD in a Kart repository.

### Summary of steps to be taken
1. Detect changes to attachments between two commits during `kart diff COMMIT1 COMMIT2`
2. Modify the internal structure in which diffs are stored so it can hold diffs to attachments, not just diffs to datasets
3. Modify the output of `kart diff` so that it can show these attachment diffs in some way

4. Create a file-based working copy, separate from the database working copy that Kart so far has, so that all checkout operations continue to check out database contents to the database but now also attachments to the filesystem
(Big job. The following existing Git technologies may be useful here: index files, and sparse checkouts)
5. Detect when attachments have been changed in the user's filesystem when running commands like `kart diff` and `kart status`, and include that information to the output
6. Make sure attachment changes are committed when running `kart commit`

[...]

PR #713, opened 2022-09-16
Support patches to attached files

olsen232:
Includes some cleanup of apply.py - moving things into helper functions etc.
Still TODO - validate that the patch still applies, ie, that the files being updated haven't changed since the patch was created.

## Related links:

https://github.com/koordinates/kart/issues/583

## Checklist:


kart/structure.py
```diff
@@ -308,7 +321,6 @@ def check_values_match_schema(self, repo_diff):
                 if new_schema is None:
                     raise InvalidOperation(
                         f"Can't {feature_delta.type} feature {feature_delta.new_key} in deleted dataset {ds_path}",
-                        exit_code=PATCH_DOES_NOT_APPLY,
```
craigds, line 311: why this change?

kart/structure.py
```diff
@@ -235,6 +236,14 @@ def create_tree_from_diff(
                 object_builder.insert(path, blob)
 
         for ds_path, ds_diff in repo_diff.items():
+            if ds_path == FILES_KEY:
+                self.apply_file_diff(
```
craigds, line 240: this method should maybe be `apply_files_diff` since it's not just _one_ file, it's all the files

kart/apply.py
```diff
@@ -178,6 +186,59 @@ def _build_signature(patch_metadata, person, repo):
     return repo.author_signature(**signature)
 
 
+def parse_file_diff(file_diff_input, allow_minimal_updates=None):
+    def convert_half_delta(half_delta):
+        if half_delta is None:
+            return None
+        elif half_delta.value.startswith("base64:"):
+            return (half_delta.key, b64decode_str(half_delta.value))
+        else:
```
craigds, line 195:
```
"+": "base64:\nbase64 is a way of encoding binary data in strings that you can use in JSON, like this patch."
```

to keep the encoding prefix purely unambiguous maybe we need an enforced prefix for text content? maybe just `:` or `text:`. Ugly though. But it'd be annoying to have a list of prefixes you're not allowed in text data.

craigds (approved):

kart/structure.py
olsen232: hmm good question. I was looking at some exit-codes to make sure they made sense - eg PATCH_DOES_NOT_APPLY should be for the patch is valid, but doesn't work on HEAD, but there should be different error codes for "can't parse the patch". But this one seems pretty reasonably how it is. I'll revert

kart/apply.py
olsen232: Yeah I did this prefix thing a bit late and it's a bit ugly - again, it's because there's not really any room for anything out-of-band in the JSON diffs. JSON-LINES on the other hand has more room and it has a "binary" attribute that can disambiguate these two cases. I was hoping we'd just switch to something more like JSON-LINES for everything including patches but we're not really ready for that... this will work for now, until and unless someone starts creating patches which modify text files to have "base64:" at the start of them. It will go away if we ever manage to drop the JSON format.

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 - Bugfix: directing geojson diff output to the console didn't work in a multi-dataset repo, even if only one dataset had changed. [#702](https://github.com/koordinates/kart/issues/702)
 - Improve argument parsing for `kart diff` and `kart show`. [#706](https://github.com/koordinates/kart/issues/706)
 - Bugfix: don't allow `kart merge` to fast-forward if the user specifies a merge message. [#705](https://github.com/koordinates/kart/issues/705)
-- Kart now shows diffs to attached files that have been created using `kart commit-files`, if any [#583](https://github.com/koordinates/kart/issues/583)
+- Kart now shows diffs to attached files that have been created using `kart commit-files`, if any, and patches can edit these files [#583](https://github.com/koordinates/kart/issues/583)
```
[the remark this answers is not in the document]
olsen232: Done

kart/apply.py
olsen232: Modified the patch format for text files to start with "text:", but left the other diffs as they were.

PR #714, opened 2022-09-20
Set stdout and stderr when calling subprocess.run

olsen232:
This is what the caller actually intends - subtly different to the default, we reuse the same FDs instead of creating new ones that are connected to the same place. This is required for the Kart-Helper to work properly.

### DIFF
--- a/kart/subprocess.py
+++ b/kart/subprocess.py
@@ -12,8 +12,13 @@ from subprocess import *  # noqa
 def run(*args, **kwargs):
     if "_KART_TEST" in os.environ:
         return run_for_test(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


craigds (approved):

craigds (changes requested):
This has effectively been replaced by fb11ab58626c2b1f6ef33cfc6f71b512252e414c on master; do we need this PR actually?

olsen232:
Would be good to figure this out some more one day - having a separate code path for every unit test is a bit of a liability since it means things could actually break but keep working in the unit tests due to the subtly different code path.
But for now I'll close it

PR #715, opened 2022-09-20
Exec refactor

pfw:
## Description

Refactor the handling of subprocesses which in normal operation don't have their output captured, eg. 'git log'

It is necessary to have at least one mode of capturing output for test running but in normal operation using the standard IO streams directly is more efficient.

## Checklist:


tests/conftest.py
```diff
@@ -446,7 +446,7 @@ def cli_runner(request):
     """A wrapper round Click's test CliRunner to improve usefulness"""
     return KartCliRunner(
         # kart.cli._execvp() looks for this env var to prevent fork/exec in tests.
```
olsen232, line 448: Update this comment - there's no kart.cli._execvp any more

kart/subprocess_util.py
```diff
@@ -138,3 +141,46 @@ def subprocess_tee(cmd, **kwargs):
...
+    # it gets treated as a file like object and fails.
+    p = subprocess.run([cmd] + args[1:], capture_output=True, encoding="utf-8", env=env)
+    sys.stdout.write(p.stdout)
+    sys.stdout.flush()
+    sys.stderr.write(p.stderr)
+    sys.stderr.flush()
+    sys.exit(p.returncode)
+
+
+def run(cmd, args):
```
olsen232, line 163: We should fix up the parameters here - at the calling location, we are sending this:
`run("git", ["git, "gc"])`
And then this helper function always has to get throw away one of the redundant "git"s. We should just send a single array of commands, since that's what subprocess takes

olsen232 (approved):

pfw:
@olsen232 I sorted out the things you mentioned - are you happy with this to be merged?

kart/subprocess_util.py
olsen232: This got closer to keeping the convention of the subprocess API but stopped slightly short for some reason:
exec API: `run("git", ["git", "gc"])` <- where we were
current run function: `run("git", ["gc"])` <- where we are
subprocess API: `run(["git", "gc"])` <- why not just do this

olsen232 (approved):

PR #730, opened 2022-11-07
Add logging.config into pyinstaller packages

craigds:
## Description
If ext-run code wants to configure logging properly it is likely to want `logging.config.dictConfig()`, and this is not currently included in Kart. It's also hard to include via sys.path hackery since `logging` *is* included in Kart.

This change just adds all sub-modules of `logging` into the pyinstaller packages.

## Related links:

n/a

## Checklist:

### DIFF
--- a/kart.spec
+++ b/kart.spec
@@ -36,7 +36,7 @@ if is_win:
             vr.write(vr_doc)
 
 if is_linux:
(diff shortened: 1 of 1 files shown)


craigds:
macOS build failures are unrelated AFAICT

rcoup:
Can you add it to the changelog?

PR #731, opened 2022-11-08
Fix diff feature count estimates for pointclouds

craigds:
## Description

`kart log --output-format=json --with-feature-count=veryfast` was failing for pointcloud datasets, because it was written for tabular datasets and hadn't been updated.

This fixes the issues. To make the implementation simpler, I opted to ignore the accuracy argument, and always return an exact feature-count (tile count) for pointcloud datasets, since there will rarely be >100,000 and so it should be fast.

## Related links:

n/a

## Checklist:

### DIFF
--- a/kart/diff.py
+++ b/kart/diff.py
@@ -103,7 +103,8 @@ def feature_count_diff(
     help=(
         "Returns only a feature count (the number of features modified in this diff). "
         "If the value is 'exact', the feature count is exact (this may be slow.) "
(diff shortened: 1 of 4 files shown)


pfw:
> since there will rarely be >100,000

What are features in this context, the files that make up the dataset?

craigds:
yep, it's a blob count (tile count) essentially.

rcoup:
@craigds docs need updating to say that feature=tile for non-tabular datasets?

olsen232 (approved):

PR #738, opened 2022-11-28
Use tqdm for better progress reporting during WC checkout

olsen232:
Consistent, used for both vector and PC checkout, has some nice features:
- don't need to explicitly log every 10th or 10,000th feature - tqdm decides.
- logs to stderr, but keeps it brief if it is not isatty()
- draws a nice progress bar

Still TODO, one day maybe: use tqdm during clone and fetch operations too

### DIFF
--- a/kart/tabular/table_dataset.py
+++ b/kart/tabular/table_dataset.py
@@ -1,14 +1,11 @@
 import functools
-import logging
-import time
 
 from kart.base_dataset import BaseDataset
 from kart.spatial_filter import SpatialFilter
 from kart.working_copy import PartType
+from kart.progress_util import progress_bar
 
 from .import_source import TableImportSource
-
-L = logging.getLogger("kart.tabular.table_dataset")
 
 
 class TableDataset(BaseDataset, TableImportSource):
@@ -116,7 +113,7 @@ class TableDataset(BaseDataset, TableImportSource):
         geom_columns = self.schema.geometry_columns
         return geom_columns[0].name if geom_columns else None
 
-    def features(self, spatial_filter=SpatialFilter.MATCH_ALL, log_progress=False):
+    def features(self, spatial_filter=SpatialFilter.MATCH_ALL, show_progress=False):
         """
         Yields a dict for every feature. Dicts contain key-value pairs for each feature property,
         and geometries use kart.geometry.Geometry objects, as in the following example::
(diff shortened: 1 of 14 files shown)


kart/tabular/table_dataset.py
```diff
@@ -134,25 +132,16 @@ def features(self, spatial_filter=SpatialFilter.MATCH_ALL, log_progress=False):
...
-        n_chunk = 0
         n_matched = 0
-        n_total = self.feature_count
-        t0 = time.monotonic()
-        t0_chunk = t0
-
         if log_progress:
-            plog("0.0%% 0/%d features... @0.0s", n_total)
+            n_total = self.feature_count
+            progress = tqdm(total=n_total, unit="F", desc=self.path)
```
craigds, line 141: worth using it as a contextmanager instead? makes it clearer that `close()` will be called. and fallback to `contextlib.null_context()` when `not log_progress`

kart/tabular/table_dataset.py
```diff
@@ -165,44 +154,20 @@ def features(self, spatial_filter=SpatialFilter.MATCH_ALL, log_progress=False):
                 n_matched += 1
                 yield feature
 
-            if log_progress and n_chunk == self.NUM_FEATURES_PER_PROGRESS_LOG:
-                t = time.monotonic()
-                self._log_feature_progress(
-                    plog, n_read, n_chunk, n_matched, n_total, t0, t0_chunk, t
+            if log_progress:
+                progress.update(1)
```
craigds, line 158: does the close() do this anyway?

craigds (approved):
Screen Shot 2022-11-29 at 12 08 53 PM



On wide screens it's a bit much; maybe limit to ncols=120 or so?

Otherwise 👍

kart/tabular/table_dataset.py
olsen232: Discussed offline - this actually increments by 1, rather than setting to 100%

kart/tabular/table_dataset.py
olsen232: Done

olsen232:
Setting ncols to a maximum is not done - currently, setting this explicitly seems to ruin tqdm's smart sizing (whereby it will make a smaller progress bar and cut out some cruft if the window is small). We can maybe get this fixed in tqdm one day, otherwise, current behaviour is better even if the progress bars look a bit extreme on a widescreen.

craigds:
ok 👍 
I think I misread the `ncols` doc as meaning a maximum column width, but it's actually an override for the autodetected terminal width.

PR #740, opened 2022-11-30
Fix for #238 - changing PK column (was not supported)

### LINKED ISSUE #238: Changing pk column is not supported
In the V2 structure, changing which column is the primary key means that every feature has to be deleted and rewritten in a new location.
Preliminary tests indicate this works to some extent, but definitely the diff shown for this change will be useless. We need to do more work if we want to support this properly.

olsen232:
Note that changing the PK column disconnects every past feature from every future feature, since the way we know that a feature is the same feature going forward is that it continues to have the same primary key value.

https://github.com/koordinates/kart/issues/238

Fixes #238


craigds (approved):
👍 

does it support a case where the PK field has the same name but a different type? Would it be worth adding a test for that?

PR #750, opened 2022-12-20
Add a kart lfs+ ls-files command for listing LFS tiles

olsen232:
More efficient than git-lfs command since it only looks in relevant datasets. Also improves efficiency of `kart lfs+ pre-push`

TODO: Revisit this if and when git rev-list --objects pathspec behaviour is fixed, so we can make more use of it.

## Checklist:

- [ ] Have you updated the [changelog](https://github.com/koordinates/kart/blob/master/CHANGELOG.md)? Part of point cloud changes, CHANGELOG yet to be updated with documentation for that release

### DIFF
--- a/kart/rev_list_objects.py
+++ b/kart/rev_list_objects.py
@@ -51,16 +51,16 @@ def rev_list_object_oids(repo, start_commits, stop_commits, pathspecs):
         "--stdin",
     ]
     try:
-        p = subprocess.Popen(
+        with subprocess.Popen(
             cmd,
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             encoding="utf8",
             env=tool_environment(),
-        )
-        p.stdin.write(os.linesep.join(["--", *pathspecs]))
-        p.stdin.close()
-        yield from _parse_revlist_output(repo, p.stdout)
+        ) as p:
+            p.stdin.write(os.linesep.join(["--", *pathspecs]))
+            p.stdin.close()
+            yield from _parse_revlist_output(repo, p.stdout)
     except subprocess.CalledProcessError as e:
         raise SubprocessError(
             f"There was a problem with git rev-list: {e}", called_process_error=e
@@ -71,6 +71,7 @@ def get_dataset_pathspecs(repo, start_commits, stop_commits, dirname_filter):
     """
     Get the list of dataset paths we need to search in for the objects we are insterested in.
     Without this, all datasets are searched - even irrelevant ones - which could waste a lot of time.
+    dir_name_filter should match whichever dataset-dirnames are relevant to search, eg ".table-dataset"
     """
     # TODO - if git rev-list --objects pathspec filtering is fixed, there would be other approaches
     # we could take, some potentially using only a single call to rev-list.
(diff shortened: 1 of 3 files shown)


kart/rev_list_objects.py
```diff
@@ -28,51 +34,99 @@ def _rev_list_objects_command(repo):
...
 
-def rev_list_blobs(repo, start_commits, stop_commits):
+def get_dataset_pathspecs(repo, start_commits, stop_commits, dirname_filter):
+    """
+    Get the list of dataset paths we need to search in for the objects we are insterested in.
+    Without this, all datasets are searched - even irrelevant ones - which could waste a lot of time.
+    """
+    # TODO - if git rev-list --objects pathspec filtering is fixed, there would be other approaches
+    # we could take, some potentially using only a single call to rev-list.
+    # See https://public-inbox.org/git/CAPJmHpWWJ4sssfG2oym7K=MsT3+KTHQP-QK88nfXNOtfcv07ew@mail.gmail.com/
```
craigds, line 77: did you get any replies? can't find any way to browse by thread

kart/rev_list_objects.py
```diff
@@ -28,51 +34,99 @@ def _rev_list_objects_command(repo):
...
+    cmd = [
+        *_rev_list_commits_command(repo),
+        *start_commits,
+        "--not",
+        *stop_commits,
+    ]
+    result = set()
+    try:
+        p = subprocess.Popen(
```
craigds, line 86: use the Popen as a contextmanager here to ensure the process is waited for if an exception happens while iterating stdout below

kart/rev_list_objects.py
```diff
@@ -28,51 +34,99 @@ def _rev_list_objects_command(repo):
...
+            cmd,
+            stdout=subprocess.PIPE,
+            encoding="utf8",
+            env=tool_environment(),
+        )
+        for line in p.stdout:
+            commit = repo[line.strip()]
+            for tree_path, tree in all_trees_with_paths_in_tree(commit.tree):
+                if tree_path in result:
```
craigds, line 95: this looks incorrect - shouldn't this be skipping based on tree.oid rather than tree path? Otherwise you'll only iterate a dataset's tree in the first commit you come across

kart/rev_list_objects.py
olsen232: Not so far :-/ and I also didn't spend the necessary time to debug it. It already looks complicated and it doesn't work - so either I need to fix some complicated logic that's gone wrong, or, I need to make it even more complicated to make it actually fulfil the spec

kart/rev_list_objects.py
olsen232: This is correct although it may not be doing exactly what you think:
1. all_trees_with_paths_in_tree() iterates over all the trees that aren't "hidden" ie a, a/1, a/2, b, b/1, b/2
2. This code here checks each of those trees to see if they contain any hidden directories that mean they are datasets we should be checking. If it does, we add the path to the result set. But there's no point adding the path to the result set if it's already in the result set - so in that case, we can skip the check to see if it's a dataset, we already know it sometimes is.

I'll add a comment

kart/rev_list_objects.py
olsen232: Done

olsen232:
Forgot to add the test - added now

craigds (approved):

PR #751, opened 2022-12-21
Help the user fix their working copy if it gets into a bad state

olsen232:
If Kart crashes or otherwise is aborted partway through a checkout operation, it is possible for the contents of the working copy or individual working copy parts to end up out of sync with what should be checked out based on the git file `.kart/HEAD`.

Until now, this caused an internal error to be raised - WorkingCopyTreeMismatch, which gave some clues as to what had gone wrong but no real help on how to fix it. Since Kart checks the WC state is valid as a precondition to many commands, arriving in this state with no clear idea of how to get out of it makes a Kart repo feel pretty broken, as most commands will simply raise the same internal error.

This change adds an external error including error code (which also suppresses the stack trace). The error message
contains instructions on how to fix the repo so that it can be used again, and if run in an interactive terminal, prompts
the user to decide if they want to try this remedy immediately.

### DIFF
--- a/kart/working_copy.py
+++ b/kart/working_copy.py
@@ -471,8 +471,11 @@ def handle_working_copy_tree_mismatch(wc_type_name, actual_tree_id, expected_tre
         f"tree {expected_tree_id}" if expected_tree_id else "the empty tree"
     )
 
+    summary = (
+        f"The {wc_type_name} working copy appears to be out of sync with the repository"
+    )
     message = [
-        f"The {wc_type_name} working copy appears to be out of sync with the repository:",
+        f"{summary}:",
         f"  * The working copy's own records show it is tracking {actual_tree_id};",
         f"  * Based on the repository it should be tracking {expected_tree_id}.",
         "The simplest fix is generally to recreate the working copy (losing any uncommitted changes in the process.)",
@@ -486,7 +489,7 @@ def handle_working_copy_tree_mismatch(wc_type_name, actual_tree_id, expected_tre
     click.echo("\n".join(message))
     click.echo()
     if not click.confirm("Do you want to recreate the working copy?"):
-        raise BadStateError(message[0], exit_code=BAD_WORKING_COPY_STATE)
+        raise BadStateError(f"{summary}.", exit_code=BAD_WORKING_COPY_STATE)
 
     from kart.create_workingcopy import create_workingcopy
 
@@ -497,9 +500,4 @@ def handle_working_copy_tree_mismatch(wc_type_name, actual_tree_id, expected_tre
     subctx.invoke(create_workingcopy, delete_existing=True, discard_changes=True)
 
     orig_command = f"{ctx.command_path} {' '.join(ctx.unparsed_args)}"
-    click.echo(f"\nThe original command was: {orig_command}")
-
-    if not click.confirm("Do you want to continue running the original command?"):
-        sys.exit(0)
-
-    click.echo()
+    click.echo(f"\nContinuing with the original command: {orig_command}\n")
--- a/kart/working_copy.py
+++ b/kart/working_copy.py
@@ -472,22 +472,20 @@ def handle_working_copy_tree_mismatch(wc_type_name, actual_tree_id, expected_tre
     )
 
     message = [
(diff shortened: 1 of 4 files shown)


kart/working_copy.py
```diff
@@ -462,3 +463,38 @@ def _check_for_unsupported_structural_changes(
...
+
+    if get_input_mode() != InputMode.INTERACTIVE:
+        message.append("To do so, try the following command:")
+        message.append("kart create-workingcopy --delete-existing --discard-changes")
+        raise BadStateError("\n".join(message), exit_code=BAD_WORKING_COPY_STATE)
+
+    click.echo("\n".join(message))
+    click.echo()
+    if not click.confirm(
+        "Do you want to recreate the working copy from the ground up?"
```
craigds, line 489:
```suggestion
        "Do you want to recreate the working copy?"
```

I'd bet that expression doesn't make sense for non-native English speakers and it doesn't really add anything

kart/working_copy.py
```diff
@@ -462,3 +463,38 @@ def _check_for_unsupported_structural_changes(
...
+    actual_tree_id = f"tree {actual_tree_id}" if actual_tree_id else "the empty tree"
+    expected_tree_id = (
+        f"tree {expected_tree_id}" if expected_tree_id else "the empty tree"
+    )
+
+    message = [
+        f"The {wc_type_name} working copy appears to be out of sync with the repository.",
+        f"The working copy's own records show it is tracking {actual_tree_id};",
+        f"based on the repository it should be tracking {expected_tree_id}.",
+        "The most straight-forward fix is generally to recreate the working copy from the ground up (losing any uncommitted changes in the process.)",
```
craigds, line 478:
```suggestion
        f"The {wc_type_name} working copy appears to be out of sync with the repository:",
        f"  * The working copy's own records show it is tracking {actual_tree_id};",
        f"  * Based on the repository it should be tracking {expected_tree_id}.",
        "The simplest fix is generally to recreate the working copy (losing any uncommitted changes in the process.)",
```

kart/working_copy.py
```diff
@@ -462,3 +463,38 @@ def _check_for_unsupported_structural_changes(
...
+    ]
+
+    if get_input_mode() != InputMode.INTERACTIVE:
+        message.append("To do so, try the following command:")
+        message.append("kart create-workingcopy --delete-existing --discard-changes")
```
craigds, line 483:
```suggestion
        message.append("\nTo do so, try the following command:")
        message.append("\tkart create-workingcopy --delete-existing --discard-changes")
```

Otherwise it's very

kart/working_copy.py
```diff
@@ -462,3 +463,38 @@ def _check_for_unsupported_structural_changes(
...
+        raise BadStateError(message[0], exit_code=BAD_WORKING_COPY_STATE)
+
+    from kart.create_workingcopy import create_workingcopy
+
+    click.echo("Recreating working copy ...")
+    ctx = click.get_current_context()
+    subctx = click.Context(ctx.command, parent=ctx)
+    subctx.obj = ctx.obj
+    subctx.invoke(create_workingcopy, delete_existing=True, discard_changes=True)
+    sys.exit(0)
```
craigds, line 500: If you get here by running another command, seems like we should not exit here and instead just run the original command

craigds (changes requested):

kart/working_copy.py
olsen232: Done

kart/working_copy.py
olsen232: Done

kart/working_copy.py
olsen232: Added a prompt to see if the user still wants to do it. Didn't want to do it by default since it may not still make sense (ie, probably the command depends on the state of the WC in some way or we wouldn't have got here).

craigds: Can you give an example of what they might be trying to do that would now be a bad thing to do?

status/diff/commit would all be sensible here AFAICT (commit would just noop due to no changes)

craigds (approved):

PR #760, opened 2023-01-10
Delete vendor/ - obsoleted by vcpkg-vendor/

olsen232:
Also delete the Makefile that builds using vendor, and update the docs slightly.


rcoup:
LGTM

Should clean up `platforms/` too (a few files from there are used by CPack). And I guess we should make `-DUSE_VCPKG=ON` the default?

rcoup:
> Should clean up platforms/ too (a few files from there are used by CPack). 

Added this

> And I guess we should make -DUSE_VCPKG=ON the default?

lets leave this for a little bit

PR #761, opened 2023-01-11
Improve contributing.md

olsen232:
- More detail on downloading CI artifacts, including scrolling issue
- More detail on DPython_ROOT
- More consistency between windows and posix
- Added some section links at the top

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,12 +37,12 @@ $ cd kart
 $ git submodule update --init --recursive
 ```
 
-### Setting Python3_ROOT
+### Setting Python3
 
 CMake will try to automatically find a Python3 install on your system with which to build Kart.
 However, to avoid any issues caused by different Python implementations, it is recommended to force CMake
-to use Python 3.10, by supplying the flag -DPython3_ROOT - for example:
-`-DPython3_ROOT=$(realpath $(which python3.10) | sed 's@/bin/.*@/@')`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -34,11 +36,21 @@ $ git clone https://github.com/koordinates/kart.git
 $ cd kart
 $ git submodule update --init --recursive
 ```
+
+### Setting Python3_ROOT
+
+CMake will try to automatically find a Python3 install on your system with which to build Kart.
+However, to avoid any issues caused by different Python implementations, it is recommended to force CMake
+to use Python 3.10, by supplying the flag -DPython3_ROOT - for example:
+`-DPython3_ROOT=$(realpath $(which python3.10) | sed 's@/bin/.*@/@')`
```
rcoup, line 45: can also do `-DPython3_EXECUTABLE=$(command -v python3.10)` — dunno which is better? Executable might be easier for users.

rcoup (approved):
LGTM. Minor suggestion but feel free to ignore.

CONTRIBUTING.md
olsen232: Fine by me as long as this is as powerful as setting DPython3_ROOT, ie, cmake correctly finds the root from the executable and doesn't require you to set any other DPython3 variables

olsen232: Seems to work, done

PR #767, opened 2023-01-17
Fix for bug where wheels can't be found...

olsen232:
... when the minor-version of the osx deployment target is implicitly 0

Isn't causing CI build failures currently, but may prevent Kart from building locally, depending on your setup.


craigds (approved):
👍 This makes my build on macos ventura get a lot further

PR #775, opened 2023-01-20
Warn if reflink isn't available...

olsen232:
... and make sure we're using reflink everywhere we can
- reflink implementation was initially a POC and so was missing lots of places

### DIFF
--- a/kart/reflink_util.py
+++ b/kart/reflink_util.py
@@ -8,33 +8,11 @@ def reflink(from_, to):
     rl.reflink(str(from_), str(to))
 
 
-def try_reflink(*args):
-    """
-    Usage:
-    >>> try_reflink(from_, to) - tries to use reflink to copy, but falls back to shutil.copy
-    >>> copy_ = try_reflink() - returns a callable object that tries to use reflink to copy until
-        that fails, and from then on it uses shutil.copy
-    """
+def try_reflink(from_, to):
+    """Same as reflink_util.reflink, but falls back to shutil.copy if reflinking fails."""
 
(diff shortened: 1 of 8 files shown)


kart/reflink_util.py
```diff
@@ -0,0 +1,40 @@
...
+
+import reflink as rl
+
+
+def reflink(from_, to):
+    """Same as reflink.reflink, but can be called with pathlib.Path objects."""
+    rl.reflink(str(from_), str(to))
+
+
+def try_reflink(*args):
```
craigds, line 11: this thing's a bit wacky? IMO callers should just call two different functions, rather than one function that does two different things.

kart/reflink_util.py
```diff
@@ -0,0 +1,40 @@
...
+    reflink_copier = ReflinkCopier()
+    if args:
+        return reflink_copier(*args)
+    else:
+        return reflink_copier
+
+
+class ReflinkCopier:
+    """
+    A callable object that, when called with (from_, to), calls reflink(from_, to) -
```
craigds, line 28: Is this necessary? I can see the usefulness _if_ trying and failing to create a reflink is expensive, otherwise I'd say we should just try to reflink each file and then copy it if that doesn't work out

craigds (approved):

kart/reflink_util.py
olsen232: You're right. Done

kart/reflink_util.py
olsen232: Done

PR #777, opened 2023-01-22
improve startup time by lazy-importing jsonschema 

rcoup:
## Description

About 19% of kart initialisation relates to importing [jsonschema](https://github.com/python-jsonschema/jsonschema) (via `-X importtime` analysis). We only use it in a single place, so lazy-import it. In addition, versions 4.0-4.2 have significant performance issues, so upgrade to the latest release.

Fix some issues wrt py-requirements compilation to enable that upgrade.

## Checklist:

### DIFF
--- a/kart/cli_util.py
+++ b/kart/cli_util.py
@@ -11,7 +11,6 @@ import sys
 from pathlib import Path
 
 import click
-import jsonschema
 import pygit2
 from click.core import Argument
 from click.shell_completion import CompletionItem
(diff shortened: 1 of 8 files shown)


olsen232:
I downloaded the linux version of this and checked for any startup speed up compared to 0.12.0b9 - couldn't detect it, so maybe it doesn't affect the commands that I'd been testing, or maybe it was nearly exactly counteracted by something else that's going on (not sure if non-release builds differ from release builds in some way on linux).

But anyway not importing jsonschema up front, and fixing requirements, can only be good so 👍

PR #778, opened 2023-01-23
Add pages about working with point cloud datasets...

olsen232:
... and table datasets, for consistency / completeness. Stopped short of adding another whole tutorial that uses point cloud tiles, the original tutorial does a reasonable job of covering this anyway.


docs/pages/pointcloud_datasets.rst
```diff
@@ -0,0 +1,37 @@
...
+
+- ``kart import <tile> [<tile>] [<tile>]``
+- ``kart import --dataset=<dataset_name> <tile> [<tile>] [<tile>]``
+
+This command imports one or more point-cloud tiles into the kart repository in the
+working directory. All tiles are imported into the same dataset.
+
+The tiles to import must be LAS files of some variant - compressed tiles (LAZ) and cloud-optized tiles (COPC) are allowed.
+By default, all tiles are converted to COPC (`Cloud Optimized Point Cloud <copc_>`_) as they are imported.
+You can specify ``--keep-existing-format`` to keep them as they are.
```
craigds, line 21: I'd prefer to steer users towards COPC rather than be totally agnostic. Is it worth adding a note about why it might be best to allow Kart to convert them? Just something simple like

> COPC is recommended as it improves the efficiency with which apps and cloud services can access large point cloud datasets.

docs/pages/table_datasets.rst
```diff
@@ -0,0 +1,46 @@
+Working with Vector / Table Datasets
+====================================
+
+.. Note:: Before using this quick guide, it will help your understanding to do the :doc:`Basic Usage Tutorial </pages/basic_usage_tutorial>`.
+
+Dataset type
+~~~~~~~~~~~~
+Both spatial and non-spatial tabular datasets are stored as "Table Datasets", currently implemented as :doc:`Table Datasets V3 </pages/development/table_v3>`. Version controlled table rows are stored in an abstract way, not unlike JSON, and independent of any of the supported working copy types, and so are converted to fit into a specific type of database table during checkout.
```
craigds, line 8:
```suggestion
Both spatial and non-spatial tabular datasets are stored as "Table Datasets", currently implemented as :doc:`Table Datasets V3 </pages/development/table_v3>`. Version controlled table rows are stored in an abstract way, independent of any of the supported working copy types, and so are converted to fit into a specific type of database table during checkout.
```

craigds (approved):

docs/pages/pointcloud_datasets.rst
olsen232: Done

docs/pages/table_datasets.rst
olsen232: Done

PR #783, opened 2023-02-01
Modified `point-cloud-import --replace-existing` to reuse tiles

olsen232:
Modified `point-cloud-import --replace-existing` to reuse previously imported tiles, rather than reimport them, if tiles that have already been imported are imported again - potentially saving time and disk space.

This is particularly relevant when importing with convert-to-copc, since converting to COPC can make the hash differ each time and so cause the same points to be stored more than once in equivalent but not identical files, at different hashes.

Note that `point-cloud-import --update-existing` already had this optimization.

## Checklist:


rcoup (approved):
👍

Issue #508, opened 2021-11-19
"`<option>` is unknown to Kart and will be passed directly to git"


craigds:
Hi! You might have arrived at this issue page via this deprecation warning from `kart log`:

```
RemovalInKart012Warning: '<option>' is unknown to Kart and will be passed directly to git.
This will be removed in Kart 0.12! Please comment on https://github.com/koordinates/kart/issues/508 if you
need to use this option.
```

We're seeking feedback about which of these `log` options are being used, so that we can add them properly to Kart. Please comment below, so we can avoid breaking things for you in Kart 0.12.

PR #784, opened 2023-02-02
Reintroduce support for `kart log --graph`

olsen232:
Also update the CHANGELOG which was missing the fact that support for this among other options was dropped as part of the 0.12 release (sorry)


craigds (approved):
LGTM 

>  (sorry)

we deprecated it as best we could, for over a year (#508) so 🤷 


We probably need a change to the qgis-plugin repo to raise the minimum kart version to 0.12.1

PR #785, opened 2023-02-06
Add check_git_user to more commands

olsen232:
If we don't check_git_user, these commands will fail. With the check, we still fail, but with the appropriate error message and the appropriate error code, which the caller can pick up on.


rcoup:
Would it be easier to make this a command decorator and doing it upfront? At present it's at the "just about to make changes", skipped for dry-runs, etc — but I'm not sure whether that's a good thing or a footgun (the test doesn't error but the actual one does).

olsen232:
Not sure I'm convinced - consider `merge`. This is invoked by `pull`. A user might not ever make any changes, and only ever run `pull` to keep their repo up to date - they don't actually need a git identity for this, as long as nothing more complicated happens.
However, it could be confusing that - as and when something more complicated does happen - the first they learn about it is that Kart starts asking for their email address. So, maybe asking up front is a good idea. Basically it would be these commands:

kart commit
kart merge
kart pull
And then the ones that directly update files / metadata: kart meta set, kart commit-files

Issue #788, opened 2023-02-10
Terminal-based editor not working with kart helper


olsen232:
**Describe the bug**
Launching a terminal-based editor isn't working from within Kart 0.12.x on macos or linux where the helper mode is enabled. GUI-based editors are unaffected. This is the editor launched in the function `run_editor_cmd`, that runs `$EDITOR`

**To Reproduce**
Steps to reproduce the behaviour:
1. Make sure you are using Kart 0.12 or later
2. Make an uncommitted edit to a kart repo - eg, connect to the working copy and change a dataset's title / add a new feature
3. Run `kart commit` to commit the edit
4. Editor fails to open, with error message a something like the following:
`Error: There was a problem with the editor 'nano': Command 'nano /Users/olsen232/mykart/repo123/.kart/COMMIT_EDITMSG' returned non-zero exit status 1.`

**Expected behaviour**
An editor should open, in order that you can write the commit message.

**Workaround**
Disable the helper by running kart like so:
`KART_USE_HELPER=0 kart commit`
Or, avoid the need to use the editor by specifying a message up front like so:
`kart commit -m "commit message"`

**Version Info**
 - OS: macOS or linux
 - Kart version - 0.12.x

pfw:
This is due to the helper itself being disconnected from the terminal by the double fork done - there are a couple of cases where this can happen, I believe if GIT_ASKPASS is set and required you will get a similar error. I had considered making a list of commands in the calling process which would automatically skip using the helper as appropriate.

olsen232:
Fixed as of Kart 0.12.2

PR #789, opened 2023-02-12
Fix helper mode + editor incompatibility

olsen232:
Fix is to not use helper mode for kart commit, merge or pull.
See https://github.com/koordinates/kart/issues/788


craigds (approved):
LGTM

PR #790, opened 2023-02-14
List untrackeddd tables issue 609

fedch:
## Description
- Added support for `--list-untracked-tables` option under `kart status` command to list the tables/datasets that are added to the repository but are not tracked by Kart.

## Testing:

Imported a package into the kart repo:
```
test % kart status --list-untracked-tables
On branch main

Nothing to commit, working copy clean

No untracked tables found.
```
---
json output:
```
test % kart status --list-untracked-tables -o json
{
  "kart.status/v2": {
    "commit": "14cd5e4ad4accbc0124d59ef07b794509d9543a8",
    "abbrevCommit": "14cd5e4",
    "branch": "main",
    "upstream": null,
    "spatialFilter": null,
    "untrackedTables": [],
    "workingCopy": {
      "parts": {
        "tabular": {
          "location": "test1.gpkg",
          "type": "gpkg",
          "status": "ok"
        },
        "workdir": {
          "status": "notFound"
        }
      },
      "changes": {},
      "untrackedTables": []
    }
  }
}
```
---
Added 2 test tables in QGIS:

```
test % kart status --list-untracked-tables
On branch main

Nothing to commit, working copy clean

Untracked tables:
  test_table1
  test_table_2
  test_table3
```
---
json output:
```
test % kart status --list-untracked-tables -o json
{
  "kart.status/v2": {
    "commit": "72683b15cbdf727c5092c9ad7ce7534c3093e3f5",
    "abbrevCommit": "72683b1",
    "branch": "main",
    "upstream": null,
    "spatialFilter": null,
    "untrackedTables": [
      "test_table3",
      "test_table1",
      "test_table_2"
    ],
    "workingCopy": {
      "parts": {
        "tabular": {
          "location": "test.gpkg",
          "type": "gpkg",
          "status": "ok"
        },
        "workdir": {
          "status": "notFound"
        }
      },
      "changes": {},
      "untrackedTables": [
        "test_table3",
        "test_table1",
        "test_table_2"
      ]
    }
  }
}
```

## Checklist:


kart/status.py
```diff
@@ -100,8 +105,33 @@ def status(ctx, output_format):
...
     if output_format == "json":
         dump_json_output({"kart.status/v2": jdict}, sys.stdout)
-    else:
+    else:       
         click.echo(status_to_text(jdict))
+        
+        if list_untracked_tables:
+            """Check for any untracked tables in working copy"""
+            wc = repo.working_copy.tabular 
+            click.echo(wc)
```
olsen232, line 114: Debug output

kart/status.py
```diff
@@ -100,8 +105,33 @@ def status(ctx, output_format):
...
+            if wc is not None:
+                if wc.session() is not None:        
+                    with wc.session() as sess: 
+                        wc_items = wc.adapter.list_tables(sess)
+                    # Get all tables in working copy
+                    all_tables = [table_name for table_name, title in wc_items.items()]
+                    # Get tables shown in kart data ls
+                    datasets_paths = [datasets.path for datasets in repo.datasets()]
```
olsen232, line 122: Two comments here:
1. Use the attribute `dataset.table_name` - a dataset's path is allowed to be slightly different to its table name, eg the dataset with path `aaa/bbb` is checked out into a table called `aaa__bbb` I think
However: not all datasets have a table_name attribute, so, use repo.datasets(filter_dataset_type="table") to make sure you only get the table datasets which have this attribute.
2. (Nitpick) `datasets_paths = [datasets.path for datasets in repo.datasets()]` - the variable `datasets` should be `dataset` since it is singular - it points to a single dataset (at a time)

kart/status.py
```diff
@@ -100,8 +105,33 @@ def status(ctx, output_format):
 
     if output_format == "json":
         dump_json_output({"kart.status/v2": jdict}, sys.stdout)
-    else:
+    else:
```
olsen232, line 108: This code is looking good as far as it goes - my next challenge to you is to update it to be more in line with the existing code, which has the following pattern:

1. Generate the status in a JSON-like-format.
2. a) If "json": dump that status as JSON
2. b) Else if "text": convert the status to text and print it

There's a couple of upsides to doing it this way - anything we expose in "text" output we are forced to also expose in "json" output, and, we can guarantee that the output is consistent (ie, if we output that there are untracked tables when doing "json" output, then we will be sure to also output it when doing "text" output.

I don't know exactly where the untracked tables should go in the JSON output. For now, you could put them in a list under "workingCopy" called "untrackedTables", and hopefully we'll know if that's a good place by by the time this is ready to merge.

kart/status.py
```diff
@@ -78,7 +78,12 @@ def get_type_counts(self):
     type=click.Choice(["text", "json"]),
     default="text",
 )
-def status(ctx, output_format):
+@click.option(
+    "--list-untracked-tables",
+    is_flag=True,
+    help="Shows which tables haven't yet been tracked by cart"
```
craigds, line 84:
```suggestion
    help="Shows which tables haven't yet been tracked by Kart"
```

kart/status.py
```diff
@@ -96,14 +103,28 @@ def status(ctx, output_format):
...
+                    with wc.session() as sess:
+                        wc_items = wc.adapter.list_tables(sess)
+                    # Get all tables in working copy
+                    all_tables = [table_name for table_name, title in wc_items.items()]
+                    # Get tables shown in kart data ls
+                    datasets_paths = [dataset.table_name for dataset in repo.datasets(filter_dataset_type="table")] 
+                    # Get untracked tables
+                    untracked_tables = list(set(all_tables) - set(datasets_paths))
+                    jdict["untrackedTables"] = untracked_tables
+        jdict["workingCopy"] = get_working_copy_status_json(repo, untracked_tables, is_untracked_tables_call)
```
olsen232, line 121: The code here is fine, but it is in a strange order - it could be clearer if it was moved around.

Here is a nice way to structure code:
```
def do_big_task():
  result = ...
  result += do_first_small_task()
  result += do_second_small_task()
  if necessary:
    result += do_third_small_task()
```

The code right now however looks a bit more like this:
```
if necessary:
  third_small_task = doThirdSmallTask()
do_big_task(third_small_task=third_small_task)
```

So the big task you are doing is `get_working_copy_status_json`, and the smaller subtask you are doing within that is finding the untracked tables. `get_working_copy_status_json` could take a bool which controls whether or not to add the untracked tables. Then actually finding the untracked tables could be a subtask of that big task. Does that make sense?

olsen232 (commented):
I have another task for you - add a unit test which shows that this work / will break if this stops working.

You would need to modify test_status.py with a new test that loads a test-repository, modifies the working copy to have untracked tables, and then runs the command and asserts that it gets the expected output. I can help with this

kart/status.py
```diff
@@ -78,9 +78,16 @@ def get_type_counts(self):
...
+@click.option(
+    "--list-untracked-tables",
+    is_flag=True,
+    help="Shows which tables haven't yet been tracked by Kart"
+)
+def status(ctx, output_format, list_untracked_tables):
     """Show the working copy status"""
     repo = ctx.obj.get_repo(allowed_states=KartRepoState.ALL_STATES)
+    untracked_tables = []
+    is_untracked_tables_call = False
```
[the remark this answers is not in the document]
fedch: Hm, I did this to separate 4 scenarios:

Calling --untracked_tables while having untracked tables
Calling --untracked_tables while not having untracked tables
Not calling --untracked_tables (only calling `kart status`) while having untracked tables
Not calling --untracked_tables  while not having untracked tables

If only checking the presence of untracked tables, it will append the output with the result of this check even when calling only `kart status`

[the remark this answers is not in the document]
olsen232: I understand the four separate scenarios - I will address them in a moment -  but this variable is still a duplicate: you have a variable `list_untracked_tables` and another variable `is_untracked_tables_call`. You copy the value of `list_untracked_tables` onto `is_untracked_tables_call`. Anything you use `is_untracked_tables_call` you could more simply use `list_untracked_tables` for to avoid creating a duplicate variable.

Regarding separating the four scenarios. As noted above, you can resolve them without creating a duplicate variable - but I propose the following logic, which I think is simpler:
1. When generating JSON, you wire through the `list_untracked_tables` variable so you know whether you should generate the list of untrackd tables or not.
2. If it is True, you set the status["workingCopy"]["untrackedTables"] field. If there are untracked tables, you set it to a list of those, if there are none, you set it to an empty list.
3. When converting the JSON to text, you don't need to wire through any variables at all.
Instead, you test for the presence of the "untrackedTables" key. You can do this with something like `if "untrackedTables" in some_dict:`
4. If it is not present, you don't output anything about untracked tables. If it is present, you check if it is empty or not in order to decide what to output.

(Note that there are really only 3 separate scenarios since, if the user doesn't call --untracked-tables, we don't actually know if there are untracked tables or not)

olsen232 (commented):
Looks good! Can you add a unit test? I can help you with that

olsen232 (approved):
In CONTRIBUTING.md it recommends using black to format your code - you can do this by installing the pre-commit hooks: `pre-commit install --install-hooks`

And finally, when merging this to master, use the squash-and-merge option - you have 17 commits which is fine on your dev branch, but it would be more noisy than useful to copy all 17 of those commits to master

Issue #249, opened 2020-09-09
No way to create layers by modifying the working copy


olsen232:
If you delete a layer in the working copy, then you should be able to commit this to delete the layer in your repo.

If you add a layer to a working copy, then you should be able to commit this to add the layer to your repo.
This is a bit trickier: the layer will need to be modified to have the triggers added etc.

olsen232:
Deleting works, but you still can't add a layer just by manually creating it in your working copy.

Zumitify:
Hello @olsen232, [Sumit](https://github.com/koordinates/kart/discussions/607) here, 
I am new to the Kart community and I would like to work on this issue if no one else is already looking into it.

I have a few questions regarding the issue.
So I tried the following:
1) Initialized Kart repo and opened the default layer using QGIS
2) Tried editing attributes and they showed up in `kart diff` correctly. ( This was followed by committing the changes)
3) Created 2 more layers (Shapefile and SpatiaLite layers) in the same QGIS project and edited those two layers.
After editing and saving these changes, nothing was reflected in `kart diff`.
Note: These 2 layers are visible in the layers panel of QGIS and not under the Kart plugin-> repo ->Datasets.
> If you add a layer to a working copy
>the layer will need to be modified to have the triggers added etc.

-  Is there any other procedure to add layers to the working copy? Because in my case, there was no diff, and hence no chance of committing even after modifying the changes for the newly added layers.

Also, the expectation of the issue is that: Once a layer is added into the working copy, that itself should set the `diff` trigger (that something has changed in the repo and therefore allow a `commit`) without having to explicitly make changes in the newly added layer.
-  Is my understanding of the issue correct?
 
Thanks


> If you delete a layer in the working copy, then you should be able to commit this to delete the layer in your repo.
> 
> If you add a layer to a working copy, then you should be able to commit this to add the layer to your repo. This is a bit trickier: the layer will need to be modified to have the triggers added etc.

craigds:
Currently the only eay to add a dataset is by using the `kart import` command. Creating a dataset directly in the working copy doesn't work - kart doesn't start tracking changes and producing diffs and allow commits for tables it doesn't know about, hence this commit.

Zumitify:
Hey @craigds, thanks for clarifying my query. I misunderstood QGIS layers for datasets/tables. 
So the high-level tasks to achieve the solution would be: 
1) To have some kind of a listener on the working copy that detects a new dataset(layer) is being added.
[...]

PR #801, opened 2023-03-03
Commit new table to a kart repo  #249

fedch:
## Description

- Added support for committing new layers to a kart repository. When adding a new layer to a working copy, you can now commit the layer to add it to your kart repository. Previously, you could only commit changes to existing layers that were already tracked by kart.
To commit a new layer, use the `add-dataset` command with the name of the layer and a commit message. If the layer is not yet tracked by kart, it will be added to the repository as a new layer.

## Related links:

[Issue 249](https://github.com/koordinates/kart/issues/249)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 - Fix the issue where re-importing the same shapefile with `--replace-existing` would show spurious schema changes (and therefore row changes). [#791](https://github.com/koordinates/kart/issues/791)
 
+- Added support for committing new layers to a kart repository. When adding a new layer to a working copy, you can now commit the layer to add it to your kart repository. Previously, you could only commit changes to existing layers that were already tracked by kart.
+To commit a new layer, use the `add-dataset` command with the name of the layer and a commit message. If the layer is not yet tracked by kart, it will be added to the repository as a new layer.
+
 ## 0.12.2
 
 - Fix for issue with launching an editor while using helper mode, affects macOS and Linux. [#788](https://github.com/koordinates/kart/issues/788)
--- a/kart/add_dataset.py
+++ b/kart/add_dataset.py
@@ -9,7 +9,7 @@ from .commit import (
 )
 from .output_util import dump_json_output
 from .working_copy import WorkingCopyPart
-from .exceptions import NO_CHANGES, NotFound
+from .exceptions import NO_CHANGES, NotFound, InvalidOperation
 from .status import get_untracked_tables
 
 
@@ -41,12 +41,27 @@ from .status import get_untracked_tables
     default="text",
 )
 def add_dataset(ctx, table_name, message, launch_editor, output_format):
+    """
+    Add a new table to the Kart repository.
+
+    To check the new untracked tables, run 'kart status --list-untracked-tables'
+    """
     repo = ctx.obj.repo
+
     # Check that the table is in the list of untracked tables:
     untracked_tables = get_untracked_tables(repo)
     if table_name not in untracked_tables:
+        # Check if the table is already tracked:
+        ds_paths = list(repo.datasets().paths())
+        for ds_path in ds_paths:
+            if ds_path == table_name:
+                raise InvalidOperation(
+                    f"Table '{table_name}' is already tracked\n",
+                    exit_code=NO_CHANGES,
+                )
+
         raise NotFound(
-            f"""Table {table_name} is either not found or already tracked by kart.\n\nTry running 'kart status --list-untracked-tables'\n""",
+            f"""Table '{table_name}' is not found\n\nTry running 'kart status --list-untracked-tables'\n""",
             exit_code=NO_CHANGES,
         )
 
--- a/kart/add_dataset.py
+++ b/kart/add_dataset.py
@@ -1,0 +1,98 @@
+import click
+import sys
+from .cli_util import StringFromFile
+from .commit import (
+    commit_obj_to_json,
+    commit_json_to_text,
+    get_commit_message,
+    CommitDiffWriter,
+)
+from .output_util import dump_json_output
+from .working_copy import WorkingCopyPart
+from .exceptions import NO_CHANGES, NotFound
+from .status import get_untracked_tables
+
+
+@click.command()
+@click.argument("table_name")
+@click.pass_context
+@click.option(
+    "--message",
+    "-m",
+    multiple=True,
+    help=(
+        "Use the given message as the commit message. If multiple `-m` options are given, their values are "
+        "concatenated as separate paragraphs."
+    ),
+    type=StringFromFile(encoding="utf-8"),
+)
+@click.option(
+    " /--no-editor",
+    "launch_editor",
+    is_flag=True,
+    default=True,
+    hidden=True,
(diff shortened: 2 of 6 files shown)


kart/tabular/working_copy/base.py
```diff
@@ -430,27 +430,30 @@ def is_dirty(self):
             return True
 
     def diff_repo_to_working_copy(
```
craigds, line 432: suggest not calling this at all - just pull your changes out separately

kart/tabular/working_copy/base.py
```diff
@@ -483,13 +496,19 @@ def diff_dataset_to_working_copy(
         ds_diff["feature"] = feature_diff
         return ds_diff
 
+
     def diff_dataset_to_working_copy_meta(self, dataset, raise_if_dirty=False):
```
craigds, line 500: you don't really need to use this - just pull the bits out that you need

kart/tabular/working_copy/base.py
```diff
@@ -681,20 +703,36 @@ def diff_dataset_to_working_copy_feature(
         return feature_diff
 
     def all_features_diff(self, dataset, meta_diff, delta_type=Delta.insert):
```
craigds, line 705:
```suggestion
    def all_features_diff(self, table_name, schema, delta_type=Delta.insert):
```

caller ([diff_dataset_to_working_copy_feature](https://github.com/koordinates/kart/blob/9bf9e164cf16f81ea7709a9aeafa1f68a0cf1b06/kart/tabular/working_copy/base.py#L633)) will need to pass in `dataset.table_name` and a schema derived from the meta_diff (see `_get_dataset_schema`)

and you'll need to change `_execute_all_rows_query` to take a table name and a schema object too.

craigds (changes requested):

kart/tabular/working_copy/base.py
```diff
@@ -454,13 +454,43 @@ def diff_repo_to_working_copy(
...
         Returns False if the given dataset cannot be created in this working copy.
         Generally returns True since we do our best to create all datasets, even if not 100% accurately.
         """
         return True
 
+    def diff_table(self, table, meta_items):
+        """
+        Generates a diff between a working copy DB and the underlying repository tree,
+        for a single table in the given repository structure..
+        """
```
craigds, line 479: this function actually doesn't look at any existing dataset at all - it generates a diff with only a 'new' side and no old side. I'd suggest renaming it to `get_diff_for_table_creation` or something similar, and updating the docstring to match

kart/tabular/working_copy/base.py
```diff
@@ -454,13 +454,43 @@ def diff_repo_to_working_copy(
             repo_diff.prune()
             return repo_diff
 
+    def diff_table_repo_to_working_copy(self, ds_diff):
+        """
+        Generates a diff between a working copy DB and the underlying repository tree,
+        for a single dataset.
+        """
```
craigds, line 461: this no longer does what the docstring says - it just produces a RepoDiff from a single dataset diff. Probably you can remove the function and just do that in your command.

I'm not sure what `repo_diff.prune()` does - have a look and decide whether you actually need to call it or not

craigds (changes requested):

tests/test_commit.py
```diff
@@ -326,3 +330,141 @@ def test_commit_schema_violation(cli_runner, data_working_copy):
...
+                    "meta": {
+                        "inserts": 1
+                    },
+                    "feature": {
+                        "inserts": 5
+                    }
+                }
+            },
+            "commitTime": datetime.utcnow().replace(microsecond=0).isoformat() + 'Z',
+            "commitTimeOffset": "+13:00"
```
craigds, line 380: the utcnow() thing is going to be flaky if the test runs a bit slow, and (I think) the offset will vary depending on who runs the test.

Rather, hardcode a specific time/offset in an environment variable. There's some examples of this in the `commit` tests: https://github.com/koordinates/kart/blob/9bf9e164cf16f81ea7709a9aeafa1f68a0cf1b06/tests/test_commit.py#L292-L298

tests/test_commit.py
```diff
@@ -326,3 +330,141 @@ def test_commit_schema_violation(cli_runner, data_working_copy):
...
+
+        flat_diff = ""
+        for table, table_diff in diff.items():
+            flat_diff += f"  {table}:\n"
+            for section, section_diff in table_diff.items():
+                for op, count in section_diff.items():
+                    flat_diff += f"    {section}:\n"
+                    flat_diff += f"      {count} {op}\n"
+
+        date = datetime.now().strftime('%a %b %d %H:%M:%S %Y %z+1300\n')
```
craigds, line 429: same here. If you remove this timestamp stuff (by hardcoding it) could the expected output be described as a simple string instead of the 3-loop thing above?

tests/test_commit.py
```diff
@@ -326,3 +330,141 @@ def test_commit_schema_violation(cli_runner, data_working_copy):
...
+        with repo.working_copy.tabular.session() as sess:
+            sess.execute(
+                f"""CREATE TABLE IF NOT EXISTS {new_table} (test_id int primary key, field1 text, field2 text);"""
+            )
+
+        try:
+            cli_runner.invoke(["add-dataset", "test_table", "-m", message1, "-o", "text"])
+            cli_runner.invoke(["add-dataset", "test_table", "-m", message2, "-o", "text"])
+        except NotFound as e:
+            assert str(e) == f"Table wrong_test_table is either not found or already tracked by kart.\n\nTry running 'kart status --list-untracked-tables'\n"
```
craigds, line 468: could we differentiate the two cases (table doesn't exist, table already tracked by kart)? I think I'd find this error message a bit confusing as a user

craigds (changes requested):

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 - Fix the issue where re-importing the same shapefile with `--replace-existing` would show spurious schema changes (and therefore row changes). [#791](https://github.com/koordinates/kart/issues/791)
 
+- Added support for committing new layers to a kart repository. When adding a new layer to a working copy, you can now commit the layer to add it to your kart repository. Previously, you could only commit changes to existing layers that were already tracked by kart.
+To commit a new layer, use the `add-dataset` command with the name of the layer and a commit message. If the layer is not yet tracked by kart, it will be added to the repository as a new layer.
+
```
craigds, line 13: This is quite wordy - I'd suggest shortening it to a single-sentence. Mention the name of the command and what it does; that's all we need here. Add the link to this PR, like the other changelog notes have

kart/add_dataset.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    repo = ctx.obj.repo
+
+    # Check that the table is in the list of untracked tables:
+    untracked_tables = get_untracked_tables(repo)
+    if table_name not in untracked_tables:
+        # Check if the table is already tracked:
+        ds_paths = list(repo.datasets().paths())
+        for ds_path in ds_paths:
+            if ds_path == table_name:
```
craigds, line 57: `if table_name in ds_paths:`

craigds (approved):
This looks great; nice work 🎉

Issue #794, opened 2023-02-23
Add support for version-controlled rasters


olsen232:
Tracking bug - will add PRs here, and make notes of functionality that is missing. Later when rasters gets closer to launch, this can be used to file individual raster issues.

Some bullet points:

- [x] Basic import from geotiff - extract metadata from geotiff, write as Kart a dataset
- [x] Diffing: commit<>commit
- [x] Diffing: commit<>working copy
- [x] Diffs can be committed
- [x] Support for datasets that are constrained to be COG, vs datasets that are just geotiff
- [x] Support for convert-to-COG during import
- [x] Support for convert-to-COG during diff / commit
- [x] Spatial filters working

Maybe: any more conversions supported? Any relaxable constraints eg "optional alpha band"?

olsen232:
The gdal module is already included in the Kart build, which is useful - but tests are failing on CI on macos due to a crash in the gdal module which we'll need to investigate:
```
  1: tests/raster/test_imports.py::test_import_single_geotiff Fatal Python error: Aborted
  
  1: Current thread 0x000000011c82e600 (most recent call first):
  1:   File "/Users/runner/work/kart/kart/build/venv/lib/python3.10/site-packages/osgeo/gdal.py", line 4757 in InfoInternal
  1:   File "/Users/runner/work/kart/kart/build/venv/lib/python3.10/site-packages/osgeo/gdal.py", line 248 in Info
  1:   File "/Users/runner/work/kart/kart/tests/raster/fixtures.py", line 14 in requires_gdal_info
```
Not much useful info there

olsen232:
- category-tables and band descriptions are now imported / exposed as meta items, as are RAT headings
- color tables are not imported (or at least not exposed as meta items) - there's no guarantee they'll be homogenous across the dataset

Some missing bits to come back to:
- category-tables aren't merged well - if one of your tiles is missing some of the categories, it will conflict with the others. Certain tiles missing certain categories should be allowed.
- spatial filtering doesn't affect PAM files - all PAM files match the spatial filter.

olsen232:
PAM files are now editable and committable, including the meta items - category-tables, RAT headings

Some missing bits to come back to:
Changing the number of meta items changes during a diff / commit is not handled properly - eg if you delete all the PAM files and try to commit, it will crash as it tries to go from some category-tables to no category tables.

olsen232:
[...]

PR #807, opened 2023-03-20
Raster: extract more data into meta/

olsen232:
- noData value
- category labels (from the RATs in .tif.aux.xml)
- RAT column headings (from the RATs in .tif.aux.xml)

## Checklist:


kart/base_dataset.py
```diff
@@ -215,6 +215,7 @@ def get_data_at(
     def meta_tree(self):
         return self.get_subtree(self.META_PATH)
 
+    @classmethod
     def get_meta_item_definition(self, meta_item_path):
```
craigds, line 219: rename `self` to `cls`

kart/raster/metadata_util.py
```diff
@@ -1,21 +1,32 @@
...
+
+
 def rewrite_and_merge_metadata(tile_metadata_list):
     """
     Given a list of tile metadata, merges the parts we expect to be homogenous into a single piece of tile metadata in
     the same format that describes the whole list.
     """
     # TODO - this will get more complicated as we add support for convert-to-COG.
     result = {}
+    all_keys = set().union(*tile_metadata_list)
```
craigds, line 22: initially I thought this was just `set(tile_metadata_list)` but then I noticed the `*` and realised maybe `union` can take more than one argument - is this unioning the contents of sublists together?

if so, maybe make that more obvious:
```suggestion
    all_keys = set()
    for tm in tile_metadata_list:
        all_keys.update(tm)
```

craigds (approved):

kart/raster/metadata_util.py
olsen232: Done

Issue #721, opened 2022-10-19
`show`: Add a way to not show a full diff


craigds:
We'd like to be able to get structured details for a commit without producing or parsing megabytes of JSON.

I'd suggest adding a `--diff-format` option with the following semantics:

### `kart show -o json --diff-format=none <sha1>`:

No diff at all. display commit metadata only (author/message/parent commit/etc)

### `kart show -o json --diff-format=full <sha1>`

Full diff (same as no flag)

### `kart show -o json --diff-format=no-data-changes <sha1>`

Display a diff, but don't include the `feature`/`tile` tree for datasets. Replace with a `data_changes: true|false` boolean to say _if_ the `feature`/`tile` tree changed, so the caller knows whether they might need to request the full diff later.

PR #808, opened 2023-03-21
Diff format 721

fedch:
## Description

This pull request adds a new `--diff-format` option to the `kart show` command that allows users to control the level of detail in the diff output.

The `--diff-format=none` option displays only commit metadata (author, message, parent commit, etc.) without any diff information. The `--diff-format=full` option displays the full diff, which is the same as not using the `--diff-format` option. The `--diff-format=no-data-changes` option displays a diff that replaces the feature/tile diff with a data_changes boolean to indicate whether the tree has changed.

The implementation considers the tradeoff between providing detailed diff information and minimising the size of the output.

## Related links:

[show: Add a way to not show a full diff #721](https://github.com/koordinates/kart/issues/721)

## Checklist:


tests/test_show_diff_format.py
```diff
@@ -0,0 +1,42 @@
...
+        assert output["kart.diff/v1+hexwkb"] == {
+            "nz_pa_points_topo_150k": {"data_changes": True}
+        }
+
+
+def test_show_diff_format_none_text(cli_runner, data_archive):
+    # Check that the text output is a commit metadata only
+    with data_archive("points.tgz"):
+        r = cli_runner.invoke(["show", "--diff-format=none", "-o", "text"])
+        output = r.stdout
```
olsen232, line 28: Use r.stdout.splitlines() - then the expected output gets to be formatted better, instead of being on one long line with "\n" literals in it

kart/tile/tile_dataset.py
```diff
@@ -292,14 +293,38 @@ def get_tile_summary_from_filesystem_path(self, path, *, tile_metadata=None):
...
+        ds_filter=DatasetKeyFilter.MATCH_ALL,
+        reverse=False,
+        diff_format=DiffFormat.FULL,
+    ):
         """
         Generates a Diff from self -> other.
         If reverse is true, generates a diff from other -> self.
         """
+        # If the user is not asking for a tile diff, we can skip getting the diff for the feature subtree.
+        if diff_format == DiffFormat.NONE.value:
```
olsen232, line 308: I don't think this special-case is really helping you:
- if the user has requested DiffFormat.NONE, we should be able to generate the output they want without getting this far - without calling each dataset one by one and asking it to generate a diff of format DiffFormat.NONE. So in practise, this should not happen
- if this is called with DiffFormat.NONE, I think returning None is quite likely to break things, but the error would not happen here but somewhere else.

I think a better option here would be to assert that diff_format is not DiffFormat.NONE - then if this code is called with DiffFormat.NONE it will fail immediately. Calling this code with DiffFormat.NONE is not useful since it effectively asks "can I have the diff of dataset xyz but actually just return None". 

It's not necessarily bad to write a function that simply returns None in certain circumstances but I think in this case, it is much better if we make sure we don't even start the whole complicated diff code, instead of generating an "None" diff for each dataset in turn and then trying to join them all together.

kart/dataset_mixins.py
```diff
@@ -55,7 +63,9 @@ def diff_to_working_copy(
         # Subclasses to override.
         pass
 
-    def get_raw_diff_for_subtree(self, other, subtree_name, reverse=False):
+    def get_raw_diff_for_subtree(
+        self, other, subtree_name, reverse=False, data_changes_required=False
```
olsen232, line 67: would prefer `data_changes_required=True` I think
- it's the current behaviour
- it's what I would expect by default
- it's consistent with the other defaults that you set everywhere: diff_format=DiffFormat.FULL

olsen232 (commented):
I think Craig is back in charge again but I finished off some comments that I started writing

olsen232 (commented):
Looks like my comments have been addressed - Craig wanted to review this so I'll leave with him

kart/json_diff_writers.py
```diff
@@ -102,6 +105,11 @@ def default(self, obj):
                     key: self._postprocess_simple_delta(value)
                     for key, value in ds_diff["meta"].items()
                 }
+            if "data_changes" in ds_diff:
+                if ds_diff["data_changes"] is True:
+                    result["data_changes"] = True
+                else:
+                    result["data_changes"] = False
```
craigds, line 112:
```suggestion
                result["data_changes"] = ds_diff["data_changes"]
```

kart/quiet_diff_writer.py
```diff
@@ -1,8 +1,13 @@
...
+from kart.diff_format import DiffFormat
 from .base_diff_writer import BaseDiffWriter
 
 
 class QuietDiffWriter(BaseDiffWriter):
-    def write_diff(self):
+    # TODO: Pass DiffFormat.NODATACHANGES to make it more efficient while providing enough info to exit with the correct code.
+    def write_diff(self, diff_format=DiffFormat.FULL):
+        if diff_format != DiffFormat.FULL:
+            raise ClickException.UsageError("Quiet format only supports full diffs")
```
craigds, line 10: typo? ClickException.UsageError

tests/test_show_diff_format.py
```diff
@@ -0,0 +1,57 @@
...
+
+def test_show_diff_format_no_data_changes_json(cli_runner, data_archive):
+    # Check that the json output is a commit metadata (if any) and a boolean for data_changes (feature/tile changes)
+    with data_archive("points.tgz"):
+        r = cli_runner.invoke(["show", "--diff-format=no-data-changes", "-o", "json"])
+        output = json.loads(r.stdout)
+        assert output["kart.diff/v1+hexwkb"] == {
+            "nz_pa_points_topo_150k": {"data_changes": True}
+        }
+
```
craigds, line 22: I think we also need a test of a meta-only commit (ie `data_changes` will be False).

Look at `tests/data/meta-updates.tgz` - the latest commit in that repo has only meta changes

kart/base_diff_writer.py
```diff
@@ -256,9 +257,25 @@ def write_warnings_footer(self):
...
+            return
+
+        # If the diff format is NODATACHANGES, check if there is a diff and print True or False
+        elif diff_format == DiffFormat.NODATACHANGES:
+            for ds_path in self.all_ds_paths:
+                ds_diff = self.get_dataset_diff(ds_path)
+                has_changes = bool(ds_diff)
+            if has_changes:
+                click.secho(f"    Data changes: {has_changes}\n")
+            self.write_warnings_footer()
```
craigds, line 275: there's something wrong here. Things to think about:

1. what happens when there is a meta-only diff (no feature changes)? Perhaps we need a test to make sure this does the right thing in that case.
2. what happens when there are multiple datasets?
3. do any subclasses of `BaseDiffWriter` call this base `write_diff` implementation? Is the output appropriate for BaseDiffWriter, or should this code move to a subclass? (we might need to discuss this one on a call)

kart/diff_structs.py
```diff
@@ -221,7 +223,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def ensure_child_type(self, key, value):
-        if type(value) != self.child_type:
+        if not isinstance(value, self.child_type):
             raise TypeError(
                 f"{type(self).__name__} accepts children of type {self.child_type.__name__} "
```
craigds, line 228: this error message probably doesn't work correctly when `child_type` is a tuple of types

craigds (changes requested):
great progress 💯

kart/base_diff_writer.py
olsen232: (This part of the change has been reverted / reworked, so these comments have been addressed)

olsen232 (approved):

PR #809, opened 2023-03-28
Switch crs84Extent to a 4-point-polygon

olsen232:
## Description

Switch from an axis-aligned extent for the crs84Extent field for point clouds:
`crs84Extent = 174.73844833207193,174.74945404214898,-36.85123712200056,-36.84206322341377,-1.66,99.83`
to a 4 point polygon (ie, a rectangle or similar, but no longer axis-aligned):
`POLYGON((174.7384483 -36.8512371,174.7382443 -36.8422277,174.7494540 -36.8420632,174.7496594 -36.8510726,174.7384483 -36.8512371))`

Lots of tile hashes have changed due to re-import of the test datasets, which was done as the simplest way to update the tile metadata in line with the new spec - but re-importing with convert-to-copc isn't deterministic and so the tile hashes have all changed.

## Checklist:


rcoup:
> if you have a point cloud dataset that you already created, it will still have the old style value in that field. and it will stay like that forever until and unless you reimport or otherwise update the tiles in some way. In Kart terms that's compatible since Kart doesn't actually use the field for anything, it just exposes it to the user / client

PR #810, opened 2023-03-30
Start of PAM sidecar support

olsen232:
Support for PAM (Persistent Auxiliary Metadata) files aka .tif.aux.xml files:

Changes:
- PAM files are imported
- PAM files are checked out
- PAM oid / size appears in the tile diff.
- Relevant tests pass

Not yet done (to be done in subsequent changes):
- Changes to PAM files in working copy cannot be diffed / committed.
- PAM files are not properly spatial filtered

## Checklist:

### DIFF
--- a/kart/point_cloud/tilename_util.py
+++ b/kart/point_cloud/tilename_util.py
@@ -64,5 +64,5 @@ def get_tile_path_pattern(
     version_pattern = (
         r"(?:\.ancestor|\.ours|\.theirs)?" if include_conflict_versions else ""
     )
-    ext_pattern = r"(?i:\.copc)?(?i:\.la[sz])"
+    ext_pattern = r"(?i:\.copc\.laz|\.la[sz])"
     return re.compile(parent_pattern + tile_pattern + version_pattern + ext_pattern)
--- a/kart/point_cloud/tilename_util.py
+++ b/kart/point_cloud/tilename_util.py
@@ -64,5 +64,5 @@ def get_tile_path_pattern(
     version_pattern = (
         r"(?:\.ancestor|\.ours|\.theirs)?" if include_conflict_versions else ""
     )
-    ext_pattern = r"(?:\.[Cc][Oo][Pp][Cc])?\.[Ll][Aa][SsZz]"
+    ext_pattern = r"(?i:\.copc)?(?i:\.la[sz])"
     return re.compile(parent_pattern + tile_pattern + version_pattern + ext_pattern)
--- a/kart/raster/import_.py
+++ b/kart/raster/import_.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 import logging
 
 import click
@@ -163,3 +164,10 @@ class RasterImporter(TileImporter):
             source_oid = "sha256:" + source_oid
 
         return existing_summary.get("oid") == source_oid
+
+    def sidecar_files(self, source):
+        source = str(source)
+        if self.DATASET_CLASS.remove_tile_extension(source) != source:
+            sidecar_path = source + ".aux.xml"
+            if Path(sidecar_path).is_file():
+                yield sidecar_path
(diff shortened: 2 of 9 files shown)


kart/raster/import_.py
```diff
@@ -163,3 +164,10 @@ def existing_tile_matches_source(self, source_oid, existing_summary):
             source_oid = "sha256:" + source_oid
 
         return existing_summary.get("oid") == source_oid
+
+    def sidecar_files(self, source):
+        source = str(source)
+        if self.DATASET_CLASS.remove_tile_extension(source) != source:
+            sidecar_path = source + ".aux.xml"
```
rcoup, line 171: If GDAL has a tile `FOO.TIF` (uppercase), is the sidecar file `FOO.TIF.AUX.XML`, `FOO.TIF.aux.xml`, `FOO.tif.aux.xml`, or something else? Does GDAL care? Do we?

rcoup (commented):
PAM files are LFS/pointer-file based, correct? What does a pointer file for PAM metadata look like?

kart/raster/import_.py
olsen232: Good point: GDAL is completely indifferent to case when it looks for sidecars. Updated Kart to behave the same way.

kart/tile/tilename_util.py
```diff
@@ -20,3 +23,24 @@ def remove_any_tile_extension(filename):
         if len(filename) != orig_len:
             return filename
     return filename
+
+
+def case_insensitive(name):
+    """
+    Given a name eg "baz", returns a case-insensitive version that can be understood by pathlib.Path.glob
+    or the fnmatch module - eg "[Bb][Aa][Zz]"
```
craigds, line 31: minor: fnmatch is actually case insensitive by default, so I think it makes more sense to just reference `glob` here since its why this function is actually needed

kart/tile/tile_dataset.py
```diff
@@ -299,21 +299,23 @@ def diff(self, other, ds_filter=DatasetKeyFilter.MATCH_ALL, reverse=False):
         """
         ds_diff = super().diff(other, ds_filter=ds_filter, reverse=reverse)
         tile_filter = ds_filter.get("tile", ds_filter.child_type())
-        ds_diff["tile"] = DeltaDiff(self.diff_tile(other, tile_filter, reverse=reverse))
+        ds_diff["tile"] = self.diff_tile(other, tile_filter, reverse=reverse)
         return ds_diff
 
     def diff_tile(self, other, tile_filter=FeatureKeyFilter.MATCH_ALL, reverse=False):
         """
         Yields tile deltas from self -> other, but only for tile that match the tile_filter.
```
craigds, line 307: docstring seems wrong now

kart/raster/v1.py
```diff
@@ -43,6 +45,43 @@ def get_tile_path_pattern(
             include_conflict_versions=include_conflict_versions,
         )
 
+    def diff_tile(self, other, tile_filter=FeatureKeyFilter.MATCH_ALL, reverse=False):
+        """
+        Yields tile deltas from self -> other, but only for tile that match the tile_filter.
```
craigds, line 50: docstring seems wrong

craigds (approved):

kart/tile/tilename_util.py
olsen232: Done

kart/point_cloud/tilename_util.py
```diff
@@ -64,5 +64,5 @@ def get_tile_path_pattern(
     version_pattern = (
         r"(?:\.ancestor|\.ours|\.theirs)?" if include_conflict_versions else ""
     )
-    ext_pattern = r"(?:\.[Cc][Oo][Pp][Cc])?\.[Ll][Aa][SsZz]"
+    ext_pattern = r"(?i:\.copc)?(?i:\.la[sz])"
```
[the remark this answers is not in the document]
olsen232: Tightened up regex to only accept .copc.laz, .las or .laz

kart/tile/tile_dataset.py
olsen232: Fixed

kart/raster/v1.py
olsen232: Fixed

PR #811, opened 2023-04-03
Fix PDAL SSL cert handling in linux

craigds:
## Description

When kart runs its bundled `pdal` on linux (`/opt/kart/pdal`) it seems unable to handle S3 URLs, as it can't find the certificate bundle file.

output with `VERBOSE=1` in environment shows this:

```
* Connection #0 to host 169.254.169.254 left intact
*   Trying 3.5.166.161:443...
* Connected to {BUCKETNAME}.s3-ap-southeast-2.amazonaws.com (3.5.166.161) port 443 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* error setting certificate file: /etc/pki/tls/certs/ca-bundle.crt
* Closing connection 0
```

`/etc/pki/tls/certs/ca-bundle.crt` doesn't exist and isn't the cert bundle that Kart ships with.

This PR sets `CURL_CA_INFO` to the correct path (`/opt/kart/certifi/cacert.pem`) and this fixes the issue with PDAL.

## Checklist:

### DIFF
--- a/kart/__init__.py
+++ b/kart/__init__.py
@@ -66,6 +66,13 @@ path_extras = [prefix]
 if not is_frozen:
     path_extras += [os.path.join(prefix, "scripts"), os.path.join(prefix, "lib")]
 
+if is_linux:
+    import certifi
+
+    os.environ["SSL_CERT_FILE"] = certifi.where()
+    # affects PDAL's libcurl, and possibly other libcurls
+    os.environ["CURL_CA_INFO"] = certifi.where()
+
 # Git
 # https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables
 os.environ["GIT_CONFIG_NOSYSTEM"] = "1"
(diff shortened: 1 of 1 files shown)


rcoup:
Do we need it on macOS and/or Windows too? I _suspect_ not, since I think curl falls back more consistently to the system certificate stores.

rcoup (approved):

craigds:
No idea tbh.

craigds:
https://curl.se/libcurl/c/CURLOPT_CAINFO.html

> (iOS and macOS) When curl uses Secure Transport this option is supported. If the option is not set, then curl will use the certificates in the system and user Keychain to verify the peer.
> 
> (Schannel) This option is supported for Schannel in Windows 7 or later but we recommend not using it until Windows 8 since it works better starting then. If the option is not set, then curl will use the certificates in the Windows' store of root certificates (the default for Schannel). 

sounds likely to work by default on those OSes as you say

PR #813, opened 2023-04-04
Commit raster tiles and PAM files

olsen232:
- Diffs to raster tiles can now be committed
- Diffs to PAM files can now be committed

Attempts to deal with the possibility that the
PAM file's case doesn't match up with the raster tile, at least at commit time - tries to normalise everything as follows:

- the raster tilename can be anything (upper or lowercase)
- internally, the raster tile extension is not stored
- externally, the raster tile extension is ".tif" in the WC (we could make this configurable to .tiff if it would help anyone)
- internally, the PAM filename is the same as the tilename, + ".aux.xml"
- externally the WC PAM name is the same as the tile + ".aux.xml" in the WC

but, we have to allow for deviations from this at commit-time and fix them up during the commit.

Also: renamed meta/bands to meta/band - we don't use plurals for any of the other dataset trees.
Some adjustment to how pam deltas are generated from previous PR.


kart/raster/v1.py
```diff
@@ -163,3 +195,135 @@ def diff_to_working_copy(
...
+        """
+        Given a list of dirty paths, relative to the workdir -
+        return the paths of the tiles (relative to the workdir) that have been affected
+        either by editing directly or by editing their PAM files.
+        """
+        result = set()
+        for path in tile_and_pam_paths:
+            if path.lower().endswith(".aux.xml"):
+                tile_paths = find_similar_files_case_insensitive(
+                    self._workdir_path(path[:-8])
```
craigds, line 209: magic 8 is going to bite us when we inevitably add another type of sidecar file. maybe use `len(PAM_SUFFIX)` and then at least it's greppable

kart/raster/v1.py
```diff
@@ -163,3 +195,135 @@ def diff_to_working_copy(
...
+        """The total number of tiles in this dataset, not including PAM files."""
+        if self.inner_tree is not None:
+            try:
+                subtree = self.inner_tree / self.TILE_PATH
+                return sum(
+                    1
+                    for blob in all_blobs_in_tree(subtree)
+                    if not blob.name.endswith(".aux.xml")
+                )
+            except KeyError:
```
craigds, line 238: what's this catching exactly?

kart/raster/v1.py
```diff
@@ -163,3 +195,135 @@ def diff_to_working_copy(
...
+                    delta.new_value.update(
+                        oid=pointer_dict["oid"], size=pointer_dict["size"]
+                    )
+
+                    pam_name, pam_source_name, pam_oid, pam_size = (
+                        delta.new_value.get("pamName"),
+                        delta.new_value.get("pamSourceName"),
+                        delta.new_value.get("pamOid"),
+                        delta.new_value.get("pamSize"),
+                    )
```
craigds, line 310: just have four assignments :)

kart/raster/v1.py
olsen232: Done

kart/raster/v1.py
olsen232: No such subtree at `inner_tree / "tile"`
I've rearranged it so that the KeyError is only caught where its useful, instead of over the whole function

kart/raster/v1.py
olsen232: Done

kart/workdir.py
```diff
@@ -606,6 +615,22 @@ def _update_dataset_in_workdir(
...
 
+                pam_name = tile_delta.new_value.get("pamName")
+                if pam_name:
+                    pam_path = ds_tiles_dir / pam_name
+                    lfs_path = get_local_path_from_lfs_hash(
+                        self.repo, tile_delta.new_value["pamOid"]
+                    )
+                    if not lfs_path.is_file():
+                        click.echo(
+                            f"Couldn't find PAM file {tilename} locally - skipping...",
```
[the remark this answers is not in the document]
olsen232: Done

Issue #806, opened 2023-03-16
add-dataset: Throws an exception if table has no primary key


craigds:
The new `add-dataset` command doesn't handle adding datasets that have no primary key field.
```
sqlite> create table newtable (a integer);
```

```
$ kart add-dataset newtable
Traceback (most recent call last):
  File "/Users/cdestigter/code/kart/build/venv/bin/kart", line 8, in <module>
    sys.exit(entrypoint())
  File "/Users/cdestigter/code/kart/kart/cli.py", line 406, in entrypoint
    cli()
  File "/Users/cdestigter/code/kart/build/venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/cdestigter/code/kart/build/venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/cdestigter/code/kart/kart/cli_util.py", line 72, in invoke
    return super().invoke(ctx)
  File "/Users/cdestigter/code/kart/build/venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/cdestigter/code/kart/build/venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/cdestigter/code/kart/build/venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/cdestigter/code/kart/build/venv/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/cdestigter/code/kart/kart/add_dataset.py", line 68, in add_dataset
    table_diff = repo.working_copy.tabular.get_diff_for_table_creation(
  File "/Users/cdestigter/code/kart/kart/tabular/working_copy/base.py", line 477, in get_diff_for_table_creation
    feature_diff = self.all_features_diff(
  File "/Users/cdestigter/code/kart/kart/tabular/working_copy/base.py", line 722, in all_features_diff
    pk_field = schema.pk_columns[0].name
IndexError: tuple index out of range
```

PR #815, opened 2023-04-11
Added a check for a pk in a table when using add-dataset command

fedch:
## Description

The `add-dataset` command was breaking when a table didn't have a pk. I added a try-catch block to ensure this is handled properly.

## Related links:

[806](https://github.com/koordinates/kart/issues/806)

## Checklist:

### DIFF
--- a/kart/tabular/working_copy/base.py
+++ b/kart/tabular/working_copy/base.py
@@ -719,7 +719,15 @@ class TableWorkingCopy(WorkingCopyPart):
         Returns a diff containing all the features in the working copy.
         """
         assert delta_type in (Delta.insert, Delta.delete)
-        pk_field = schema.pk_columns[0].name
+
+        # TODO - support altering the user-provided table to add the PK column
+        # Check that the table has a primary key:
+        try:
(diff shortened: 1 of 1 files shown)


kart/tabular/working_copy/base.py
```diff
@@ -719,7 +719,16 @@ def all_features_diff(self, table, schema, delta_type=Delta.insert):
...
         """
         assert delta_type in (Delta.insert, Delta.delete)
-        pk_field = schema.pk_columns[0].name
+
+        # TODO - support altering the user-provided table to add the PK column
+        # Check that the table has a primary key:
+        try:
+            pk_field = schema.pk_columns[0].name
+        except IndexError:
+            print(
```
olsen232, line 728: Just put this message inside the NotYetImplemented exception:
- this message explains everything you need to know, so no need to have the other message that you currently have in the exception below
- putting the whole message inside the exception makes things simple and avoids worrying about things like whether to print to stdout or stderr

olsen232:
LGTM

Issue #805, opened 2023-03-16
add-dataset: Tracking tables are not created


craigds:
#801 added a new `add-dataset` command to commit tables from the working copy.

There was an oversight: the command doesn't actually create the tracking triggers for the table. So future changes to the table never show up in diff/status commands, and can't be committed.

See `_create_triggers` in the various workingcopy classes. We will want to call this function to create the triggers (perhaps it needs to be renamed if we need to call it from outside the workingcopy classes)

Transactions: We should create the triggers in the same database transaction that we use to read the diff while committing. This will prevent us from missing any rows while committing - if a row is added/changed after the transaction that reads the diff, it will trigger the tracking code and therefore show up in diffs after the commit.

PR #816, opened 2023-04-11
Added a test to check changes made to the table after creating it and committing it

fedch:
## Description

Added triggers to add-dataset command for tracking changes

Add-dataset command lacked proper tracking triggers for tables, causing changes to be undetected.

Fixed by ensuring the _create_triggers function is called during the commit process within the same database transaction as reading the diff, preventing missed rows during commit.

## Related links:

[805](https://github.com/koordinates/kart/issues/805)

## Checklist:

### DIFF
--- a/kart/add_dataset.py
+++ b/kart/add_dataset.py
@@ -1,5 +1,6 @@
 import click
 import sys
+
 from .cli_util import StringFromFile
 from .commit import (
     commit_obj_to_json,
@@ -11,6 +12,7 @@ from .output_util import dump_json_output
 from .working_copy import WorkingCopyPart
 from .exceptions import NO_CHANGES, NotFound, InvalidOperation
 from .status import get_untracked_tables
+from kart.tabular.v3 import TableV3
 
 
 @click.command()
@@ -46,7 +48,9 @@ def add_dataset(ctx, table_name, message, launch_editor, output_format):
 
     To check the new untracked tables, run 'kart status --list-untracked-tables'
     """
+
     repo = ctx.obj.repo
+    wc = repo.working_copy.tabular
 
     # Check that the table is in the list of untracked tables:
     untracked_tables = get_untracked_tables(repo)
@@ -65,6 +69,15 @@ def add_dataset(ctx, table_name, message, launch_editor, output_format):
         )
 
     meta_items = repo.working_copy.tabular.meta_items(table_name)
+
+    # Create a table object from the table_name:
+    table = TableV3.new_dataset_for_writing(table_name, meta_items["schema.json"], repo)
+
+    # Add triggers on a new table if they aren't already present:
+    with wc.session() as sess:
+        wc.create_triggers(sess, table)
+
+    # Add the table to the working copy:
     table_diff = repo.working_copy.tabular.get_diff_for_table_creation(
         table_name, meta_items
     )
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 - Added support for `--diff-format` option under `kart show` command to display different level of detail depending on the user's needs.  [#721](https://github.com/koordinates/kart/issues/721)
 
+- Resolved issue [#805](https://github.com/koordinates/kart/issues/805) by ensuring the `add-dataset` command creates tracking triggers for committed tables, enabling future changes to be detected and committed.
```
craigds, line 16: add-dataset is a new feature in this release (L10) so we don't need to mention this - the whole feature is new

kart/tabular/working_copy/mysql.py
```diff
@@ -160,6 +160,10 @@ def _sno_tracking_name(self, trigger_type, dataset=None):
         """Returns the sno-branded name of the trigger reponsible for populating the sno_track table."""
         return f"_sno_track_{trigger_type}"
 
+    def create_triggers(self, sess, dataset):
```
craigds, line 163: I'd suggest just renaming the original method, no point having two

kart/add_dataset.py
```diff
@@ -69,6 +73,14 @@ def add_dataset(ctx, table_name, message, launch_editor, output_format):
         table_name, meta_items
     )
 
+    # Create a table object from the table_name:
+    path = TableV3.dataset_path_to_table_name(table_name)
```
craigds, line 77: I'm not sure what this line adds - table_name is already a table name so it seems like this line converts it the wrong way and basically isn't required?

kart/add_dataset.py
```diff
@@ -69,6 +73,14 @@ def add_dataset(ctx, table_name, message, launch_editor, output_format):
...
+    table = TableV3.new_dataset_for_writing(path, meta_items["schema.json"], repo)
+
+    # Add triggers on a new table if they aren't already present:
+    with wc.session() as sess:
```
craigds, line 81: there's a race condition we should think about:

* user runs `kart add-dataset` on a table with 10 rows
* we build a diff
* user inserts a new row
* then we create triggers
* then we commit 10 rows to the dataset

This sequence of events causes the triggers to miss an insert - the row is not tracked by the tracking table, nor is it committed. User won't see a diff until they change the row and it gets picked up by the trigger again (it will be considered an update rather than an insert, even though the row hasn't been committed)

What if instead, we created the triggers before we fetch the diff?

* user runs `kart add-dataset` on a table with 10 rows
* we create triggers
* user inserts a new row
* trigger tracks a new inserted row
* we fetch a diff of the whole table
* we commit the diff

In this situation^ we have a different issue - the row is _both_ committed and in the tracking table - in both cases as an 'insert'. I'm not sure if this causes problems or not. I'd hope it doesn't show up in `kart diff` but I might be mistaken.


@olsen232 is this something we should address?

olsen232: Having unneeded primary-key-values in the tracking table isn't that bad - they don't cause the diffs to contain any extra deltas, they just slightly add to Kart's workload (in the extreme case, we could put every row in the tracking table for no reason, kart would still work fine, but it would lose the efficiency that the tracking table provides). So, creating triggers early is preferable to creating them late.

Apart from that preference, I wouldn't put too much more thought into it... all of this probably happens inside a single transaction, so, I don't know if reordering the pieces of it will make any difference, or if it all appears atomic to another editor. We could test it further but it seems like a lot of work for a weird edge case

craigds (approved):

PR #818, opened 2023-04-13
Add multi-worker support for tile import.

olsen232:
Affects both point cloud and raster tile import.
Doesn't affect diffing or committing..

Resurrected existing (deprecated) option --num-processes, but made it an alias for the preferred name: --num-workers.

## Checklist:

- [ ] Have you included test(s)? Existing tests pass

### DIFF
--- a/kart/tile/importer.py
+++ b/kart/tile/importer.py
@@ -340,6 +340,11 @@ class TileImporter:
         return self.DATASET_CLASS.extract_tile_metadata_from_filesystem_path(tile_path)
 
     def extract_multiple_tiles_metadata(self, sources):
+        """
+        Like extract_tile_metadata, but works for a list of several tiles. The metadata may
+        be extracted serially or with a thread-pool, depending on the value of self.num_workers.
+        Yields a tuple (source, metadata) for each tile in turn, in some unspecified order.
+        """
         # Single-threaded variant - uses the calling thread.
         if self.num_workers == 1:
             for source in sources:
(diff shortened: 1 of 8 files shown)


kart/tile/importer.py
```diff
@@ -541,3 +608,22 @@ def missing_parameter(self, param_name):
...
+            num_workers = 1
+        elif num_workers > self.MAX_WORKERS:
+            raise click.UsageError(
+                f"Importing with more than {self.MAX_WORKERS} workers is not supported"
+            )
+        return num_workers
+
+    def get_default_num_workers(self):
+        num_workers = get_num_available_cores()
+        click.echo(num_workers)
```
craigds, line 627:
```suggestion
```

kart/tile/importer.py
```diff
@@ -541,3 +608,22 @@ def missing_parameter(self, param_name):
...
+    MAX_WORKERS = 64
+
+    def check_num_workers(self, num_workers):
+        if num_workers is None:
+            num_workers = self.get_default_num_workers()
+        if num_workers < 1:
+            num_workers = 1
+        elif num_workers > self.MAX_WORKERS:
+            raise click.UsageError(
+                f"Importing with more than {self.MAX_WORKERS} workers is not supported"
```
craigds, line 621: why? this was here for the old v2 tabular import code because each process handled one dest subtree, but it doesn't seem like it makes sense for tile-based imports

kart/tile/importer.py
```diff
@@ -524,6 +572,25 @@ def import_tiles_to_stream(self, stream, sources):
 
                 p.update(1)
 
+    def copy_multiple_files_to_lfs_cache(self, copy_and_convert_tasks):
```
craigds, line 575: needs a docstring

craigds (approved):

kart/tile/importer.py
olsen232: Removed arbitrary MAX_WORKERS limit

kart/tile/importer.py
olsen232: Fixed

PR #823, opened 2023-04-19
Support convert-to-cog during import

olsen232:
And make it the default. Doesn't yet work during diffs / commits.

Updated format.json in line with point clouds - here is a PC format.json: {
  "compression": "laz",
  "lasVersion": "1.4",
  "optimization": "copc",
  "optimizationVersion": "1.0",
  "pointDataRecordFormat": 7,
  "pointDataRecordLength": 36
}

And here is the current raster format.json:
{
  "fileType": "image/tiff; application=geotiff; profile=cloud-optimized"
}

Here is the new raster format.json:
{
  "fileType": "image/tiff; application=geotiff",
  "profile": "cloud-optimized"
}

Consistent with the PC format.json, each of these items represents a constraint:
1. the fileType must be geotiff
2. the profile must be "cloud-optimized" One of these constraints can be removed to make a dataset which allows geotiffs of any profile, by omitting that item from the JSON.

This meant updating the test data again.

## Checklist:


kart/raster/metadata_util.py
```diff
@@ -18,12 +19,24 @@
 CATEGORIES_PATTERN = re.compile(r"band/band-(.*)-categories\.json")
 
 
-def rewrite_and_merge_metadata(tile_metadata_list):
+class RewriteMetadata(IntEnum):
```
craigds, line 22: since you're using these with bitwise operators maybe it should be an `IntFlag`

craigds (approved):

kart/raster/metadata_util.py
olsen232: Done

PR #824, opened 2023-04-20
Add support for convert-to-cog during diff / commit

olsen232:
Just like point-clouds, uses --convert-to-dataset-format to decide whether to show diffs
(roughly) as they would be if the file was converted during commit to match the dataset,
and this flag is required during commit to go ahead and commit the change (otherwise,
of course, the tile doesn't match the dataset).

TODO - slightly related - make sure all the names are properly
 standardized during commit eg

foo.TIFF and foo.tif.AUX.XML should become foo.tif and foo.tif.aux.xml, this should actually rename the WC files as they are committed.

TODO - refactor some very similar raster / point-cloud code to avoid repetition.

## Checklist:

### DIFF
--- a/kart/raster/v1.py
+++ b/kart/raster/v1.py
@@ -2,7 +2,7 @@ import functools
 import re
 
 from kart.core import all_blobs_in_tree
-from kart.diff_structs import DatasetDiff, DeltaDiff, Delta, KeyValue, WORKING_COPY_EDIT
+from kart.diff_structs import DatasetDiff, DeltaDiff, Delta, WORKING_COPY_EDIT
 from kart.key_filters import DatasetKeyFilter, FeatureKeyFilter
 from kart.lfs_util import (
     copy_file_to_local_lfs_cache,
@@ -10,9 +10,12 @@ from kart.lfs_util import (
     merge_pointer_file_dicts,
 )
 from kart.meta_items import MetaItemDefinition, MetaItemFileType
+from kart.raster.gdal_convert import convert_tile_to_format
 from kart.raster.metadata_util import (
+    RewriteMetadata,
     extract_raster_tile_metadata,
     rewrite_and_merge_metadata,
+    get_format_summary,
 )
 from kart.raster.tilename_util import (
     get_tile_path_pattern,
@@ -125,14 +128,13 @@ class RasterV1(TileDataset):
             are missing almost all of the info about the new tiles, but this is faster and more
             reliable if this information is not needed.
         """
-        if convert_to_dataset_format:
-            raise NotImplementedError(
-                "Sorry, convert_to_dataset_format is not yet implemented for raster datasets"
-            )
-
         tile_filter = ds_filter.get("tile", ds_filter.child_type())
 
         current_metadata = self.tile_metadata
+        dataset_format_to_apply = None
+        if convert_to_dataset_format:
+            dataset_format_to_apply = get_format_summary(current_metadata)
+
         tilename_to_metadata = {}
 
         wc_tiles_path_pattern = get_tile_path_pattern(parent_path=self.path)
@@ -160,6 +162,13 @@ class RasterV1(TileDataset):
                 tilename_to_metadata[wc_path.name] = tile_metadata
                 new_tile_summary = tile_metadata["tile"]
 
+                if dataset_format_to_apply and not self.is_tile_compatible(
+                    dataset_format_to_apply, new_tile_summary
+                ):
+                    new_tile_summary = self.pre_conversion_tile_summary(
+                        dataset_format_to_apply, new_tile_summary
+                    )
+
                 new_half_delta = tilename, new_tile_summary
             else:
                 new_half_delta = tilename, {"name": wc_path.name}
@@ -178,10 +187,23 @@ class RasterV1(TileDataset):
         if not is_clean_slate:
             metadata_list.insert(0, current_metadata)
 
+        rewrite_metadata = RewriteMetadata.NO_REWRITE
+        profile_constraint = current_metadata["format.json"].get("profile")
+        if profile_constraint == "cloud-optimized":
+            rewrite_metadata = (
+                RewriteMetadata.AS_IF_CONVERTED_TO_COG
+                if convert_to_dataset_format
+                else RewriteMetadata.NO_REWRITE
+            )
+        else:
+            rewrite_metadata = RewriteMetadata.DROP_PROFILE
+
         if no_new_metadata:
             merged_metadata = current_metadata
         else:
-            merged_metadata = rewrite_and_merge_metadata(metadata_list)
+            merged_metadata = rewrite_and_merge_metadata(
+                metadata_list, rewrite_metadata
+            )
 
         meta_diff = DeltaDiff()
         all_keys = set()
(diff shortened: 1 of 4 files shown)


kart/raster/v1.py
```diff
@@ -276,6 +298,35 @@ def get_tile_summary_from_pointer_blob(
             result["pamSize"] = pam_summary["size"]
         return result
 
+    def is_tile_compatible(self, ds_format, tile_summary):
```
craigds, line 301: Add a docstring

kart/raster/v1.py
```diff
@@ -276,6 +298,35 @@ def get_tile_summary_from_pointer_blob(
...
+        tile_format = tile_summary["format"]
+        if isinstance(ds_format, dict):
```
craigds, line 303: when might this not be a dict?

kart/raster/v1.py
```diff
@@ -276,6 +298,35 @@ def get_tile_summary_from_pointer_blob(
...
+            if envisioned_summary.get(key):
+                result[key] = envisioned_summary[key]
+            if key in envisioned_summary:
+                result["source" + key[0].upper() + key[1:]] = value
+            else:
+                result[key] = value
+        return result
+
     def apply_tile_diff(
         self, tile_diff, object_builder, *, resolve_missing_values_from_ds=None
```
craigds, line 331: add a docstring

kart/raster/v1.py
```diff
@@ -284,33 +335,44 @@ def apply_tile_diff(
...
+                            self.repo, path_in_wc, conversion_func
+                        )
+                        pointer_dict = merge_pointer_file_dicts(new_val, pointer_dict)
+                    else:
+                        # Committing in a new tile, preserving its format
+                        oid_and_size = new_val["oid"], new_val["size"]
+                        pointer_dict = copy_file_to_local_lfs_cache(
+                            self.repo, path_in_wc, oid_and_size=oid_and_size
+                        )
+                        pointer_dict = merge_pointer_file_dicts(new_val, pointer_dict)
```
craigds, line 362:
```suggestion
                    else:
                        # Committing in a new tile, preserving its format
                        oid_and_size = new_val["oid"], new_val["size"]
                        pointer_dict = copy_file_to_local_lfs_cache(
                            self.repo, path_in_wc, oid_and_size=oid_and_size
                        )
                    pointer_dict = merge_pointer_file_dicts(new_val, pointer_dict)
```

tests/raster/test_workingcopy.py
```diff
@@ -527,3 +528,110 @@ def _set_category_labels(pam_path, category_labels):
...
+            assert r.stdout.splitlines() == [
+                "On branch main",
+                "",
+                "Changes in working copy:",
+                '  (use "kart commit" to commit)',
+                '  (use "kart restore" to discard changes)',
+                "",
+                "  aerial:",
+                "    meta:",
+                "      1 updates",
```
craigds, line 559: why's there an update?

tests/raster/test_workingcopy.py
```diff
@@ -527,3 +528,110 @@ def _set_category_labels(pam_path, category_labels):
...
+            r = cli_runner.invoke(["diff"])
+            assert r.exit_code == 0, r.stderr
+            assert r.stdout.splitlines() == [
+                "--- aerial:meta:format.json",
+                "+++ aerial:meta:format.json",
+                "- {",
+                '-   "fileType": "geotiff",',
+                '-   "profile": "cloud-optimized"',
+                "- }",
+                "+ <<<<<<< ",
```
craigds, line 573: I don't understand why there's a conflict here?

kart/raster/v1.py
olsen232: Done

kart/raster/v1.py
olsen232: ah... maybe never, but there's a few methods like this that need to be given something like "target format". Depending on where they were used, it was sometimes more useful to have them accept the "format.json" meta item, or sometimes the little format-summary that we store per tile eg "geotiff/cog" or "laz-1.4/copc-1.0" but after a while I tried to make them accepting of either one, where possible.

kart/raster/v1.py
olsen232: Done

kart/raster/v1.py
olsen232: Done

tests/raster/test_workingcopy.py
olsen232: Hmm. Basically because we are going from
```
- {
-   "fileType": "geotiff",
-   "profile": "cloud-optimized"
- }
```
to... something else - more on this below in next comment. This is not special-cased in any way, so it's an update.
Maybe we should call it something else... `kart status` now has `inserts` `updates` `deletes` and `spatialFilterConflicts` - this could be a `formatConflict` or `formatViolation` or something

tests/raster/test_workingcopy.py
olsen232: This too may only make sense as part of a broader check that is not special cased.
Let's say your tiles have a band-schema of 32-bit RGBA.
If you insert more tiles with the 32-bit RGBA, you may commit.
If you delete all tiles and replace them with tiles with 24-bit RGB, you may also commit. This now counts as a meta-update - the schema.json meta item will change.
If you insert one tile that has 24-bit RGB, you may not commit. This shows up as a tile insert, and as a meta-item update - the old value is 32-bit RGBA, and the new value is a ListOfConflicts containing two values: (32-bit RGBA, 24-bit RGB).

format.json uses much same code as schema.json. So, right now you have a format.json saying geotiff, cloud-optimized.
If you insert more COGs, you may commit.
If you delete all tiles and replace with non-COG TIFFs, you may commit.
If you insert a single tile that is non-COG, you may not commit. This shows up as a tile insert, and as a meta-item update - the old value is geotiff+cloud-optimized, and the new value is a ListOfConflicts containing two values: (geotiff+cloud-optimized, and vanilla geotiff). That's what you can see here.

There was an idea at some point to have some kind of `--update-dataset-to-match-new-tiles` flag in addition to the opposite flag that we do have `--convert-to-dataset-format` but - we don't have one so far. (You could do this manually using `kart meta set`) So, if you get a conflict like this, where the new tile format doesn't match the old one, the best option is probably `--convert-to-dataset-format`. Maybe we should add a helpful hint to the diff letting the user know about this flag.

olsen232:
You raise some philosophical questions in this PR - everything works in a way that makes sense to me - the author - who has an all-purpose model in my head that works not just for COG and non COG, but also for COPC and non COPC, for different LAZ versions, and for changes in schema. I think the model is probably mostly fine but maybe it needs some tweaks as to how it's displayed - if it's not clear to you, a Kart expert, what's conflicting with what, then, we should at least add a help section in the kartproject docs and a link to that whenever one of these kind of "conflicts" occurs.

My preference would be to get this merged anyway since it's in review already, and all it does is make `--convert-to-dataset-format` work the same for rasters as it does for PC - but open a new issue for improving understandability of these kinds of conflicts.

tests/raster/test_workingcopy.py
craigds: yeah, that makes sense. However, I think we're going to hit this quite a bit:

* `kart import` _by default_ converts your files to COPC/COG, so most new tiled datasets will be COPC/COG
* Adding new tiles via committing _by default_ does not convert to the dataset format

Since users will inevitably use the same processing pipeline for the stuff they import and commit, I feel like almost all datasets that ever get updated will hit this conflict.

Should `--convert-to-dataset-format` just be the default?

craigds (approved):
This looks good 👍 although we need to discuss those open questions more for future development

Issue #728, opened 2022-11-07
Unhandled exception when importing from a Postgres DB which doesn't have the postgis extension


pfw:
**Describe the bug**

`kart import` dumps a stack trace rather than showing an error when importing from a Postgres DB which hasn't had `CREATE EXTENSION...' run.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new PG database
2. Import some data eg. `ogr2ogr -f PostgreSQL PG:dbname=data_gov_au mdb-boundaries.geojson`
3. In an existing Kart repo try to import that table `kart import postgresql://pfw:@localhost/data_gov_au mdb_boundaries`

If between step 1 and 2 `create EXTENSION postgis;` is run then it will complete correctly.

**Expected behaviour**
~Kart should probably show an error on Postgis missing~

Kart show allow importing (non-spatial) tables from PostgreSQL installs that don't have postGIS installed.

**Output**
```
Traceback (most recent call last):
  File "sqlalchemy/engine/base.py", line 1901, in _execute_context
  File "sqlalchemy/engine/default.py", line 732, in do_execute
psycopg2.errors.UndefinedFunction: function postgis_typmod_type(integer) does not exist
LINE 6:                 upper(postgis_typmod_type(A.atttypmod)) AS g...
                              ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "kart_cli.py", line 4, in <module>
  File "kart/cli.py", line 392, in entrypoint
  File "click/core.py", line 1130, in __call__
  File "click/core.py", line 1055, in main
  File "kart/cli_util.py", line 68, in invoke
  File "click/core.py", line 1657, in invoke
  File "click/core.py", line 1404, in invoke
  File "click/core.py", line 760, in invoke
  File "click/decorators.py", line 26, in new_func
  File "kart/tabular/import_.py", line 322, in import_
  File "kart/tabular/import_source.py", line 57, in check_valid
  File "kart/list_of_conflicts.py", line 91, in check_sources_are_importable
  File "kart/tabular/sqlalchemy_import_source.py", line 229, in meta_items
  File "kart/tabular/sqlalchemy_import_source.py", line 237, in meta_items_from_db
  File "kart/sqlalchemy/adapter/base.py", line 155, in all_v2_meta_items
  File "kart/utils.py", line 22, in wrapper
[...]

PR #828, opened 2023-04-26
Importing non-spatial tables from postgres dbs wich don't have postgis installed

fedch:
## Description

Kart should allow importing (non-spatial) tables from PostgreSQL installs that don't have postGIS installed.

## Related links:

[Issue 728](https://github.com/koordinates/kart/issues/728)

## Checklist:


tests/test_working_copy_postgis.py
```diff
@@ -801,3 +801,67 @@ def test_checkout_custom_crs(
                 "    },",
                 "  ]",
             ]
+
+
+def test_import_from_non_postgis_db(cli_runner, tmp_path, chdir):
```
craigds, line 806: should rename this file (`postgis` -> `postgres`) or put this test in its own file

tests/conftest.py
```diff
@@ -990,7 +989,8 @@ def postgis_db():
         try:
             conn.execute("SELECT postgis_version();")
         except sqlalchemy.exc.DBAPIError:
-            raise pytest.skip("Requires PostGIS")
+            pass
+            # raise pytest.skip("Requires PostGIS")
```
craigds, line 993: This fixture's called `postgis_db`, so I think marking a test with it should ensure we have postgis available. Probably we need to split this into two fixtures called (`postgres_db`, `postgis_db`)

`postgis_db` will require `postgres_db`, and additionally check that postgis is installed. Your new test should use `postgres_db` rather than hardcoding db settings itself

olsen232: FYI what you see here is the hack I helped Elena write that allows her to develop the test `test_import_from_non_postgis_db` but I didn't settle on any particular plan going forward - if we need to env flags eg KART_POSTGRES_URL and KART_POSTGIS_URL pointing to two different PG instances, if we need to spin up both during CI. That was basically the only plan I thought of - that or not running this test very often since it requires a special environment - and it seemed a bit heavyweight for a single test, probably there are other options though.

craigds: The other way to do this is to have a third fixture (`postgres_db`, `postgis_db`, and `no_postgis_db`). The latter depends on `postgres_db` and just runs `DROP EXTENSION IF EXISTS postgis`, within a transaction, and rolls it back afterwards.


It seems like you can actually do that (yay postgres!):

```
x=# begin;
BEGIN
x=*# drop extension postgis;
DROP EXTENSION
x=*# select st_asewkt('point(0 0)'::geometry);
ERROR:  type "geometry" does not exist
LINE 1: select st_asewkt('point(0 0)'::geometry);
                                       ^
x=!# rollback;
ROLLBACK
x=# select st_asewkt('point(0 0)'::geometry);
 st_asewkt
------------
 POINT(0 0)
(1 row)
```

However, for this to work we do need superuser access to the database.

fedch: what about now?

tests/conftest.py
```diff
@@ -994,6 +993,47 @@ def postgis_db():
...
+
+    with postgres_db.begin() as transaction:
+        try:
+            transaction.execute("DROP EXTENSION IF EXISTS postgis;")
+            # Test connection without postgis support
+            transaction.execute("SELECT postgis_version();")
+        except sqlalchemy.exc.DBAPIError:
+            pass
+        else:
+            raise pytest.skip("Expected PostGIS to be removed but it is still present")
```
craigds, line 1032: I'm not sure why this would happen - perhaps better to let it raise an error in this case rather than quietly skip some tests. Something unexpected happened

tests/conftest.py
```diff
@@ -994,6 +993,47 @@ def postgis_db():
...
+
+    if not is_superuser:
+        raise pytest.skip("Requires superuser access to the database")
+
+    with postgres_db.begin() as transaction:
+        try:
+            transaction.execute("DROP EXTENSION IF EXISTS postgis;")
+            # Test connection without postgis support
+            transaction.execute("SELECT postgis_version();")
+        except sqlalchemy.exc.DBAPIError:
```
craigds, line 1029: I think the try/except block should probably be outside the transaction contextmanager - the contextmanager will rollback when an exception propagates through it.

Otherwise it might try to commit,  and then explode because there's been an error on the database side.

tests/conftest.py
```diff
@@ -994,6 +993,47 @@ def postgis_db():
...
+
+@pytest.fixture()
+def no_postgis_db(postgres_db):
+    # Check if the user has superuser access to the database
+    is_superuser = False
+    with postgres_db.connect() as conn:
+        result = conn.execute(
+            "SELECT rolsuper FROM pg_roles WHERE rolname = current_user;"
+        )
+        is_superuser = result.scalar()
```
craigds, line 1019: you can do these in one go - see `conn.scalar(...)`

tests/test_working_copy_postgis.py
```diff
@@ -12,6 +12,7 @@
 from kart.sqlalchemy.adapter.postgis import KartAdapter_Postgis
 from test_working_copy import compute_approximated_types
 
+import psycopg2
```
craigds, line 15: unused?

vcpkg-vendor/vcpkg
craigds, line 1: This submodule on master is now pointing to 7a3bd64ed5d4cb3d665d4b124d168286be3fd7e5 , best to make sure it's doing that on this branch too

craigds (approved):

tests/conftest.py
craigds: looks great 👍

tests/conftest.py
```diff
@@ -994,6 +993,47 @@ def postgis_db():
     yield engine
 
 
+@pytest.fixture()
+def postgres_db():
```
olsen232, line 997: The postgres / postGIS fixtures are very similar, as is the little snippet of documentation inside them explaining how to run them. Let's merge them together a bit more - hopefully that will help avoid a situation where a contributor is confused about if they should be running postgres tests or postgis tests.

If preferred, you (or I) could do it in a follow up PR

Here's the idea:
```
@pytest.fixture()
def postgres_db():
    """
    Using docker, you can run all PostgreSQL /  PostGIS tests - such as test_postgis_import - as follows:
        docker run -it --rm -d -p 15432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust postgis/postgis
        KART_POSTGRES_URL='postgresql://postgres:@localhost:15432/postgres' pytest -k "postgres OR postgis" --pdb -vvs
    """
    # postgres_db implementation is the same as you have it now
    
@pytest.fixture()
def postgis_db(postgres_db):
  # See postgres_db for help on running PostGIS tests
  # postgis_db just takes postgres_db fixture and makes sure it has postGIS extension, otherwise it skips

@pytest.fixture()
def no_postgis_db(postgres_db):
  # See postgres_db for help on running no_postgis_db tests.
  # Same implementation as you have currently
```

olsen232 (approved):

tests/conftest.py
fedch: Hmm, I actually made the changes in this PR, would you take a look?

tests/conftest.py
```diff
@@ -974,24 +973,58 @@ def disable_editor():
...
+
+    try:
+        with postgres_db.begin() as transaction:
+            transaction.execute("DROP EXTENSION IF EXISTS postgis;")
+            # Test connection without postgis support
+            transaction.execute("SELECT postgis_version();")
+    except sqlalchemy.exc.DBAPIError:
+        pass
+    else:
+        raise pytest.skip("Expected PostGIS to be removed but it is still present")
```
olsen232, line 1025: (From earlier) [craigds](https://github.com/craigds) [19 hours ago](https://github.com/koordinates/kart/pull/828#discussion_r1178563198)
I'm not sure why this would happen - perhaps better to let it raise an error in this case rather than quietly skip some tests. Something unexpected happened

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 - Fixed a bug where Kart would require `user.name` and `user.email` to be set, even when `GIT_AUTHOR_EMAIL` and similar variables were set in the environment. [#812](https://github.com/koordinates/kart/issues/812)
 - Update native dependencies. [#820](https://github.com/koordinates/kart/pull/820)
 - Drop CI testing of Debian Stretch/oldoldstable. [#820](https://github.com/koordinates/kart/pull/820)
+- Allow kart to import non-spatial tables from PostgreSQL databases which don't have postGIS extension installed. [#728](https://github.com/koordinates/kart/issues/728)
```
olsen232, line 19: You should merge / rebase with master - git may not flag this as a conflict, but this should bullet point now be in 0.12.4 (UNRELEASED)

tests/conftest.py
```diff
@@ -974,24 +973,58 @@ def disable_editor():
...
+        is_superuser = conn.scalar(
+            "SELECT rolsuper FROM pg_roles WHERE rolname = current_user;"
+        )
+
+    if not is_superuser:
+        raise SystemError("Requires superuser access to the database")
+
+    try:
+        with postgres_db.begin() as transaction:
+            transaction.execute("DROP EXTENSION IF EXISTS postgis;")
```
craigds, line 1019: i think we need this fixture to `yield` inside this transaction, so that the test runs in a transaction where postgis extension has been dropped, but when we rollback (for other tests) it gets un-dropped again.

The line below (which always causes an error) causes the transaction to roll back, but we'd probably be better off with an explicit rollback - does the connection object have a rollback method?

craigds (approved):

PR #831, opened 2023-04-30
Suppress diffs caused by gdal info -stats -hist

olsen232:
or similar commands / software.
QGIS and other gdal clients can create these diffs without the user really being aware of them, and we don't want to distract the user with diffs they don't care about until and unless they make further changes.

Not yet done - unsuppress these diffs with the right options.

## Checklist:

### DIFF
--- a/kart/raster/pam_util.py
+++ b/kart/raster/pam_util.py
@@ -1,0 +1,53 @@
+import re
+
+
+def is_same_xml_ignoring_stats(lhs, rhs):
+    """
+    Returns True if two strings (left-hand-side, right-hand-side) of XML are the same,
+    or if they are the same except that gdal has inserted statistics into one of them,
+    in the form of a <Histograms> and/or <Metadata> block.
+    """
+    return (
+        lhs == rhs
+        or _lhs_is_rhs_minus_stats(lhs, rhs)
+        or _lhs_is_rhs_minus_stats(rhs, lhs)
+    )
+
+
+ANY_TAG = r"(<[^<>]+>)"
+HISTOGRAMS_PATTERN = re.compile(
+    rf"{ANY_TAG}\s*<Histograms>(.*)</Histograms>\s*{ANY_TAG}", re.DOTALL
+)
+# TODO - maybe check for non-statistics metadata.
+METADATA_PATTERN = re.compile(
+    rf"{ANY_TAG}\s*<Metadata>(.*)</Metadata>\s*{ANY_TAG}", re.DOTALL
+)
+
+
+def _lhs_is_rhs_minus_stats(lhs, rhs):
+    # Regex is sufficient for doing what we need to do here, which saves us parsing the XML.
+    for rhs_pattern in (HISTOGRAMS_PATTERN, METADATA_PATTERN):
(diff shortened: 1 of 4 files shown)


kart/raster/v1.py
```diff
@@ -237,11 +245,47 @@ def _non_pam_wc_paths(self, tile_and_pam_paths):
...
+        is that stats have been added or removed.
+        """
+        old_value = tile_delta.old_value
+        new_value = tile_delta.new_value
+
+        all_keys = set()
+        all_keys.update(old_value)
+        all_keys.update(new_value)
+        for key in all_keys:
+            if not key.startswith("pam") and old_value.get(key) != new_value.get(key):
```
craigds, line 268:
```suggestion
            if (not key.startswith("pam")) and old_value.get(key) != new_value.get(key):
```

kart/raster/v1.py
```diff
@@ -237,11 +245,47 @@ def _non_pam_wc_paths(self, tile_and_pam_paths):
...
+                return False
+
+        old_pam_oid = old_value.get("pamOid")
+        new_pam_oid = new_value.get("pamOid")
+        if old_pam_oid == new_pam_oid:
+            return True
+        if old_pam_oid is None or new_pam_oid is None:
+            return False
```
craigds, line 276: this is where the PAM is created/deleted?
if there's only stats in the PAM file, then maybe this should return True?

kart/raster/pam_util.py
```diff
@@ -0,0 +1,53 @@
...
+    rf"{ANY_TAG}\s*<Histograms>(.*)</Histograms>\s*{ANY_TAG}", re.DOTALL
+)
+# TODO - maybe check for non-statistics metadata.
+METADATA_PATTERN = re.compile(
+    rf"{ANY_TAG}\s*<Metadata>(.*)</Metadata>\s*{ANY_TAG}", re.DOTALL
+)
+
+
+def _lhs_is_rhs_minus_stats(lhs, rhs):
+    # Regex is sufficient for doing what we need to do here, which saves us parsing the XML.
```
craigds, line 28: not a big fan of this - any reason to not parse the XML?

I can see one issue with the regex which is where there are multiple Metadata or Histogram tags with something else in between - the regex will greedily match the whole lot and this function will return True.

That specific issue could be fixed by adding a `?` in the right place to make it match non-greedily but it seems like it'd be cleaner to parse the XML

olsen232: Yeah you're right - this function doesn't work properly when there are multiple bands, so I need to rework it.

Basically using regex was the most straightforward and also efficient way of doing `==` comparisons on large chunks of XML. The XML libraries we're using so far don't provide an easy way to see if an XML element in one document is equivalent to another XML element in the other document. I can write such a comparison operator but it seemed a bit circuitous - first we parse the XML into a tree, then we flatten the tree back into one of (a string, some bytes, a list of strings) so we can compare it to another flattened tree. If I leave it flat (as the regex does) I've saved a few steps and it will run faster... might be on the wrong track though

kart/raster/v1.py
olsen232: Done

craigds (approved):

Issue #729, opened 2022-11-07
Unhandled exception when creating WC in a Postrges working copy which doesn't have PostGIS extension


pfw:
**Describe the bug**
If the PostGIS extension isn't enabled on a database then there will be an unhandled exception  after init'ing a repo with a PG working copy and then importing data.

**To Reproduce**
Steps to reproduce the behaviour:

1. Create an empty PG database and don't enable PostGIS
2. `mkdir pgrepo`
3. `cd pgrepo`
4. `kart init --workingcopy postgresql://localhost/nopostgis/public`
5. `kart import /Volumes/GIS/data.gov.au/mdb-boundaries.gpkg`
6. See error

**Expected behaviour**

1. Kart should allow you to checkout a non-spatial working copy into a non-postgis database.
2. If there are spatial datasets in the repo, then the checkout should error out, with a message suggesting you install postGIS in your database.

**Output**
```
Starting git-fast-import...
Importing 1 features from mdb-boundaries.gpkg:mdbboundaries to mdbboundaries/ ...
Added 1 Features to index in 0.0s
Overall rate: 36 features/s)
Closed in 0s
Creating PostGIS working copy at postgresql://localhost/nopostgis/public ...
Traceback (most recent call last):
  File "sqlalchemy/engine/base.py", line 1901, in _execute_context
  File "sqlalchemy/engine/default.py", line 732, in do_execute
psycopg2.errors.UndefinedTable: relation "spatial_ref_sys" does not exist
LINE 1: SELECT auth_name, srtext FROM spatial_ref_sys WHERE srid = 4...
                                      ^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "kart_cli.py", line 4, in <module>
  File "kart/cli.py", line 392, in entrypoint
  File "click/core.py", line 1130, in __call__
  File "click/core.py", line 1055, in main
  File "kart/cli_util.py", line 68, in invoke
  File "click/core.py", line 1657, in invoke
  File "click/core.py", line 1404, in invoke
  File "click/core.py", line 760, in invoke
  File "click/decorators.py", line 26, in new_func
  File "kart/tabular/import_.py", line 350, in import_
  File "kart/working_copy.py", line 262, in reset_to_head
  File "kart/working_copy.py", line 299, in reset
  File "kart/working_copy.py", line 192, in create_parts_if_missing
  File "kart/tabular/working_copy/base.py", line 1249, in reset
[...]

PR #834, opened 2023-05-03
No postgis 729

fedch:
## Description

Making kart allow the user to checkout a non-spatial working copy into a non-postgis database.
If there are spatial datasets in the repo, then the checkout should error out, with a message suggesting you install postGIS in your database.

## Related links:

[729](https://github.com/koordinates/kart/issues/729)

## Checklist:

### DIFF
--- a/kart/tabular/v3.py
+++ b/kart/tabular/v3.py
@@ -450,3 +450,8 @@ class TableV3(RichTableDataset):
             existing_tree,
             resolve_missing_values_from_tree=resolve_missing_values_from_tree,
         )
+
+    @property
+    def is_spatial(self):
+        """Returns True if the dataset is spatial."""
+        return len(self.crs_definitions()) > 0
--- a/kart/tabular/working_copy/postgis.py
+++ b/kart/tabular/working_copy/postgis.py
@@ -1,15 +1,19 @@
 import contextlib
+
 import hashlib
 import logging
+from psycopg2.errors import UndefinedTable
 import time
 
 from kart import crs_util
+
 from kart.sqlalchemy import separate_last_path_part
 from kart.sqlalchemy.adapter.postgis import KartAdapter_Postgis
 from kart.schema import Schema
 from sqlalchemy import Index
 from sqlalchemy.dialects.postgresql.base import PGIdentifierPreparer
 from sqlalchemy.orm import sessionmaker
+from kart.exceptions import InvalidOperation, INVALID_OPERATION
 
 from .db_server import DatabaseServer_WorkingCopy
 from .table_defs import PostgisKartTables
@@ -337,3 +341,22 @@ class WorkingCopy_Postgis(DatabaseServer_WorkingCopy):
             sess.execute(
                 f"""ALTER TABLE {self.table_identifier(table)} ALTER COLUMN {self.quote(col.name)} TYPE {dest_type};"""
             )
+
+    def _check_for_unsupported_ds_types(self, sess, target_datasets):
+        """Check for spatial datasets and verify if the PostGIS extension is enabled in the database."""
+        for dataset_path, dataset in target_datasets.items():
+            if dataset.is_spatial:
+                spatial_ref_sys_exists = sess.scalar(
+                    """
+                    SELECT COUNT(*) FROM information_schema.tables
+                    WHERE table_schema='public' AND table_name='spatial_ref_sys';
+                    """
+                )
+                if spatial_ref_sys_exists:
+                    return
+                else:
+                    raise InvalidOperation(
+                        f"Dataset '{dataset_path}' requires the PostGIS extension to be installed in the working copy.\n"
+                        "Install it with `CREATE EXTENSION postgis;`",
+                        exit_code=INVALID_OPERATION,
+                    )
--- a/tests/test_working_copy_postgis.py
+++ b/tests/test_working_copy_postgis.py
@@ -798,3 +798,32 @@ def test_checkout_custom_crs(
                 "    },",
                 "  ]",
             ]
+
+
+def test_init_without_postgis(
+    cli_runner, data_archive, new_no_postgis_db_schema, tmp_path
+):
+    """This tests asserts that when trying to import a spatial gpkg into a Postgres wc, which doesn't have Postgis enabled,
+    the correct error message is shown."""
+
+    # Get the URL for the PostgreSQL database without PostGIS
+    # postgres_url = no_postgis_db.original_url + "/test_schema"
+    with data_archive("gpkg-points") as data:
+        table = "nz_pa_points_topo_150k"
+        src_gpkg_path = data / "nz-pa-points-topo-150k.gpkg"
+        with new_no_postgis_db_schema(create=True) as (postgres_url, _):
+            # Initialize the working copy
+            result = cli_runner.invoke(
+                ["init", str(tmp_path / "repo1"), "--workingcopy", postgres_url]
+            )
+            assert result.exit_code == 0, result.stderr
+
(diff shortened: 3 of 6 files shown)


tests/test_working_copy_postgis.py
```diff
@@ -798,3 +798,17 @@ def test_checkout_custom_crs(
...
             ]
+
+
+def test_init_without_postgis(cli_runner, postgres_drop_postgis):
+    # Get the URL for the PostgreSQL database without PostGIS
+    postgres_url = postgres_drop_postgis.original_url + "/test_schema"
+
+    # Initialize the working copy
+    result = cli_runner.invoke(["init", "--workingcopy", postgres_url])
+    assert result.exit_code == 0
```
craigds, line 809:
```suggestion
    assert result.exit_code == 0, r.stderr
```

kart/tabular/working_copy/postgis.py
```diff
@@ -337,3 +343,14 @@ def _apply_meta_schema_json(self, sess, dataset, src_value, dest_value):
...
             )
+
+    def _check_for_unsupported_ds_types(self, sess, target_datasets, schema="public"):
+        for target_dataset in target_datasets:
+            table = target_datasets[target_dataset]
+            if table.is_spatial:
+                inspector = inspect(sess.bind)
+                if "spatial_ref_sys" not in inspector.get_table_names():
+                    raise InvalidOperation(
+                        "The PostGIS extension is not enabled in the database.\nInstall it with `CREATE EXTENSION postgis;`",
```
craigds, line 354: This message should probably be more clearly related to what the user was trying to do - which is probably either `kart pull` or `kart clone` or `kart checkout mybranch`. They probably configured a postgres WC but may not know what postgis is, so I'd be a bit more direct with this, e.g.:

> Dataset 'roads' requires the postgis extension to be installed in the working copy.
> Install it with `CREATE EXTENSION postgis;`

kart/working_copy.py
```diff
@@ -17,6 +17,8 @@
 from kart.key_filters import RepoKeyFilter
 from kart.output_util import get_input_mode, InputMode
 
+from sqlalchemy import inspect
```
craigds, line 20: unused

kart/tabular/v3.py
```diff
@@ -450,3 +450,10 @@ def _apply_schema_json_delta_to_tree(
             existing_tree,
             resolve_missing_values_from_tree=resolve_missing_values_from_tree,
         )
+
+    @property
+    def is_spatial(self):
+        """Check if the table is spatial."""
+        if len(self.crs_definitions()) > 0:
+            return True
+        return False
```
craigds, line 459:
```suggestion
        return len(self.crs_definitions()) > 0
```

kart/tabular/working_copy/postgis.py
```diff
@@ -337,3 +343,14 @@ def _apply_meta_schema_json(self, sess, dataset, src_value, dest_value):
             sess.execute(
                 f"""ALTER TABLE {self.table_identifier(table)} ALTER COLUMN {self.quote(col.name)} TYPE {dest_type};"""
             )
+
+    def _check_for_unsupported_ds_types(self, sess, target_datasets, schema="public"):
+        for target_dataset in target_datasets:
+            table = target_datasets[target_dataset]
+            if table.is_spatial:
+                inspector = inspect(sess.bind)
```
craigds, line 351: I suspect this involves some communication with the database (not sure though) - I'd init it outside the loop rather than for each dataset

craigds (changes requested):

tests/test_working_copy_postgis.py
```diff
@@ -798,3 +798,25 @@ def test_checkout_custom_crs(
...
                 "  ]",
             ]
+
+
+def test_init_without_postgis(cli_runner, data_archive, no_postgis_db, tmp_path):
+    """This tests asserts that when trying to import a spatial gpkg into a Postgres wc, which doesn't have Postgis enabled,
+    the correct error message is shown."""
+
+    # Get the URL for the PostgreSQL database without PostGIS
+    postgres_url = no_postgis_db.original_url + "/test_schema"
```
olsen232, line 808: This will stop working very quickly when Kart finds that this schema is already in use - Kart is too scared to start messing with somebody's existing data, it likes to create a new schema or use an empty schema.  
What you need is `new_postgis_db_schema` - which is a fixture in conftest.py - except you need an equivalent fixture called `new_no_postgis_db_schema`. This will take care of giving you a new empty schema each time you run the test. You can look at other tests which use the `new_postgis_db_schema` to get an idea of how it's used.

Here's how I would modify new_postgis_db_schema so that you also have a new_no_postgis_db_schema:
```
@pytest.fixture()
def new_postgis_db_schema(request, postgis_db):
    return _new_pg_schema_for_test(request, postgis_db)


@pytest.fixture()
def new_no_postgis_db_schema(request, no_postgis_db):
    return _new_pg_schema_for_test(request, no_postgis_db)


def _new_pg_schema_for_test(request, engine):
    @contextlib.contextmanager
    def ctx(create=False):
        sha = hashlib.sha1(request.node.nodeid.encode("utf8")).hexdigest()[:20]
        schema = f"kart_test_{sha}"
        with engine.connect() as conn:
            # Start by deleting in case it is left over from last test-run...
            conn.execute(f"""DROP SCHEMA IF EXISTS "{schema}" CASCADE;""")
            # Actually create only if create=True, otherwise the test will create it
            if create:
                conn.execute(f"""CREATE SCHEMA "{schema}";""")
        try:
            url = urlsplit(engine.original_url)
            url_path = url.path.rstrip("/") + "/" + schema
            new_schema_url = urlunsplit(
                [url.scheme, url.netloc, url_path, url.query, ""]
            )
            yield new_schema_url, schema
        finally:
            # Clean up - delete it again if it exists.
            with engine.connect() as conn:
                conn.execute(f"""DROP SCHEMA IF EXISTS "{schema}" CASCADE;""")

    return ctx
```

kart/tabular/working_copy/postgis.py
```diff
@@ -337,3 +343,14 @@ def _apply_meta_schema_json(self, sess, dataset, src_value, dest_value):
             sess.execute(
                 f"""ALTER TABLE {self.table_identifier(table)} ALTER COLUMN {self.quote(col.name)} TYPE {dest_type};"""
             )
+
+    def _check_for_unsupported_ds_types(self, sess, target_datasets, schema="public"):
+        for target_dataset in target_datasets:
```
olsen232, line 348: target_datasets is a dict like { "dataset-path": dataset}
If you just want the dataset part of it, just iterate over it like this:
`for dataset in target_datasets.values()`
Or if you decide you need the path too, then do this:
`for dataset_path, dataset in target_datasets.items()`

kart/tabular/working_copy/postgis.py
```diff
@@ -337,3 +341,20 @@ def _apply_meta_schema_json(self, sess, dataset, src_value, dest_value):
...
+        """Check for spatial datasets and verify if the PostGIS extension is enabled in the database."""
+        for dataset_path, dataset in target_datasets.items():
+            if dataset.is_spatial:
+                spatial_ref_sys_exists = sess.scalar(
+                    """
+                    SELECT COUNT(*) FROM information_schema.tables
+                    WHERE table_schema='public' AND table_name='spatial_ref_sys';
+                    """
+                )
+                if not spatial_ref_sys_exists:
```
olsen232, line 355: We can early return here if spatial_ref_sys_exists is True - if spatial_ref_sys_exists is True then there's no reason to keep checking further, since we know we'll never hit the error condition

olsen232 (approved):

PR #838, opened 2023-05-11
Fix pthread_key exhaustion due to mod_spatialite

olsen232:
## Description

In kart, the mod_spatialite extension is loaded once per sqlite session - extensions aren't shared between sessions by default, since subsequent sessions might need different extensions.
This change patches our mod_spatialite so that it doesn't unload itself when the sqlite session ends - it stays loaded (if not necessarily active) until the process exits.
This avoids a leak in pthread_keys that was happening due to repeated loading and unloading of the mod_spatialite extension, which avoids test failures in gdal / raster specific tests caused by gdal crashing once pthread_keys are exhausted.

## Checklist:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 - Allow kart to import non-spatial tables from PostgreSQL databases which don't have postGIS extension installed. [#728](https://github.com/koordinates/kart/issues/728)
 - Fixes a bug where the current spatial filter isn't stored in the filesystem working copy, affecting the spatial filtering of point cloud datasets. [#833](https://github.com/koordinates/kart/pull/833)
+- Fixes pthread_key leaks in a long running Kart process due to repeated loading and unloading of mod_spatialite. [#838](https://github.com/koordinates/kart/pull/838)
```
rcoup, line 11: should add the amazonlinux change here too

.github/workflows/build.yml
rcoup, line 1: Good change, but this should probably be pulled out of this PR. Maybe cherry-pick the commit to master, then rebase this?

tests/raster/fixtures.py
```diff
@@ -1,28 +1,19 @@
...
-    with data_archive_readonly("raster/cog-aerial.tgz") as aerial:
-        try:
-            gdal.Info(f"{aerial}/aerial.tif")
-            has_gdal_info = True
-        except Exception as e:
-            click.echo(e, err=True)
-            has_gdal_info = False
+        assert gdal.Info is not None
+        has_gdal_info = True
+    except Exception as e:
```
rcoup, line 14: When would we have `gdal` but not `gdal.Info()`? I know we were trying to track down a bug, but do we need to keep this?

rcoup (approved):
I think this is good as a workaround in the meantime. Longer-term we should drop mod_spatialite in favour of loading libgdal as a sqlite module for the necessary geopackage support

CHANGELOG.md
olsen232: Done

.github/workflows/build.yml
olsen232: Done

tests/raster/fixtures.py
olsen232: Done

Issue #842, opened 2023-05-17
Improved cloud-optimized options / prompting


olsen232:
There's more work to do with respect to informing / prompting / warning the user about cloud-optimized vs non-cloud-optimized tiles - both for rasters and point clouds:

Import behaviour:
1. [x] Import with --convert-to-cog: no warning
2. [x] Import with --no-convert-to-cog: no warning
3. [x] Import with no flag and and interactive terminal
"... Would you like to make a COG dataset?"
4. [x] Import with no flag no interactive terminal:
"Run this command again with --convert-to-cog or --preserve-format."

Diff/commit behaviour:
1. [x] Diff/commit to preserve-format dataset: no warnings are shown.
2. [x] Diff/commit COG to cloud-optimized dataset: no warnings are shown.
3. [x] Diff/commit non-COG to convert-to-cog dataset, no flag - shows warning during diff or commit, prevents commit.
"Adding non-COG files to a cloud-optimized dataset is not allowed. Add --convert-to-dataset-format flag to cloud-optimize the files, or add --no-convert-to-dataset-format to remove the dataset constraints."
4. [x] Diff/commit non-COG to convert-to-cog dataset, --convert-to-dataset-format: files are automatically converted.
5. [x] Diff/commit non-COG files to convert-to-cog dataset, --no-convert-to-dataset-format: dataset constraint is removed.

Slightly related - diff/commit behaviour: Wrong LAZ version:
1. [x] Diff/commit non-COG to convert-to-cog dataset, no flag - shows warning during diff or commit, prevents commit.
"Adding multiple LAZ versions in the same dataset is not allowed. Add --convert-to-dataset-format flag to automatically convert them to the dataset's LAZ version."

PR #843, opened 2023-05-17
Require user to choose cloud-optimized or preserve-format

olsen232:
If the user hasn't chosen, prompts them to pick one or the other. The prompt tries to explain and inform them of the reason.

https://github.com/koordinates/kart/issues/842

## Checklist:

### DIFF
--- a/tests/raster/test_imports.py
+++ b/tests/raster/test_imports.py
@@ -79,16 +79,17 @@ AERIAL_SCHEMA_DIFF = [
 
 
 @pytest.mark.parametrize(
-    "convert_option",
+    "convert_option,convert_prompt_input",
     [
-        "--preserve-format",
-        "convert-to-cog?no",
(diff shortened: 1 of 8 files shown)


tests/raster/test_imports.py
```diff
@@ -81,8 +81,10 @@
 @pytest.mark.parametrize(
     "convert_option",
```
craigds, line 82: use a list of [param](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-param) instances to pass multiple args into the test function, rather than trying to encode all the arguments in one string

eg the argument names here could be `extra_options` (list), `input` (str)

craigds (approved):

fedch (approved):

tests/raster/test_imports.py
olsen232: Done

PR #844, opened 2023-05-18
583 attachments

fedch:
## Description

Working on handling attachments in Kart
- [X] handle file addition
- [X] handle file deletion

## Related links:

[583](https://github.com/koordinates/kart/issues/583)

## Checklist:

### DIFF
--- a/kart/json_diff_writers.py
+++ b/kart/json_diff_writers.py
@@ -64,7 +64,6 @@ class JsonDiffWriter(BaseDiffWriter):
         # (This is not quite as bad as it looks, since parts of the diff object are lazily generated.)
         repo_diff = self.get_repo_diff(diff_format=diff_format)
         self.has_changes = bool(repo_diff)
-
         for ds_path, ds_diff in repo_diff.items():
             ds_diff.ds_path = ds_path
 
--- a/kart/workdir.py
+++ b/kart/workdir.py
@@ -1,4 +1,5 @@
 import contextlib
+import os
 import click
 import logging
 from enum import Enum, auto
@@ -7,6 +8,7 @@ from pathlib import Path
 import shutil
 import subprocess
 import sys
+from kart.structure import RepoStructure
 
 import pygit2
 import sqlalchemy as sa
@@ -34,7 +36,9 @@ from kart.tile import ALL_TILE_DATASET_TYPES
 from kart.tile.tile_dataset import TileDataset
 from kart.tile.tilename_util import remove_any_tile_extension, PAM_SUFFIX
 from kart.working_copy import WorkingCopyPart
-
+from kart.diff_structs import FILES_KEY, BINARY_FILE, DatasetDiff, RepoDiff
+from kart.diff_util import get_file_diff, get_repo_diff
+from .base_diff_writer import BaseDiffWriter
 
 L = logging.getLogger("kart.workdir")
 
@@ -380,6 +384,12 @@ class FileSystemWorkingCopy(WorkingCopyPart):
             self.write_full_datasets_to_workdir(
                 [target_datasets[d] for d in ds_inserts]
             )
+        # Update the working copy with files that have changed:
+        kart_attachments = os.environ.get("X_KART_ATTACHMENTS")
+        if kart_attachments:
+            if base_tree and target_tree:
+                self.update_file_diffs(base_tree, target_tree)
+
         for ds_path in ds_updates:
             self._update_dataset_in_workdir(
                 ds_path,
@@ -394,6 +404,25 @@ class FileSystemWorkingCopy(WorkingCopyPart):
             self._update_state_table_spatial_filter_hash(
                 sess, self.repo.spatial_filter.hexhash
             )
+
+    def update_file_diffs(self, base_tree, target_tree):
+        """Get the deltas for attachment files and write them to the working copy."""
+        repo = self.repo
+        base_rs = RepoStructure(repo, base_tree)
+        target_rs = RepoStructure(repo, target_tree)
+        attachment_deltas = get_file_diff(base_rs, target_rs)
+
+        for filename, file_delta in attachment_deltas.items():
+            new_path = self.path / filename
+
+            # Delete the old file
+            if file_delta.old and new_path.is_file():
+                new_path.unlink()
+
+            # Create a new file
+            if file_delta.new:
+                blob_data = repo[file_delta.new.value].data
+                new_path.write_bytes(blob_data)
 
     def _diff_to_reset(
         self, ds_path, base_datasets, target_datasets, workdir_diff_cache, ds_filter
--- a/tests/test_meta.py
+++ b/tests/test_meta.py
@@ -1,6 +1,8 @@
 import json
+import os
 import pytest
 import subprocess
+import pathlib
 
 import kart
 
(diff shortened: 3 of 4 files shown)


kart/diff_util.py
```diff
@@ -182,9 +184,19 @@ def get_file_diff(
     # We don't yet support attachment diffs in the workdir
     assert not include_wc_diff
 
-    old_tree = base_rs.tree
-    new_tree = target_rs.tree
-    repo = target_rs.repo
+    if hasattr(base_rs, "type") and hasattr(target_rs, "type"):
```
craigds, line 187: I'd suggest changing the method signature above so it always takes trees and a repo, rather than RepoStructure objects. It's only called from a couple of places, so this won't be a big deal for callers

tests/test_meta.py
```diff
@@ -254,6 +254,38 @@ def test_commit_files(data_archive, cli_runner):
         assert r.exit_code == 44, r.stderr
 
 
+def test_commit_files_mine(data_archive, cli_runner):
+    with data_archive("points"):
```
craigds, line 258: look into using `data_working_copy` for this test so it actually tests your new checkout code

kart/diff_util.py
```diff
@@ -166,6 +167,7 @@ def get_dataset_diff(
 def get_file_diff(
```
olsen232, line 167: FYI: In workdir.py, in the reset() function, it's possible - at the place where the Datasets objects are obtained - to obtain RepoStructure objects there as well. If we did that, we wouldn't have to change this function here.

However, I can't really provide a strong reason why that way is obviously better than this way... so I just added this comment while I think about it a bit more

kart/json_diff_writers.py
```diff
@@ -62,9 +62,10 @@ def add_json_header(self, obj):
     def write_diff(self, diff_format=DiffFormat.FULL):
         # TODO - optimise - no need to generate the entire repo diff before starting output.
         # (This is not quite as bad as it looks, since parts of the diff object are lazily generated.)
+        breakpoint()
```
olsen232, line 65: Make sure you remove all breakpoint() and print() before you merge

olsen232 (commented):
This looks like it is starting to do something useful - I think it would work to check out files into the workdir.

- It still needs tests.
- Maybe we need a way to turn this off until it's complete - checking out these files into the working copy probably isn't that great a user experience if we still don't have a way to detect changes to them. They might be better remaining hidden for a bit longer.

kart/workdir.py
```diff
@@ -380,6 +384,14 @@ def reset(
             self.write_full_datasets_to_workdir(
                 [target_datasets[d] for d in ds_inserts]
             )
+        # Update the working copy with files that have changed:
+        kart_attachments = os.environ.get("KART_ATTACHMENTS")
+        if kart_attachments == "true":
```
olsen232, line 389: Previously we've just checked if the value of these env-variables is truthy by python standards - this means that "True" or "YES" or "1" will also work.
It does mean that "0" or "False" don't disable this properly but generally people will disable this by completely unsetting the variable, so it works pretty well for how simple it is.

kart/workdir.py
```diff
@@ -380,6 +384,14 @@ def reset(
...
+            if base_tree and target_tree:
+                self.update_file_diffs(base_tree, target_tree)
+        else:
+            print("No environment variable KART_ATTACHMENTS set to true")
```
olsen232, line 393: Don't print anything here - the idea here is that 
a) you can develop against master, which is easier than developing on a feature branch for long periods of time
b) we can continue to release the master branch but without anything noticeably changing, until we're ready to effective turn on KART_ATTACHMENTS for everyone.
Printing out that its turned off will be a noticeable change for users that won't be of any benefit to them.

olsen232: Also, this extra output is causing some tests to fail - easier to remove this output than to fix the tests.

tests/test_meta.py
```diff
@@ -254,6 +255,75 @@ def test_commit_files(data_archive, cli_runner):
         assert r.exit_code == 44, r.stderr
 
 
+def test_commit_files_add_and_delete(data_working_copy, cli_runner):
```
olsen232, line 258: This test doesn't actually test your new functionality - if I revert your new functionality (or disable KART_ATTACHMENTS) the test still passes.

This test tests that the output of git ls-tree contains the expected attachment. git ls-tree is used for looking inside the Git object database, so, this is testing that the Git ODB contains the expected attachment. It is the job of `kart commit-files` to make sure that the attachment is saved to the ODB, and it was already working.

Your new code doesn't change the ODB - it changes what's on the filesystem. You can use the pathlib.Path library in this test to check what's on the file system and make sure the file you're concerned with doesn't exist before it is created, does exist once it is created and has the right contents, and doesn't exist after it is deleted.

tests/test_meta.py
```diff
@@ -254,6 +255,75 @@ def test_commit_files(data_archive, cli_runner):
...
+                "my_attachment.txt=My text",
+            ]
+        )
+        assert r.exit_code == 0, r.stderr
+
+        p1 = subprocess.Popen(
+            ["git", "ls-tree", "-t", "-r", "HEAD"], stdout=subprocess.PIPE
+        )
+        p2 = subprocess.Popen(
+            ["grep", "-v", "dataset"], stdin=p1.stdout, stdout=subprocess.PIPE
```
olsen232, line 276: This will go away anyway if you make the changes suggested above, but, best practice is to filter python lists using python, eg use list comprehensions.
grep might not be present on every machine, or might work differently on different machines, whereas something like [line for line in all_lines if "dataset" not in line] will always work

tests/test_meta.py
```diff
@@ -254,6 +256,50 @@ def test_commit_files(data_archive, cli_runner):
...
 
+def test_commit_files_add_and_delete(data_working_copy, cli_runner):
+    with data_working_copy("point-cloud/auckland") as (repo_dir, wc):
+        # Set the environment variable
+        os.environ["X_KART_ATTACHMENTS"] = "true"
+        # Define file path
+        file_path = pathlib.Path(repo_dir) / "my_attachment.txt"
+
+        # Make sure the file does not exist yet
+        assert not file_path.exists(), r.stderr
```
olsen232, line 267: Remove r.stderr - I don't think r is defined at this point

Basically any time you run a command that you expect to succeed, you can do this:
`assert r.exit_code == 0, r.stderr`
and for other assertions, you shouldn't be referring to whatever is in r.stderr since it may or may not be relevant. You don't necessarily need an extra message for other assertions - you can if you like add a bit of extra info - eg the value of file_path - but you'll see that in most places we haven't bothered

tests/test_meta.py
```diff
@@ -254,6 +256,50 @@ def test_commit_files(data_archive, cli_runner):
...
+        )
+        assert r.exit_code == 0, r.stderr
+
+        # Check that the file exists now
+        assert file_path.exists(), f"Expected file {file_path} not found."
+
+        # Check file contents
+        expected_contents = "My text"
+        actual_contents = file_path.read_text()
+        assert actual_contents == expected_contents, r.stderr
```
olsen232, line 286: r.stderr shouldn't be here

tests/test_meta.py
```diff
@@ -254,6 +256,50 @@ def test_commit_files(data_archive, cli_runner):
...
+                "commit-files",
+                "-m",
+                "Deleting a file",
+                "--remove-empty-files",
+                "my_attachment.txt=",
+            ]
+        )
+
+        # Check that the file does not exist now
+        assert not file_path.exists(), r.stderr
```
olsen232, line 300: r.stderr shouldn't be here

craigds (approved):
Looks great, nice work 👍

Issue #847, opened 2023-05-23
CI (windows): error: Microsoft Visual C++ 14.0 or greater is required.


olsen232:
On windows, during building of the cryptography wheel, the following error occurs:
```
error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
error: subprocess-exited-with-error
    
Building wheel for cryptography (pyproject.toml) did not run successfully.
exit code: 1
```

Seemingly, a straight forward issue to solve: the error contains instructions on how to fix it. However, according the GH actions, the runner we are using already has a whole pile of Microsoft Visual C++ paraphernalia - see https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md#microsoft-visual-c - specifically, `Microsoft Visual C++ 2019 Debug Runtime v14.29.30139 for x64` looks like it should be adequate. We also have the "redistributable" part of that runtime (in case that's a possible issue, it's not clear to me which entities - debug runtimes, additional runtimes, minimum runtimes, redistributable - might or might not be a superset of the other ones, but, having the redistributable bit seems to be important, and I can confirm we have the same version of the redistributable.)

Furthermore the environment variables are set such that it should be easy for anyone to find the tools + redistribable:
```
    VCINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\
    VCToolsInstallDir: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\
    VCToolsRedistDir: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Redist\MSVC\14.29.30133\
    VCToolsVersion: 14.29.30133
```

However when SSHing in I did find the following potential issue: by the time we try to build the cryptography package, our PATH is dangerously massive, to the point where it may well be interfering with MSVC tooling. For instance, running the command:
`c:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat amd64`
fails with
`The input line is too long.`
[...]

PR #848, opened 2023-05-25
Harden vcvarsall.bat

olsen232:
vcvarsall.bat is a batch file to set up the environment ready for building using MSVC++ - it sets ENV variables related to the host architecture, target architecture, MSVC version, the PATH to various tools, depending on the arguments it is called with.

vcvarsall.bat fails if the PATH is too long.
It tends to cause an error message like:
"Microsoft Visual C++ 14.0 or greater is required" This is because the wheel-building python module on windows tries to run vcvarsall.bat, and if it fails, it guesses that the problem is that you don't have the right MSVC version installed.

By the time we try to build the wheels, in the CMake buils step, our PATH is at least 7600 chars long, which is long enough to crash vcvarsall.bat

Happily we can fix the PATH without any adverse effects simply by removing duplicate entries, which brings it down to 5400 chars long, and so vcvarsall.bat starts working.

Finding the right place to fix the PATH has been a challenge - various tools including vcvarsall.bat itself prepend things to the PATH while they work. Trying to fix the PATH immediately before kicking off the CMake build didn't fix this issue, it just changed where it occurred. In the end, fixing the PATH in vcvarsall.bat itself in a safe way (only useless duplicate entires are removed) seemed to be the most robust option / the only one I that I could get to work.

Fixes https://github.com/koordinates/kart/issues/847

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -989,6 +989,17 @@ jobs:
       - name: "msvc setup"
         uses: ilammy/msvc-dev-cmd@v1
 
+      - name: "Harden vcvarsall.bat"
+        shell: pwsh
+        run: |
+          cd 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build'
+          mv vcvarsall.bat orig_vcvarsall.bat
+          echo @'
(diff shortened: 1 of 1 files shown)


craigds (approved):
nice find. what a mission

PR #850, opened 2023-05-29
Maintain a VRT file for each raster dataset in the workdir

olsen232:
Basic implementation: each time a raster dataset is updated by Kart, rewrites the VRT file to reference every tif file currently in the dataset, using gdal.BuildVRT.


craigds (approved):

rcoup (approved):
LGTM 👍

PR #851, opened 2023-05-31
Fix tab-completion to work with helper mode

olsen232:
Click tab completion uses the name of the program for various environment variables eg _KART_COMPLETE.
Our program name varies depending on how the program is run kart vs kart_cli depending on helper mode etc.
Since we can't be sure if the way kart is currently being run is the same as it was run when tab-completion was installed, its safest to stick to one "canoncial" program name for tab-completion at all times. Click lets us do this. The chosen name is "kart" and therefore the variable is _KART_COMPLETE.

## Checklist:

### DIFF
--- a/kart/cli.py
+++ b/kart/cli.py
@@ -403,7 +403,9 @@ def load_commands_from_args(args, skip_first_arg=True):
 def entrypoint():
     freeze_support()
     load_commands_from_args(sys.argv)
-    cli()
+    # Don't let helper mode mess up the usage-text, or the shell complete environment variables.
+    prog_name = "kart" if os.path.basename(sys.argv[0]) == "kart_cli" else None
+    cli(prog_name=prog_name, complete_var="_KART_COMPLETE")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


kart/cli.py
```diff
@@ -403,7 +403,9 @@ def load_commands_from_args(args, skip_first_arg=True):
 def entrypoint():
     freeze_support()
     load_commands_from_args(sys.argv)
-    cli()
+    # Don't let helper mode mess up the usage-text, or the shell complete environment variables.
+    prog_name = "kart" if os.path.basename(sys.argv[0]) == "kart_cli" else None
```
craigds, line 407: what's the `None` case here for? Would that handle a case where you created a shell alias/symlink to kart called `k` or something like that?

craigds (approved):

kart/cli.py
olsen232: Yes, it seems to correctly handle that case - if you rename or alias kart, you get usage info for the renamed kart.
Depending on why kart was renamed or aliased, this may or may not be desirable, but at least it's not like tab-completion where something breaks if we get it wrong.

PR #860, opened 2023-06-07
Error if the user tries to import from within the WC

olsen232:
The appropriate command to use instead is `kart add-dataset`

https://github.com/koordinates/kart/issues/249

## Checklist:

### DIFF
--- a/kart/sqlalchemy/__init__.py
+++ b/kart/sqlalchemy/__init__.py
@@ -135,6 +135,14 @@ def strip_password(uri):
     return p.geturl()
 
 
+def strip_query(uri):
+    """Removes query parameters from URI."""
+    p = urlsplit(uri)
+    if p.query is not None:
+        p = p._replace(query=None)
+    return p.geturl()
+
+
 def separate_last_path_part(uri):
     """
     Removes the last part of the path from a URI and returns it separately.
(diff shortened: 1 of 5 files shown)


kart/sqlalchemy/__init__.py
```diff
@@ -153,6 +153,26 @@ def separate_last_path_part(uri):
     return (modified_url, last_part)
 
 
+def truncate_to_path_length(uri, path_length):
+    """
+    Removes any number of path-parts from a URI to truncate it to the required length.
+    Eg, with path_length of 1, it would truncate SCHEME://HOST/DBNAME/DBSCHEMA/TABLE to SCHEME://HOST/DBNAME
```
craigds, line 159: these right? looks like the two examples have been flipped

craigds: feels like a test for this function specifically would be useful

kart/sqlalchemy/__init__.py
```diff
@@ -153,6 +153,26 @@ def separate_last_path_part(uri):
...
+    Or with path_lengh of 2, it would truncate SCHEME://HOST/DBNAME/DBSCHEMA/TABLE to SCHEME://HOST/DBNAME/DBSCHEMA
+    If the uri's path length is already as short or shorter as the given path length, then it has no effect.
+    """
+    url = urlsplit(uri)
+    url_path = PurePosixPath(url.path)
+    current_path_length = len(url_path.parents)
+    if current_path_length <= path_length:
+        return uri
+
+    modified_url_path = str(url_path.parents[-1 - path_length])
```
craigds, line 169: _feels_ like this should just be `-path_length` - the `-1` seems wrong because this is a list of parents.
ie if path_length=1 this should return the last parent ie index -1

kart/tabular/import_.py
```diff
@@ -347,3 +350,44 @@ def table_import(
...
+        )
+        is_wc_import = wc_location == source
+
+    if is_wc_import:
+        raise click.UsageError(
+            "Import-source is already inside working-copy.\n"
+            "The `kart import` command creates a new Kart dataset that is a copy of a supplied import-source that is external to Kart.\n"
+            "It looks like you want Kart to start tracking and take control of a table that is already inside Kart's working copy.\n"
+            "To have Kart start tracking a table that is already inside the working copy, use `kart add-dataset TABLENAME`\n"
+            "To see a list of tables inside the working copy that are untracked, use `kart status --list-untracked-tables`\n"
```
craigds, line 388:
```suggestion
            "The `kart import` command creates a new Kart dataset by copying from an import-source that is external to Kart.\n"
            "To have Kart start tracking a table that is already inside the working copy, use `kart add-dataset TABLENAME`\n"
            "To see a list of tables inside the working copy that are untracked, use `kart status --list-untracked-tables`\n"
```

kart/sqlalchemy/__init__.py
olsen232: Dropped this function - realised "startswith" works pretty well here

kart/sqlalchemy/__init__.py
olsen232: Dropped this function - realised "startswith" works pretty well here

kart/tabular/import_.py
olsen232: Done

craigds (approved):

Issue #863, opened 2023-06-09
Kart truncates trailing zeroes in values of type `numeric`


craigds:
**Describe the bug**
Importing from a database table where a field has `NUMERIC` type causes any trailing zeroes in the PK to be stripped.

**To Reproduce**
1. init a postgres db with this:

```sql
begin;

create table importsource (
    mypk numeric PRIMARY KEY,
    otherfield numeric NULL
);
INSERT INTO importsource (mypk, otherfield) VALUES ('13390', '13390');
INSERT INTO importsource (mypk, otherfield) VALUES ('200000', '200000');
INSERT INTO importsource (mypk, otherfield) VALUES ('12345', '12345');
INSERT INTO importsource (mypk, otherfield) VALUES ('8888', '8888');
INSERT INTO importsource (mypk, otherfield) VALUES ('0', '0');
INSERT INTO importsource (mypk, otherfield) VALUES ('9999', '9999');

commit;
```

2. import the table into a Kart repo

```
kart init --bare xyz
kart import postgresql://postgres:@localhost:15432/postgres importsource:dataset --message='initial' --no-checkout --primary-key=mypk
```

**Expected behaviour**

the values should be unchanged

**Output**

```$ kart show
commit ed000ef99004290ba89c878ef649add804b8a076
Author: Craig de Stigter <craig@destigter.nz>
Date:   Fri Jun  9 13:07:09 2023 +1200

    initial

+++ dataset:meta:schema.json
+ [
+   {
+     "id": "4f5068af-f107-7712-9a5e-c762426bdb41",
+     "name": "mypk",
+     "dataType": "numeric",
+     "primaryKeyIndex": 0
+   },
+   {
+     "id": "b9dfa572-020a-1f24-d7e4-fb0db17f668e",
+     "name": "otherfield",
+     "dataType": "numeric"
+   }
+ ]
+++ dataset:feature:
+                                     mypk =
+                               otherfield =
+++ dataset:feature:12345
+                                     mypk = 12345
+                               otherfield = 12345
+++ dataset:feature:1339
+                                     mypk = 1339
+                               otherfield = 1339
+++ dataset:feature:2
+                                     mypk = 2
+                               otherfield = 2
+++ dataset:feature:8888
+                                     mypk = 8888
+                               otherfield = 8888
+++ dataset:feature:9999
+                                     mypk = 9999
+                               otherfield = 9999
```

Values `13390`, `200000`, and `0` all had their trailing zeroes stripped.

[...]

PR #867, opened 2023-06-12
Fix: whole-number numerics read from PG were losing trailing zeroes

olsen232:
## Description

Also strengthened our roundtrip tests to catch any other bugs like this.
Didn't find any other data-loss bugs, but found a weird property that reading float32s from PG would be rounded to the nearest decimal value that accurately describes the float32. The resulting value actually has more bits than a float32. Kart can store the extra bits no problem since it stores everything in float64, but it's better if we don't store these extra bits - they don't mean anything, and they don't roundtrip through other database types.

## Related links:

Fixes https://github.com/koordinates/kart/issues/863

## Checklist:


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 - Fixes a bug where git-lfs (and potentially other subprocesses) wasn't working reliably with helper mode, affects macOS and Linux. [#853](https://github.com/koordinates/kart/issues/853)
 - `import` now informs the user to use `add-dataset` instead if they try to import a table that is already inside the working copy. [#249](https://github.com/koordinates/kart/issues/249)
 - Fixes a bug where datasets with no CRS couldn't be checked out into a Geopackage working copy. [#855](https://github.com/koordinates/kart/issues/855)
+- Fixes a bad data-loss bug where whole-number numerics read from a Postgres database were losing trailing zeroes (eg 100 was read as 1). [#863](https://github.com/koordinates/kart/issues/863)
```
craigds, line 22: I'd drop the 'bad data-loss bug' here. The bug isn't necessarily bad - depends what you're doing and for many users it won't be relevant or important.

It's worth highlighting bigger than a bottom-of-list item though, so the user can make their own decision. Maybe let's move this to its own section? how about this:

### Fix: Corruption of Postgres `NUMERIC` values

⚠️ This release fixes a bug causing corruption of `NUMERIC` fields when read from Postgres databases.

If your repositories use the `NUMERIC` or `NUMERIC(a,b)` types and were imported/committed from a Postgres source or working copy, you may have lost data.

Specifically, any numeric values with a scale of zero (no digits after the decimal point) were affected. If the only numerics in your repositories were defined with a scale greater than zero, no data corruption should have occured.

For example, a value of `100` in a postgres database with the type `NUMERIC(10, 0)` or `NUMERIC` would have been imported as `1` - losing its trailing zeroes.

However, a value of `100.0` in a field declared as `NUMERIC(10, 1)` would not have been affected.

Other data types (integer, floating-point etc) are not affected, and repos sourced from other SQL databases (Geopackage, MSSQL, MySQL) were not affected.

kart/sqlalchemy/adapter/postgis.py
```diff
@@ -409,23 +411,24 @@ def sql_read(self, column):
...
+    float32_packer = struct.Struct("@f")
 
-    def sql_read(self, column):
-        return sa.cast(column, TEXT)
+    def __init__(self, size):
+        self.size = size
+
+    def python_postread(self, value):
+        if value is not None and self.size == 32:
+            return self.float32_packer.unpack(self.float32_packer.pack(value))[0]
```
craigds, line 422: probably need to add a comment on why

kart/sqlalchemy/adapter/postgis.py
```diff
@@ -409,23 +411,24 @@ def sql_read(self, column):
...
+        return value
 
 
 @aliased_converter_type
-class NumericType(ConverterType):
-    """ConverterType to read numerics as text. They are stored in PG as NUMERIC but we read them back as text."""
+class IntervalType(ConverterType):
+    """ConverterType to that casts intervals to text - ISO8601 mode is set for durations so this does what we want."""
```
craigds, line 428: add an example showing the form of the result

craigds (commented):
What did we decide to do about unconstrained numerics imported from postgres and then checked out into MSSQL/MySQL?

AFAICT this will cause them to get truncated at scale=0.

Not sure how we can handle them better than that. We could add some sensible nonzero scale to them maybe?

CHANGELOG.md
olsen232: I'll put this in when I reformat this changelog entry for the actual release 👍

craigds (approved):

kart/sqlalchemy/adapter/postgis.py
olsen232: Done

kart/sqlalchemy/adapter/postgis.py
olsen232: Done

PR #869, opened 2023-06-14
Fix sigint handling when running in helper mode

olsen232:
## Description

If kart.c receives a SIGINT while its waiting for python kart_cli, we need to SIGINT python kart_cli too.

### DIFF
--- a/cli_helper/kart.c
+++ b/cli_helper/kart.c
@@ -282,8 +282,8 @@ int main(int argc, char **argv, char **environ)
         int fp = open(getcwd(NULL, 0), O_RDONLY);
         int fds[4] = {fileno(stdin), fileno(stdout), fileno(stderr), fp};
 
-        char *socket_filename = malloc(strlen(getenv("HOME")) + strlen(".kart.socket") + 2);
-        sprintf(socket_filename, "%s/%s", getenv("HOME"), ".kart.socket");
+        char *socket_filename = malloc(strlen(getenv("HOME")) + strlen(".kart.1234567890.socket") + 2);
+        sprintf(socket_filename, "%s/.kart.%d.socket", getenv("HOME"), getsid(0));
         int socket_fd = socket(AF_UNIX, SOCK_STREAM, 0);
 
         struct sockaddr_un addr;
@@ -304,8 +304,6 @@ int main(int argc, char **argv, char **environ)
                 // process are left open in it
                 if (fork() == 0)
                 {
-                    setsid();
-
                     // start helper in background and wait
                     char *helper_argv[] = {&cmd_path[0], "helper", "--socket", socket_filename, NULL};
 
--- a/cli_helper/kart.c
+++ b/cli_helper/kart.c
@@ -194,7 +194,7 @@ int is_helper_enabled()
 /**
  * @brief Exit signal handler for SIGALRM
  */
-void exit_on_alarm(int sig)
+void exit_on_sigalrm(int sig)
 {
     int semval = semctl(semid, SEMNUM, GETVAL);
     if (semval < 0)
@@ -207,6 +207,21 @@ void exit_on_alarm(int sig)
     semctl(semid, SEMNUM, IPC_RMID);
     debug("sigalrm: semid=%d semval=%d exit_code=%d\n", semid, semval, exit_code);
     exit(exit_code);
+}
+
+char *socket_filename;
+
+/**
+ * @brief Exit signal handler for SIGINT
+ */
+void exit_on_sigint(int sig)
+{
+    putchar('\n');
+    const char* kill_cmd_template = "lsof -t '%s' | xargs kill";
+    char* kill_cmd = malloc(strlen(kill_cmd_template) + strlen(socket_filename) + 2);
+    sprintf(kill_cmd, kill_cmd_template, socket_filename);
+    system(kill_cmd);
+    exit(128 + sig);
 }
 
 int main(int argc, char **argv, char **environ)
@@ -267,7 +282,7 @@ int main(int argc, char **argv, char **environ)
         int fp = open(getcwd(NULL, 0), O_RDONLY);
         int fds[4] = {fileno(stdin), fileno(stdout), fileno(stderr), fp};
 
-        char *socket_filename = malloc(strlen(getenv("HOME")) + strlen(".kart.socket") + 2);
+        socket_filename = malloc(strlen(getenv("HOME")) + strlen(".kart.socket") + 2);
         sprintf(socket_filename, "%s/%s", getenv("HOME"), ".kart.socket");
         int socket_fd = socket(AF_UNIX, SOCK_STREAM, 0);
 
@@ -370,7 +385,8 @@ int main(int argc, char **argv, char **environ)
         memcpy((int *)CMSG_DATA(cmsg), fds, sizeof(fds));
         msg.msg_controllen = cmsg->cmsg_len;
 
-        signal(SIGALRM, exit_on_alarm);
+        signal(SIGALRM, exit_on_sigalrm);
+        signal(SIGINT, exit_on_sigint);
 
         if (sendmsg(socket_fd, &msg, 0) < 0)
         {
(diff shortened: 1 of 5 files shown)


cli_helper/kart.c
```diff
@@ -209,6 +209,21 @@ void exit_on_alarm(int sig)
...
 
+char *socket_filename;
+
+/**
+ * @brief Exit signal handler for SIGINT
+ */
+void exit_on_sigint(int sig)
+{
+    putchar('\n');
+    const char* kill_cmd_template = "lsof -t '%s' | xargs kill";
```
craigds, line 220: kill sends TERM by default, you probably want `-s INT`

I'd probably prefer a non-subprocess implementation - `kill(pid, SIGINT)` is easy enough - are there any C replacements for lsof?

kart/__init__.py
```diff
@@ -199,7 +199,10 @@ def _cleanup_process_group(signum, stack_frame):
             if _kart_process_group_killed:
                 return
             _kart_process_group_killed = True
-            os.killpg(os.getpid(), signum)
+            try:
+                os.killpg(os.getpid(), signum)
+            except Exception:
```
craigds, line 204: can you explain this change? I don't think I get it

cli_helper/kart.c
rcoup: AFAICT `lsof` isn't part of most Linux base images. `netstat -p` might be an option on Linux but not macOS, but again `netstat` isn't always installed with the base images.

(We could add a dependency to lsof/netstat/whatever on Linux I guess)

rcoup (changes requested):
If the only solution is `lsof`, we need to add it as a dependency of our packages. Would prefer a more standalone solution if there is a practical/sane one.

cli_helper/kart.c
olsen232: Re: replacing lsof with a non-subprocess - no-extra-dependency solution:
lsof looks complicated - when I went looking for C alternatives, people suggested that you just search through the data in /proc until you find what you are looking for, which is what lsof does. Eg here:
https://unix.stackexchange.com/questions/302152/is-there-a-way-to-identify-the-pid-or-cgroup-of-a-socket-without-iterating-thr

olsen232: Re: sending out-of-band requests over the socket
It's possible. Peter actually sends very little over the socket. Return exit codes are not sent over the socket, they are sent using`kill` - so we only ever send messages over the socket in one direction. Here's the current flow:
User runs kart.c
Kart.c finds the socket
Kart.c sends a message containing its pid, FDs, environment, command-line arguments...
Kart.c then just sleeps for up to 24 hours.
helper.py is in a never ending while loop. It receives the message and invokes the appropriate click command.
if the command returns or throws an exception, helper.py sends kill SIGALRM to the calling PID

So with current flow, the helper.py while loop is single-threaded: it listens for a message, then executes the message, etc. I could try and make it multi-threaded so it can listen for a signal to kill itself:
- pro - one less dependency (lsof)
- con - maybe a bit less robust to rely on a process that's already not doing what the user wants, to be listening for signals and kill itself. We only get one chance to kill it - if kart.c is killed and helper.py survives, we end up in a bad state where helper.py is running in the background but connected to your tty, and no ^C special keypresses will do anything to it.

kart/__init__.py
rcoup: > but maybe since it was already being killed, it was no longer findable?

Is `killpg(mypid)` any different from `killpg(0)`?

cli_helper/kart.c
rcoup: Is it an option to call setpgid in the per-invocation execution fork?

eg:
1. Background Kart process is PID 100
2. `kart do stuff` → invokes C helper as PID 200
3. 100 connects to background helper, sends args/file-descriptors
4. 100 forks to 201, calls `exec()`
5. 201 calls `setpgid(100)`**
6. if the user hits ctrl-c, the progress group receives the signal: both 100 & 201 & any further subprocesses, but not 200

** it should probably be the result of `getpgrp()` rather than it's own PID.

olsen232: This looks nicer than lsof - I'll try to get it to work

kart/helper.py
```diff
@@ -169,6 +170,16 @@ def helper(ctx, socket_filename, timeout, args):
                         f"Payload:\n{repr(payload)}",
                     )
                 else:
+                    try:
+                        # Join the process group of the calling process - so that if they get killed, we get killed to.
+                        os.setpgid(0, calling_environment["pid"])
+                        os.environ["_KART_PGID_SET"] = "1"
+                    except Exception as e:
```
craigds, line 177: I suspect this could catch a more specific exception class (OSError? or one of the subclasses: https://docs.python.org/3/library/exceptions.html#os-exceptions)

craigds (approved):
looks much better now 👍

cli_helper/kart.c
```diff
@@ -221,6 +232,10 @@ int main(int argc, char **argv, char **environ)
     {
         debug("enabled %s, pid=%d\n", cmd_path, getpid());
 
+        // Make this process the leader of a process group:
+        // The procress-group ID (pgid) will be the same as the pid.
+        setpgrp();
```
rcoup, line 237: If something is invoking kart (like a bash script/pipeline); will ctrl-c'ing it work? I'll do an experiment.

kart/helper.py
rcoup: Yes, the [possible error codes from `setpgid()`](https://man7.org/linux/man-pages/man2/getpgid.2.html#ERRORS) are all mapped to subclasses of `OSError`.

cli_helper/kart.c
```diff
@@ -267,8 +282,8 @@ int main(int argc, char **argv, char **environ)
         int fp = open(getcwd(NULL, 0), O_RDONLY);
         int fds[4] = {fileno(stdin), fileno(stdout), fileno(stderr), fp};
 
-        char *socket_filename = malloc(strlen(getenv("HOME")) + strlen(".kart.socket") + 2);
-        sprintf(socket_filename, "%s/%s", getenv("HOME"), ".kart.socket");
+        char *socket_filename = malloc(strlen(getenv("HOME")) + strlen(".kart.1234567890.socket") + 2);
+        sprintf(socket_filename, "%s/.kart.%d.socket", getenv("HOME"), getsid(0));
```
rcoup, line 286: `sprintf` is evil 👿 

```suggestion
        size_t socket_filename_sz = strlen(getenv("HOME")) + strlen("/.kart..socket") + sizeof(pid_t) * 3 + 1;
        char *socket_filename = malloc(socket_filename_sz);
        int r = snprintf(socket_filename, socket_filename_sz, "%s/.kart.%d.socket", getenv("HOME"), getsid(0));
        if (r < 0 || (size_t) r >= socket_filename_sz)
        {
            fprintf(stderr, "Error allocating socket filename\n");
            exit(1);  
        }
        
```

cli_helper/kart.c
```diff
@@ -209,6 +209,17 @@ void exit_on_alarm(int sig)
     exit(exit_code);
 }
 
+/**
+ * @brief Exit signal handler for SIGINT.
+ * Tries to kill the whole process group.
+ */
+void exit_on_sigint(int sig)
+{
+    putchar('\n');
```
rcoup, line 218: what's the putchar for?

olsen232: Hitting Ctrl+C often ends up with the last line of output and your PS1 prompt on the same line / mixed together. No other program seems to care about this but I like to LF out of the way

rcoup (approved):
LGTM after tidyups

kart/helper.py
olsen232: Done

cli_helper/kart.c
olsen232: Done

kart/helper.py
```diff
@@ -169,6 +176,16 @@ def helper(ctx, socket_filename, timeout, args):
...
```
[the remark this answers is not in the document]
olsen232: Done

Issue #870, opened 2023-06-14
Kart 0.14 release tracking


olsen232:
Kart 0.14 is hopefully nearly here but still has a few different issues, some of which may be release-blocking:
- [x] helper mode: sigint (Ctrl+C) not working properly: PR is here https://github.com/koordinates/kart/pull/869 - PR is tricky to get right, currently is not right
- [x] 2x disk-size bloat: copy-on-write not working at all on macos. This is due to us compiling with MAC_OS_X_VERSION_MIN_REQUIRED less than 10.12, which switches it off... either we figure out how to fix compilation to switch it on, or, we could subprocess.check_call to `cp -c` which does the same thing.
- [x] 3x disk-size bloat? check if LFS objects are leaking into git ODB somehow. Found some config issues, PR is here: https://github.com/koordinates/kart/pull/872
- [x] Apparently gdal hangs / takes ages creating VRTs sometimes - safer to flag this off

Quick fixes
- [x] Update command-line help for `kart import` to mention rasters, double-check command-line help for `kart raster-import`

Minor issues found:
- [x] is-working-copy-dirty checks are stricter than kart diff or kart status - they still report uncommitted changes even when there are only minor diffs which are suppressed
- [ ] would be nice if kart-status gave more info about conflicts (eg more than one CRS in a single dataset) than it does
- [x] `kart help import` has some `\b` in it that are confusing to `man`

PR #872, opened 2023-06-19
Fix LFS config for TIFFs, improve tests

olsen232:
This missing piece of config will cause LFS tiles to leak into the ODB since when we run `git diff` to detect changes, that either:
- adds new pointer files to the ODB, if git LFS is configured properly
- adds the raw tiles to the ODB, if git LFS is not configured properly as a filter to be applied to the edited tiles

Adds tests to make sure that tiles are not leaking into the ODB.

https://github.com/koordinates/kart/issues/870


craigds (approved):
any idea why the three changed tgz files got way smaller? suspicious

Issue #876, opened 2023-06-21
reflink not working on macos


olsen232:
The copy-on-write module, reflink, is not working at all on macos.

This is due to it being compiled with MAC_OS_X_VERSION_MIN_REQUIRED less than 10.12, which causes the native .so library generated to just be a stub.

PR #877, opened 2023-06-21
Fix for reflink - build reflink for macos11

olsen232:
## Description

Building for macos10.9 (as we were doing previously) or equivalently pip installing it with the macos python version we have, ends up with a stub version of reflink that can't actually reflink anything. 

Fixes a test that was failing due to reflink being less powerful than shutil.copy (it can't overwrite existing files, so we have to make sure the old files are out of the way).

Adds tests that actually check if things are reflinked.

https://github.com/koordinates/kart/issues/876


tests/conftest.py
```diff
@@ -1231,3 +1231,68 @@ def _check_lfs_hashes(repo, expected_file_count=None):
...
+    import reflink
+
+    from kart import is_windows
+    from kart.lfs_util import get_hash_and_size_of_file, get_local_path_from_lfs_hash
+
+    clone_checker = shutil.which("clone_checker")
+    fienode = shutil.which("fienode")
+    has_checker = bool(clone_checker or fienode)
+
+    def _check_tile_is_reflinked(tile_path, repo, do_raise_skip=False):
```
rcoup, line 1258: wrt `do_raise_skip`, why _wouldn't_ we always skip if the OS/infra doesn't support the test/feature? That's the pattern we've used elsewhere IIUC

tests/point_cloud/test_workingcopy.py
```diff
@@ -1244,3 +1246,28 @@ def test_working_copy_conflicting_extension(cli_runner, data_archive, requires_p
         r = cli_runner.invoke(["status"])
         assert r.exit_code == INVALID_OPERATION
         assert "More than one tile found in working copy with the same name" in r.stderr
+
+
+def test_working_copy_reflink(cli_runner, data_archive, check_tile_is_reflinked):
+    if is_windows:
+        # Nothing to test, so let's mark test will show as "passed".
```
rcoup, line 1253: same here...

rcoup (approved):
LGTM 👍

tests/conftest.py
olsen232: I added a couple new tests, both called `test_working_copy_reflink`.
If reflink is missing, we skip them.
However, I also used this same fixture to add some more reflink asserts to some existing import tests. Those tests should now make extra asserts - if they are able to do so - but shouldn't skip the test if they can't.

tests/point_cloud/test_workingcopy.py
olsen232: Yeah I was unsure what to do here: wasn't sure if I should try to use pytest.skip to mean "functionality is untested" and just a pass to mean "nothing to test". But I think you're right, pytest.skip makes more sense here.

PR #884, opened 2023-07-05
Protect against XSS in diff-view by using application/json

olsen232:
## Description

As discussed in https://github.com/koordinates/kart/pull/883
No extra tests yet, this is all the time I was willing to spend on it for right now.

Not clear if tests would be super valuable - we could test the following:
- does html.escape escape the characters that it says it escapes
- does json.dumps create valid json (ie, can we load it using json.loads)
but honestly it seems like these would be pretty well tested already

We can't really test the following:
- is html.escape guaranteed XSS safe for whatever browser the user has
- is json.dumps guaranteed XSS safe for whatever browser the user has (ie, it won't get tripped up by `{"xss": "</script><script>alert(1)</script>"}` or some nonsense)
- does the user's browser respect <script type="application/json"> and not allow XSS inside it? (my one does so that's good)

## Checklist:

### DIFF
--- a/kart/diff-view.html
+++ b/kart/diff-view.html
@@ -103,8 +103,9 @@
             top: 2px;
         }
     </style>
-    <script id="kart-data">const DATA=${geojson_data};</script>
+    <script id="kart-data" type="application/json">${geojson_data}</script>
     <script type="module">
+        const DATA = JSON.parse(document.getElementById('kart-data').textContent);
         const GEOM = '⭔'
 
         function buildMap() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 ## 0.14.1 (UNRELEASED)
 
 - Fixes a bug where Git subprocesses (such as git clone) don't prompt the user for credentials or to resolve SSH issues on Windows. [#852](https://github.com/koordinates/kart/issues/852)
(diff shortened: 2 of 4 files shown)


olsen232:
@enricofer Find slightly improved diff-view.html template that is more secure against XSS attacks - you can copy the changes to your own HTML template(s).

kart/diff-view.html
```diff
@@ -103,8 +103,9 @@
             top: 2px;
         }
     </style>
-    <script id="kart-data">const DATA=${geojson_data};</script>
+    <script id="kart-data" type="application/json">${geojson_data}</script>
```
craigds, line 106: what happens if geojson_data contains a string with `</script>` in it - does the nature of this json script tag ignore html in strings like that? or do we need to escape `/` characters or something?

django's `json_script` template filter might be worth checking out: https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#std-templatefilter-json_script

craigds:
```html
<!DOCTYPE html>
<html>
    <head>
        <script type="application/json">
            {
                "key": "</script><script>alert('xss')</script><script>{
            }
        </script>
    </head>
</html>
```

this alerts for me. so I think we need to escape `/` or something

olsen232: Good catch. Done, and added a test at last

craigds (approved):

Issue #892, opened 2023-07-28
Improve `kart data ls`, `kart data version`


olsen232:
Since Kart 0.12, Kart repositories are not homogenous: there is now more than one type of dataset.
Supported datasets are table (v3), point cloud (v1) and raster (v1).

The command `kart data ls` shows you a list of datasets in the current repository, but, doesn't show you what type of dataset they are, or what version. 
(By contrast, the command `kart meta get --with-dataset-types` does show you this information.)

The command `kart data version` basically always tells you "This Kart repo uses Datasets v3"
When only table datasets existed and there was a high chance of finding v1 or v2 table datasets, this command was useful. 
This command hasn't been updated to deal with new types of dataset - it always shows you the the table-dataset version is, which is a repository wide setting. Repository-wide dataset version settings is something we've since moved away from as the new dataset types were added.

Now that there are multiple types of dataset, but not currently lots of different versions of them, this command should either go away, or, show you basically exactly the same information that `kart data ls` will once it shows you the types and versions.
(Need to make sure none of the current clients on the exact output of these commands before making changes.)

PR #895, opened 2023-07-31
Some kart-data-ls improvements

olsen232:
Backwards compatible at this point:
- You need to add a flag to change kart-data-ls behaviour
- kart-data-version is now deprecated, but still works.

See https://github.com/koordinates/kart/issues/892

## Checklist:

### DIFF
--- a/kart/data.py
+++ b/kart/data.py
@@ -28,27 +28,44 @@ def data(ctx, **kwargs):
     type=click.Choice(["text", "json"]),
     default="text",
 )
+@click.option(
+    "--with-dataset-types/--without-dataset-types",
+    is_flag=True,
+    help="When set, outputs the dataset type and version. (This may become the default in a later version of Kart)",
+)
 @click.argument("refish", required=False, default="HEAD", shell_complete=ref_completer)
 @click.pass_context
-def data_ls(ctx, output_format, refish):
+def data_ls(ctx, output_format, with_dataset_types, refish):
     """List all of the datasets in the Kart repository"""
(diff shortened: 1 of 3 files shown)


kart/data.py
```diff
@@ -28,27 +28,44 @@ def data(ctx, **kwargs):
...
+    json_list = [
+        {"path": ds.path, "type": ds.DATASET_TYPE, "version": ds.VERSION}
+        for ds in repo.datasets(refish)
+    ]
+
+    if output_format == "text" and not json_list:
+        repo_desc = (
+            "Empty repository."
+            if repo.head_is_unborn
+            else "The commit at HEAD has no datasets."
```
craigds, line 50:
```suggestion
            else f"The commit at {refish} has no datasets."
```

kart/data.py
```diff
@@ -28,27 +28,44 @@ def data(ctx, **kwargs):
...
+        click.echo(f'{repo_desc}\n  (use "kart import" to add some data)')
+        return
 
     if output_format == "text":
-        if ds_paths:
-            for ds_path in ds_paths:
-                click.echo(ds_path)
+        if with_dataset_types:
+            for ds_obj in json_list:
+                click.echo(f"{ds_obj['path']}\t({ds_obj['type']}.v{ds_obj['version']})")
```
craigds, line 58: strings inside strings are hard to read, pull this out somehow. perhaps don't use an f-string, use a `.format(**ds_obj)`

craigds (approved):

PR #897, opened 2023-08-16
Don't lock annotations.db exclusively

craigds:
## Description

The annotations session (sqlite database) is always locked exclusively, since it runs `CREATE TABLE IF NOT EXISTS...` every time. This prevents concurrent use.

This PR changes that by only running it if the table doesn't exist, and in a separate session/transaction.

## Checklist:

### DIFF
--- a/kart/annotations/db.py
+++ b/kart/annotations/db.py
@@ -3,6 +3,8 @@ import json
 import logging
 import threading
 
+import sqlalchemy
+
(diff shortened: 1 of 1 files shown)


rcoup (approved):

craigds:
I'm not sure what the CI fail is about. Perhaps we aren't using our frozen versions of dependencies in the linux dev environment in github actions? But I think it's not related to this PR so I'm going to merge anyway

Issue #905, opened 2023-09-07
Tracking bug for metadata-only-import from S3 or other hosted storage


olsen232:
A yet-to-be-fully-designed flow that creates a tile-based Kart dataset that is "backed" in some sense by the original data on S3 (or similar), rather than being necessarily stored in the local Kart repo or available-on-demand from the remote Kart repo. The storage format will be almost identical to a regular tile-based Kart dataset, except that extra information will be stored about the original location where the tile was found, so it can be fetched from there if necessary.

olsen232:
Still TODO:
[ ] Rename byod-import commands
[ ] Fix `kart lfs+ gc` to clean files that were fetched from S3
[ ] Add a per-dataset "no checkout" config option
[ ] add a `remote-storage.json` meta-item that allows for S3 fetches to be redirected
[ ] add config options that allows for S3 fetches to be redirected

PR #906, opened 2023-09-07
Adds an experimental byod-point-cloud-import command

olsen232:
Imports the metadata from some PC tiles hosted on S3 without downloading them, except, right now it does download them (using boto3), since 
a) PDAL can't extract metadata from S3 without downloading the whole tile anyway and
b) boto3 is now configured to fetch from S3 during testing, but PDAL is not, and in the long run, we might not need it to anyway (see TODO below).

Still TODO: use boto3 or similar to download certain ranges of a PC file, which we can then extract metadata from, without downloading all the points. (Or fix PDAL so that it can do this).

Some things are still TBD - eg, the name of this command - but nevertheless, getting this underway will help highlight which 
obstacles and decisions still remain to be solved, and this code will not be thrown away.

## Related links:

## Checklist:

### DIFF
--- a/kart/byod/point_cloud_import.py
+++ b/kart/byod/point_cloud_import.py
@@ -1,0 +1,132 @@
+import logging
+
+import click
+
+from kart.byod.importer import ByodTileImporter
+from kart.cli_util import StringFromFile, MutexOption, KartCommand
+from kart.point_cloud.import_ import PointCloudImporter
+from kart.point_cloud.metadata_util import extract_pc_tile_metadata
+from kart.s3_util import get_hash_and_size_of_s3_object, fetch_from_s3
+
+
+L = logging.getLogger(__name__)
+
+
+@click.command("byod-point-cloud-import", hidden=True, cls=KartCommand)
+@click.pass_context
+@click.option(
+    "--message",
+    "-m",
+    type=StringFromFile(encoding="utf-8"),
+    help="Commit message. By default this is auto-generated.",
+)
+@click.option(
+    "--checkout/--no-checkout",
+    "do_checkout",
+    is_flag=True,
+    default=True,
+    help="Whether to create a working copy once the import is finished, if no working copy exists yet.",
+)
+@click.option(
+    "--replace-existing",
+    is_flag=True,
+    cls=MutexOption,
+    exclusive_with=["--delete", "--update-existing"],
+    help="Replace existing dataset at the same path.",
+)
+@click.option(
+    "--update-existing",
+    is_flag=True,
+    cls=MutexOption,
+    exclusive_with=["--replace-existing"],
+    help=(
+        "Update existing dataset at the same path. "
+        "Tiles will be replaced by source tiles with the same name. "
+        "Tiles in the existing dataset which are not present in SOURCES will remain untouched."
+    ),
+)
+@click.option(
+    "--delete",
+    type=StringFromFile(encoding="utf-8"),
+    cls=MutexOption,
+    exclusive_with=["--replace-existing"],
+    multiple=True,
+    help=("Deletes the given tile. Can be used multiple times."),
+)
+@click.option(
+    "--amend",
+    default=False,
+    is_flag=True,
+    help="Amend the previous commit instead of adding a new commit",
+)
+@click.option(
+    "--allow-empty",
+    is_flag=True,
+    default=False,
+    help=(
+        "Usually recording a commit that has the exact same tree as its sole "
+        "parent commit is a mistake, and the command prevents you from making "
+        "such a commit. This option bypasses the safety"
+    ),
+)
+@click.option(
+    "--num-workers",
+    "--num-processes",
+    type=click.INT,
+    help="How many import workers to run in parallel. Defaults to the number of available CPU cores.",
+    default=None,
+    hidden=True,
+)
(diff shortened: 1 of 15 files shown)


kart/s3_util.py
```diff
@@ -0,0 +1,84 @@
...
+    Given an s3_path with wildcard in, uses prefix and suffix matching to find all S3 objects that match.
+    """
+    # TODO: handle any kind of failure, sanity check to make sure we don't match a million objects.
+    if "*" not in source_spec:
+        yield source_spec
+        return
+    else:
+        parsed = urlparse(source_spec)
+        bucket = get_bucket(parsed.netloc)
+        prefix, suffix = parsed.path.split("*", maxsplit=1)
```
craigds, line 62: maybe worth checking that there aren't more than one `*`

kart/byod/point_cloud_import.py
```diff
@@ -0,0 +1,132 @@
...
+        amend=amend,
+        allow_empty=allow_empty,
+        sources=list(sources),
+        num_workers=num_workers,
+    )
+
+
+class ByodPointCloudImporter(ByodTileImporter, PointCloudImporter):
+    def extract_tile_metadata(self, tile_location):
+        oid_and_size = get_hash_and_size_of_s3_object(tile_location)
```
craigds, line 123: is there a more efficient way of doing this? 

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html supports a `OptionalObjectAttributes` thing, which might let us grab filesizes and checksums in the list call, to avoid doing GetObject requests for each object.

If we can get PDAL to do the PC tile metadata bit, and we can move the checksum/filesize into the list call, then we don't need to do any per-object calls ourselves, which should make things quite a bit faster for big layers

tests/conftest.py
```diff
@@ -1166,6 +1166,31 @@ def ctx(create=False):
...
+@pytest.fixture()
+def s3_test_data_point_clouds(monkeypatch_session):
+    """
+    You can run tests that fetch a copy of the auckland test data from S3 (and so test Kart's S3 behaviour)
+    by setting KART_S3_TEST_DATA_POINT_CLOUDS=s3://some-bucket/path-to-auckland-tiles/*.laz
+    The tiles hosted there should be the ones found in tests/data/point-cloud/laz-auckland.tgz
+    """
+    if "KART_S3_TEST_DATA_POINT_CLOUDS" not in os.environ:
+        raise pytest.skip(
+            "S3 tests require configuration - read docstring at conftest.s3_test_data_point_clouds"
```
craigds, line 1178: we should stick the files in a public bucket and then we can just enable these tests unconditionally

kart/s3_util.py
olsen232: Done. Also tightened up the rules about to not match a whole "tree" of content eg "*.txt" matches "foo.txt" but not "foo/bar.txt"

kart/byod/point_cloud_import.py
olsen232: Not done - we don't necessarily do a list call if the user didn't supply us with a wildcard operator to expand. We could do a list call regardless, using whatever is the common prefix of all the supplied tiles... or multiple list calls if there are multiple distinct prefixes... but, this sounds like a whole other thing, it doesn't need to be a part of this PR

tests/conftest.py
olsen232: Done on CI but still using this variable - feels like its better than hard-coding it into the python.

kart/s3_util.py
```diff
@@ -0,0 +1,106 @@
...
+    return output_path
+
+
+def expand_s3_glob(source_spec):
+    """
+    Given an s3_path with '*' wildcard in, uses prefix and suffix matching to find all S3 objects that match.
+    Subdirectories (or the S3 equivalent - S3 is not exactly a directory hierarchy) are not matched -
+    that is, s3://bucket/path/*.txt matches s3://bucket/path/example.txt but not s3://bucket/path/subpath/example.txt
+    """
+    # TODO: handle any kind of failure, sanity check to make sure we don't match a million objects.
```
craigds, line 66: meh; a million objects is actually fine at this level, not really our job to try to guess if that's the user's intent 🤷

kart/s3_util.py
```diff
@@ -0,0 +1,106 @@
...
+
+def get_hash_and_size_of_s3_object(s3_url):
+    """Returns the (SHA256-hash-in-Base64, filesize) of an S3 object."""
+    parsed = urlparse(s3_url)
+    bucket = parsed.netloc
+    key = parsed.path.lstrip("/")
+    response = get_s3_client().head_object(
+        Bucket=bucket, Key=key, ChecksumMode="ENABLED"
+    )
+    # TODO - handle failure (eg missing SHA256 checksum), which is extremely likely.
```
craigds, line 103: this seems important

kart/byod/point_cloud_import.py
craigds: turns out this thing doesn't actually work anyway because `RestoreStatus` is the only thing you can put in `OptionalObjectAttributes` - you can't ask for checksums as part of that. Dumb but can't do much about it 👍

craigds (approved):

PR #909, opened 2023-09-26
Adds an experimental byod-raster-import command...

olsen232:
Similar to the byod-point-cloud import command -
brings it to about the same level of functionality. Actually slightly better since it avoids fetching the tiles themselves, thanks to GDAL's /vsis3/ support.

Also slightly more complex: supports PAM (tif.aux.xml) files.

## Checklist:

### DIFF
--- a/tests/byod/test_imports.py
+++ b/tests/byod/test_imports.py
@@ -67,3 +67,69 @@ def test_byod_point_cloud_import(
             "oid": "sha256:6b980ce4d7f4978afd3b01e39670e2071a792fba441aca45be69be81cb48b08c",
             "size": 51489,
         }
+
+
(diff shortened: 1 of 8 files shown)


tests/byod/test_imports.py
```diff
@@ -67,3 +67,69 @@ def test_byod_point_cloud_import(
...
+
+
+@pytest.mark.slow
+def test_byod_raster_import(
+    tmp_path,
+    chdir,
+    cli_runner,
+    s3_test_data_raster,
+):
+    # Using postgres here because it has the best type preservation
```
craigds, line 79:
```suggestion
```

craigds (approved):

PR #911, opened 2023-10-02
Fix CURL_CA_INFO incorrect when using kart helper

craigds:
## Description

Kart sets `CURL_CA_INFO` at startup in `__init__.py` and this also happens in helper mode.

Unfortunately, in helper mode the environment is overwritten with the environment of the calling process, and `CURL_CA_INFO` disappears.

Then if the helper invokes PDAL or any other subprocess it doesn't get that variable.

This fix adds `CURL_CA_INFO` to the whitelist of env vars to preserve in the helper rather than override from the calling process.

## Checklist:

### DIFF
--- a/kart/helper.py
+++ b/kart/helper.py
@@ -10,6 +10,7 @@ from pathlib import Path
 
 import click
 
+from . import HELPER_PRESERVE_ENV_VARS
 from .socket_utils import recv_json_and_fds
 from .cli import load_commands_from_args, cli, is_windows, is_darwin, is_linux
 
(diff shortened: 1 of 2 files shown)


kart/helper.py
```diff
@@ -74,23 +74,24 @@ def helper(ctx, socket_filename, timeout, args):
...
+            "GDAL_DATA",
             "GIT_CONFIG_NOSYSTEM",
             "GIT_EXEC_PATH",
-            "GIT_TEMPLATE_DIR",
+            "GIT_EXEC_PATH",
             "GIT_INDEX_FILE",
-            "GDAL_DATA",
+            "GIT_INDEX_FILE",
+            "GIT_TEMPLATE_DIR",
+            "GIT_TEMPLATE_DIR",
```
rcoup: dupe

olsen232 (approved):

rcoup (approved):

Issue #917, opened 2023-10-06
Kart help displays / endorses too many commands


olsen232:
There are kart internal commands - some of which may not be safe to use for the uninitiated - that are marked in the source as "hidden=True" but these are showing up in our help pages, which defeats the purpose of us marking them as hidden.

For example `kart --help` includes `kart git` in the list - and `kart git --help` doesn't mention anything about it being hidden / internal only.

PR #922, opened 2023-10-12
Properly hide "hidden" commands from kart help

olsen232:
Document more about why `kart git` is dangerous.
Unhide `kart commit-files` since it's a safer alternative to various things you'd need `kart git` for otherwise: amending commits and attaching files.

## Related links:

https://github.com/koordinates/kart/issues/917

## Checklist:

### DIFF
--- a/kart/meta.py
+++ b/kart/meta.py
@@ -215,7 +215,7 @@ def meta_set(ctx, message, amend, dataset, items):
     )
 
 
-@click.command("commit-files", hidden=True, cls=KartCommand)
+@click.command("commit-files", cls=KartCommand)
 @click.option(
     "--message",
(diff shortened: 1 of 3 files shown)


kart/meta.py
```diff
@@ -215,7 +215,7 @@ def _parse(key, value):
     )
 
 
-@click.command("commit-files", hidden=True, cls=KartCommand)
+@click.command("commit-files", cls=KartCommand)
```
craigds, line 218: I'd leave it hidden because I don't think we have a real use case for it

Instead we should add an `--amend` option to `commit`; i'm not really sure why we don't have that yet

craigds (approved):

PR #926, opened 2023-10-19
Improve existing --no-checkout flag

olsen232:
kart import supports --no-checkout, but it merely postpones creating a working copy. As of this commit, it will add also add the dataset to a set of non-checkout datasets in the config, so the dataset will stay out of the WC indefinitely.

Also adds `kart checkout --dataset=foo --not-dataset=bar` flags to add and remove datasets from the set of non-checkout datasets

Still TODO: show what the current set of non-checkout datasets is in `kart status`

## Related links:

## Checklist:

### DIFF
--- a/kart/repo.py
+++ b/kart/repo.py
@@ -592,7 +592,11 @@ class KartRepo(pygit2.Repository):
         result = set()
         config = self.config
         for entry in config:
-            parts = entry.name.split(".")
+            parts = entry.name.split(".", maxsplit=3)
+            if len(parts) > 3:
+                # Handle a name-containing-dots ie "dataset.NAME.CONTAINING.DOTS.checkout"
+                prefix, rest = entry.name.split(".", maxsplit=1)
+                parts = [prefix, *rest.rsplit(".", maxsplit=1)]
             if (
                 len(parts) == 3
                 and parts[0] == "dataset"
--- a/kart/repo.py
+++ b/kart/repo.py
@@ -577,6 +577,31 @@ class KartRepo(pygit2.Repository):
 
         return SpatialFilter.from_repo_config(self)
 
+    def configure_do_checkout_datasets(self, dataset_paths, do_checkout):
+        for dataset_path in dataset_paths:
+            key = f"dataset.{dataset_path}.checkout"
+            if do_checkout:
+                # Checking out a dataset is the default, we don't clutter the config with it.
+                self.del_config(key)
+            else:
+                # Specifically mark this dataset as do-not-checkout.
+                self.config[key] = False
+
+    @property
+    def non_checkout_datasets(self):
+        result = set()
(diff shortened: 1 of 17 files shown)


kart/repo.py
```diff
@@ -577,6 +577,31 @@ def spatial_filter(self):
...
+            else:
+                # Specifically mark this dataset as do-not-checkout.
+                self.config[key] = False
+
+    @property
+    def non_checkout_datasets(self):
+        result = set()
+        config = self.config
+        for entry in config:
+            parts = entry.name.split(".")
```
craigds, line 595: hm, paths can contain dots (`a.b.c` is a valid path) - should this config structure be adjusted?

olsen232: Good catch - tested config files and pygit2 behaviour, dots are valid in the middle part (which seems to be the "user-defined" part) but not in the first and last part.
Updated code to recognise and allow for entries that look like `dataset.SOME.NAME.WITH.DOTS.checkout`

craigds: wild. ok

craigds (approved):

kart/repo.py
rcoup: Sorry, I didn't get to this earlier. Git config has standard ways of referring to weird section names, this looks like it's using the older deprecated dot notation?

eg:
```
[includeIf "gitdir:/path/to/foo/.git"]
	path = /path/to/foo.inc
```

https://git-scm.com/docs/git-config#_syntax

> Sections can be further divided into subsections. To begin a subsection put its name in double quotes, separated by space from the section name, in the section header, like in the example below:
> 
> ```
> [section "subsection"]
> ```
>
> Subsection names are case sensitive and can contain any characters except newline and the null byte. Doublequote " and backslash can be included by escaping them as \" and \\, respectively. Backslashes preceding other characters are dropped when reading; for example, \t is read as t and \0 is read as 0. Section headers cannot span multiple lines. Variables may belong directly to a section or to a given subsection. You can have [section] if you have [section "subsection"], but you don’t need to.
>
> There is also a deprecated `[section.subsection]` syntax. With this syntax, the subsection name is converted to lower-case and is also compared case sensitively. These subsection names follow the same restrictions as section names.

olsen232: 👍 Yup this is fine - in fact I'm using the pygit2 API to access the config (although it's not immediately obvious that I'm doing so from this code). pygit2's API converts git's more complicated syntax`[section "subsection"] key = value` into a mostly-simpler global key that looks like `section.subsection.key` - and as of Craig's comment, I tested to see if it can handle subsections with dots in, eg `[section "subsection.with.dots"] key = value` - it can, and for API access it converts it again to `section.subsection.with.dots.key` which I have handled now, it's straightforward as long as you know that that's what it will do.

PR #928, opened 2023-10-29
Improve point-cloud schema extraction

olsen232:
- Use the real names and types of fields from the spec including, fix the incorrect sizes of <1 byte fields
- Report individual flags as single bit fields (more like new PDAL does)
- but don't actually use PDAL to learn the schema (since then we run into issues with the <1 byte field / at risk of future changes)
- always return 1 of 7 hard-coded schemas if there is no extra bytes VLR (handles backwards incompatibility)
- add a test to make sure we can read the extra bytes VLR if present

## Related links:

## Checklist:


docs/pages/development/pointcloud_v1.rst
```diff
@@ -87,103 +87,135 @@ For example, this is the schema of a dataset using "PDRF 7":
...
-        "size": 64
+        "dataType": "integer",
+        "size": 32
       },
       {
         "name": "Intensity",
         "dataType": "integer",
-        "size": 16
+        "size": 16,
+        "unsigned": true
```
craigds, line 107: i'd personally prefer `signed` rather than `unsigned` because double negatives are confusing, but I don't feel strongly about it

kart/point_cloud/metadata_util.py
```diff
@@ -307,3 +320,22 @@ def extract_format(tile_format):
...
+
+def find_vlr(metadata, user_id, record_id):
+    for key, value in metadata.items():
+        if not key.startswith("vlr"):
+            continue
+        if value["user_id"] == user_id and value["record_id"] == record_id:
+            return base64.b64decode(value["data"])
+
+
+def _remove_nones(dict_to_modify):
```
craigds, line 337: I'm generally against functions that modify arg in-place _and_ return it - do one or the other to avoid causing confusion

This function's a natural fit for returning a dict comprehension.

kart/point_cloud/schema_util.py
```diff
@@ -113,33 +150,26 @@ def equivalent_copc_pdrf(pdrf):
...
-    "floating": "float",
-    "unsigned": "integer",
-    "string": "text",
+def _vlr_type_to_kart_type(vlr_datatype, options):
+    if vlr_datatype == 0:
+        return {"dataType": "blob", "length": options}
+    if vlr_datatype <= 10:
+        return _VLR_TYPE_TO_KART_TYPE[vlr_datatype]
+    array_length = int((vlr_datatype + 9) / 10)
+    vlr_datatype = (vlr_datatype - 1) % 10 + 1
```
craigds, line 159: (a) this is confusing to reason about
(b) types >30 are actually not supported by the spec (not used as arrays) so this seems incorrect
(c) the double/triple array thing was deprecated in LAS 1.4 R14: https://github.com/ASPRSorg/LAS/issues/1

I'd suggest we drop it, YAGNI. perhaps just assert that datatype <= 10

craigds (approved):

docs/pages/development/pointcloud_v1.rst
olsen232: Not done - background: table schemas just have "signed" values but they are not marked as anything. Raster datasets added unsigned values which are now explicitly marked, so we ended up with a similar system to, say, C: integers are signed unless they are marked "unsigned".

kart/point_cloud/metadata_util.py
olsen232: Done

kart/point_cloud/schema_util.py
olsen232: Good point, done.

docs/pages/development/pointcloud_v1.rst
craigds: fair enough 👍

PR #942, opened 2023-11-16
Fix import-from-s3 for sidecar files.

olsen232:
Fix `kart import s3://bucket/path-to-tile.tif` so that it finds any PAM (.aux.xml files) that may exist for the source TIFFs.
Adds raster tests, with a focus on importing rasters with a PAM file (the rest of the import flow is basically identical to point-cloud so testing the PAM file is most important).

## Related links:

## Checklist:

### DIFF
--- a/kart/tile/importer.py
+++ b/kart/tile/importer.py
@@ -39,7 +39,7 @@ from kart.lfs_util import (
 )
 from kart.list_of_conflicts import ListOfConflicts
 from kart.meta_items import MetaItemFileType
-from kart.s3_util import expand_s3_glob, fetch_from_s3
+from kart.s3_util import expand_s3_glob, fetch_from_s3, get_error_code
 from kart.progress_util import progress_bar
 from kart.output_util import (
     format_json_for_output,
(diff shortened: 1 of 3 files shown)


kart/tile/importer.py
```diff
@@ -447,9 +448,19 @@ def extract_tile_metadata(self, source):
...
             tile_path = local_path
+            # Also fetch sidecar files, if present.
+            for sidecar_file, suffix in self.sidecar_files(source):
+                try:
+                    fetch_from_s3(
+                        sidecar_file,
+                        local_path.with_name(local_path.name + suffix),
+                    )
+                except botocore.exceptions.ClientError as e:
+                    pass  # This kind of sidecar file doesn't exist for this tile. That's okay.
```
craigds, line 459: we should probably check what the error is; don't want to silently ignore bad things. 

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html 

Also not sure what specific error codes we should look for here

craigds (approved):

PR #944, opened 2023-11-16
contrib: add experimental/proof of concept OGR drivers

rcoup:
Experimental [GDAL/OGR](https://gdal.org) plugin drivers for Kart repositories. These are mostly to provide a proof of concept for discussion and to support future feature development in Kart.

They are implemented using the [OGR Python Drivers](https://gdal.org/development/rfc/rfc76_ogrpythondrivers.html#rfc-76-ogr-python-drivers) mechanism in GDAL which is fairly rough & inefficient. It also means the drivers are read-only.

These plugins require GDAL v3.8.0 or newer. The `contrib/gdal/README.md` file contains usage information.

### CLI (`contrib/gdal/cli/`)

This plugin calls Kart CLI commands to find and expose features/layers, and diffs between revisions. Since Kart _uses_ GDAL this approach provides a safe and relatively clean process boundary. It's similar to how VSCode, Github Desktop, and other IDEs interact with Git — and how it's used server-side for many operations at the likes of Github & Gitlab. As long as the interprocess communication is efficient using this approach could be a relatively smooth path to implement a fully-featured C++ GDAL/OGR driver in the future.

### In-process (`contrib/gdal/in_process/`)

The in-process plugin uses Kart internal/private Python APIs to find and expose features/layers. Since Kart uses GDAL and GDAL is calling Kart, you can only use the GDAL library, tools, and associated drivers bundled with Kart. 

## Checklist:

### DIFF
--- a/contrib/gdal/README.md
+++ b/contrib/gdal/README.md
@@ -1,0 +1,97 @@
+GDAL/OGR Plugins
+================
+
+Experimental [GDAL/OGR](https://gdal.org) plugin drivers for Kart repositories. These are mostly to provide a proof of concept for discussion and to support future feature development in Kart.
+
+They are implemented using the [OGR Python Drivers](https://gdal.org/development/rfc/rfc76_ogrpythondrivers.html#rfc-76-ogr-python-drivers) mechanism in GDAL which is fairly rough & inefficient. It also means the drivers are read-only.
+
+> **ℹ️** These plugins require GDAL v3.8.0 or newer.
+
+CLI
+---
+
+The CLI plugin calls Kart CLI commands to find and expose features/layers, and diffs between revisions. Since Kart _uses_ GDAL this provides a relatively clean process boundary, similar to how VSCode, Github Desktop, and other IDEs interact with Git — and how it's used server-side for many forge operations at the likes of Github & Gitlab.
+
+As long as the interprocess communication is efficient using this mechanism should be a relatively smooth path to implement a fully-featured GDAL/OGR driver in the future.
+
+### Usage
+
+```console
+# Tell GDAL where to find the Kart driver
+$ export GDAL_PYTHON_DRIVER_PATH=/path/to/kart/contrib/gdal/cli
+
+# KART should appear at the bottom of the list
+$ ogrinfo --formats
+
+# list datasets
+$ ogrinfo /path/to/kart/repo
+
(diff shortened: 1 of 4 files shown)


contrib/gdal/cli/gdal_kart.py
```diff
@@ -0,0 +1,636 @@
...
+                            gdal.Error(
+                                f"KART !__iter__() {ex}\n{traceback.format_exc()}"
+                            )
+                            raise
+            gdal.Debug("KART", f"__iter__() yielded {count} features")
+        except OSError as ex:
+            gdal.Error(f"KART !__iter__() {ex}")
+            raise
+
+    # @gdal_api_wrapper
```
olsen232, line 292: Add a comment explaining the commented out code

contrib/gdal/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+In-Process
+----------
+
+The in-process plugin uses Kart internal Python APIs to find and expose features/layers. Since Kart uses GDAL and GDAL is calling Kart, you can only use the GDAL library, tools, and associated drivers bundled with Kart.
+
+### Usage
+
+This currently only works from a local build tree in `/path/to/kart/build/`. See [CONTRIBUTING](../../CONTRIBUTING.md) for details. GDAL v3.8.0 isn't currently built with Kart currently, so this will require some changes in `vcpkg-vendor/` to make it work.
```
olsen232, line 78: This comment is now slightly out of date - could change it to a note about how only the most recent Kart builds use GDAL 3.8 or could just remove it since the GDAL change at least made it in before this change, so not much chance that anyone will try to run this against an older Kart somehow.

olsen232 (approved):

contrib/gdal/README.md
rcoup: You still need to build it with ogr2ogr/ogrinfo/etc tools as well, which we don't by default. Maybe I could add a CMake option to do that.

contrib/gdal/cli/gdal_kart.py
olsen232: Removed commented out code

contrib/gdal/README.md
olsen232: Updated comment to explain the current state - will work on adding those tools next.

PR #949, opened 2023-11-23
Various wheel/python fixes

rcoup:
## Description

* sync versions between our requirements files and the wheels we're building
* upgrade boto3 to reduce size of botocore by 50+MB
* remove greenlet: it's only (now?) needed for asyncio support in SQLAlchemy, which we don't use. We haven't been good about keeping it windows-only, since pip-compile drops the `;os_name!="posix"` constraint.
* cryptography build fixes

I think dependabot might be fiddling the values in `requirements/vendor-wheels.txt`, but we need to reject those changes since they are driven from `vcpkg-vendor/CMakeLists.txt`. Either we can make it ignore them, or we could _verify_ during builds so that CI will fail for those Dependabot PRs.

Fixing up/clarifying what we're doing with pyodbc is out of scope for this change, but I think we've slipped away from how it was originally working wrt #395 

## Checklist:

- Have you included test(s)?
- Have you updated the [changelog](https://github.com/koordinates/kart/blob/master/CHANGELOG.md)?


requirements/dev.txt
```diff
@@ -1,11 +1,13 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.12
```
olsen232, line 2: ... why does the script include the python version in here? Is it likely to be give different output depending on which python version used?
Anyway vcpkg is using python 3.11 right now so I've been using that everywhere - including to generate these files - in case it matters. But I suspect it doesn't

olsen232 (approved):

requirements/dev.txt
rcoup: I think for what we're doing it most likely doesn't... but I wonder if we can encourage CMake to pick up Py3.11 if it's available during config.

PR #955, opened 2023-12-05
[WIP] point-clouds: generate virtual point cloud (.vpc)

rcoup:
Proof of concept generating a [Virtual Point Cloud (VPC)](https://www.lutraconsulting.co.uk/blog/2023/06/08/virtual-point-clouds/) file for datasets based on tiles present in the current working copy. The VPC can be opened in eg: QGIS rather than loading individual tiles, and works similarly to VRT files for rasters.

1. Currently uses `pdal_wrench` to generate the files, and we're not building/distributing that, needs to be in the PATH.
2. You also need to set `KART_POINT_CLOUD_VPCS=1`

Still need to decide either to bundle `pdal_wrench`, or write minimal VPC files ourselves.

## Related links:

* https://www.lutraconsulting.co.uk/blog/2023/06/08/virtual-point-clouds/
* https://github.com/PDAL/wrench/#virtual-point-clouds-vpc

## Checklist:

### DIFF
--- a/kart/point_cloud/mosaic_util.py
+++ b/kart/point_cloud/mosaic_util.py
@@ -1,0 +1,38 @@
+import logging
+
+from tempfile import NamedTemporaryFile
+from kart.subprocess_util import check_output
+
+
+L = logging.getLogger("kart.point_cloud")
+
+
+def write_vpc_for_directory(directory_path):
+    """
+    Given a folder containing some LAZ/LAS files, write a mosaic file that
+    combines them all into a single Virtual Point Cloud (VPC). The VPC will
+    contain references to the tiles, rather than replicating their contents.
+    """
+    vrt_path = directory_path / f"{directory_path.name}.vpc"
+
+    tiles = [str(p) for p in directory_path.glob("*.copc.laz")]
+    if not tiles:
+        vrt_path.unlink(missing_ok=True)
+        return
+
+    with NamedTemporaryFile("w+t", suffix=".kart_tiles", encoding="utf-8") as tile_list:
+        tile_list.write("\n".join(tiles))
+        tile_list.flush()
+
+        try:
+            check_output(
+                [
+                    "pdal_wrench",
+                    "build_vpc",
+                    f"--output={vrt_path}",
+                    f"--input-file-list={tile_list.name}",
+                ]
+            )
+        except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


kart/point_cloud/mosaic_util.py
```diff
@@ -0,0 +1,38 @@
...
+    contain references to the tiles, rather than replicating their contents.
+    """
+    vrt_path = directory_path / f"{directory_path.name}.vpc"
+
+    tiles = [str(p) for p in directory_path.glob("*.copc.laz")]
+    if not tiles:
+        vrt_path.unlink(missing_ok=True)
+        return
+
+    with NamedTemporaryFile("w+t", suffix=".kart_tiles", encoding="utf-8") as tile_list:
```
olsen232, line 23: Haven't tested it, but whenever I see this kind of thing I think "this will probably break on windows..." but it depends, windows can certainly share an open file between two processes if the processes feel like it, but seems like sometimes they don't.
Anyway if it does break on windows, this will be why.

olsen232:
> At some point we expect to support generating STAC indexes for datasets in Kart. Given VPCs are based on STAC, and the relative simplicity of the format, maybe we should generate them ourselves.

I was going to say the opposite - let's just bundle pdal-wrench, seems simpler (unless we hit some difficulty with that approach).
We can always switch approach later if none of the tools we already bundle / are okay to bundle do everything we want them to.

olsen232 (commented):
Two things left to do
- make this actually work by bundling pdal_wrench, or some other solution. We can still leave it turned off by default if we're not confident about it yet, but better to have a feature you can turn on and test rather than something where we have to install it to test it and then who knows if it's working the same for everyone (like we had with PDAL for a while)
- add a test - there's two very basic tests for VRTs:
`test_working_copy_vrt`, `test_working_copy_vrt_disabled`

hamishcampbell:
> Two things left to do
> 
>     * make this actually work by bundling pdal_wrench, or some other solution. We can still leave it turned off by default if we're not confident about it yet, but better to have a feature you can turn on and test rather than something where we have to install it to test it and then who knows if it's working the same for everyone (like we had with PDAL for a while)
> 
>     * add a test - there's two very basic tests for VRTs:
>       `test_working_copy_vrt`, `test_working_copy_vrt_disabled`

Ok, if bundling is straight forward that seems like a sensible step. It has a few other useful functions we might want in future too.

PR #962, opened 2024-01-30
Faster diffs: Don't always run pygit2.Tree.diff_tree

olsen232:
Don't run it when the user requests the diff of a particular feature
- we can look up the feature for both revs quicker than pygit2 can find all diffs in the entire revision.

There will be an extremely great speedup in the extreme case where the user wants to see a single feature that has changed, across a revision-range where a million features have changed.

If the user requests enough features, eventually there will be a crossover where this code path is slower. Eg if the user requests to see more features than have actually changed. However, mostly the size of the list of features requested will be much smaller than the size of the layer / size of the changes.

## Checklist:

### DIFF
--- a/kart/dataset_mixins.py
+++ b/kart/dataset_mixins.py
@@ -1,3 +1,6 @@
+from collections.abc import Iterable
+from typing import Callable, Optional
+
 import pygit2
 
 from kart.diff_structs import DatasetDiff, DeltaDiff, Delta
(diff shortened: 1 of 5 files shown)


kart/dataset_mixins.py
```diff
@@ -160,6 +166,66 @@ def get_decoder(dataset, method_name):
     _INSERT_UPDATE = _INSERT_TYPES + _UPDATE_TYPES
     _UPDATE_DELETE = _UPDATE_TYPES + _DELETE_TYPES
 
+    def get_raw_deltas_for_keys(
+        self, other, key_encoder_method, key_filter, reverse=False
```
craigds, line 170: type hints on args wouldn't go amiss, the names don't make them very obvious

(we should probably be moving towards type hints everywhere)

tests/test_diff.py
```diff
@@ -1581,7 +1582,17 @@ def test_show_points_HEAD(output_format, data_archive_readonly, cli_runner):
         ["show", "HEAD", "--", "nz_pa_points_topo_150k:1182"],
     ],
 )
-def test_diff_filtered_text(diff_command, data_archive_readonly, cli_runner):
+def test_diff_filtered_text(
+    diff_command, data_archive_readonly, cli_runner, monkeypatch
+):
+    def _get_raw_diff_for_subtree(self, *args, **kwargs):
+        # When only nz_pa_points_topo_150k:1182 is requested, a different more efficient code-path should be used.
+        raise AssertionError(
```
craigds, line 1590:
```suggestion
        pytest.fail(
```

craigds (approved):

kart/dataset_mixins.py
olsen232: Done - added type hints to this file

tests/test_diff.py
olsen232: Done

Issue #965, opened 2024-02-06
Unexpected behaviour in kart add-dataset @ postgis


FelixEickeler:
**Describe the bug**
When executing add-dataset on a table in a postgres working copy a ProgrammingError is thrown. 

**To Reproduce**
Steps to reproduce the behaviour:
1.  kart init --workingcopy postgresql://user:password@host:5432/database/schema --import ... (an actually empty gpkg)
2. Create a new table in the schema (f.e. with QGIS)
2. kart status --list-untracked-tables (interestingly this shows also the tracked table)
3. kart add-dataset shown name with schema (testet alternatives without => not found | postgres URI => not found)

**Expected behaviour**
 This works as expected when using geopackages. New tables can be added via add-dataset command.

**Output**
Traceback (most recent call last):
  File "kart_cli.py", line 4, in <module>
  File "kart\cli.py", line 404, in entrypoint
  File "click\core.py", line 1157, in __call__
  File "click\core.py", line 1078, in main
  File "kart\cli_util.py", line 72, in invoke
  File "click\core.py", line 1688, in invoke
  File "click\core.py", line 1434, in invoke
  File "click\core.py", line 783, in invoke
  File "click\decorators.py", line 34, in new_func
  File "kart\add_dataset.py", line 71, in add_dataset
  File "kart\tabular\working_copy\base.py", line 639, in meta_items
  File "kart\sqlalchemy\adapter\base.py", line 148, in all_v2_meta_items
  File "kart\utils.py", line 25, in wrapper
  File "kart\sqlalchemy\adapter\postgis.py", line 137, in all_v2_meta_items_including_empty
  File "sqlalchemy\orm\session.py", line 1744, in scalar
  File "sqlalchemy\orm\session.py", line 1714, in execute
  File "sqlalchemy\engine\base.py", line 1705, in _execute_20
  File "sqlalchemy\sql\elements.py", line 334, in _execute_on_connection
  File "sqlalchemy\engine\base.py", line 1572, in _execute_clauseelement
  File "sqlalchemy\engine\base.py", line 1943, in _execute_context
  File "sqlalchemy\engine\base.py", line 2124, in _handle_dbapi_exception
  File "sqlalchemy\util\compat.py", line 211, in raise_
  File "sqlalchemy\engine\base.py", line 1900, in _execute_context
  File "sqlalchemy\engine\default.py", line 736, in do_execute
[...]

PR #967, opened 2024-02-09
Fix add-dataset for PostGIS working copies

olsen232:
Unlike for GPKGs, we need to be looking for tables only in the appropriate schema. Happily, the code is all mostly in-place.
Added test.

## Related links:

Fixes https://github.com/koordinates/kart/issues/965

## Checklist:


craigds (approved):
LGTM apart from the test fail

PR #987, opened 2024-05-02
Gdal tools

olsen232:
## Description

Bundling gdal tools with Kart - particularly concerned with ogrinfo and ogr2ogr, but included them all.

Note that:
- executables work on macos and linux as long as the rpath is set properly (it is).
- executables work on windows as long as they are in the same directory as the DLLs they need, or, a program is run before hand that adds the relevant lib directory to the DLL path (ie, Kart).

With these two facts in hand, I made the following design choices for the various environments we create:
### Development environment aka venv aka temporary environment during building:
The gdal tools always go in $PREFIX/tools/gdal, ie, venv/tools/gdal
This means that:
- they work on all platforms when run from within Kart
- they don't work on Windows when run standalone
- they're kept separate from the other files - which means we can easily find them all during the pyinstaller step by doing a globbing rule

### Pyinstaller aka release environment
The gdal tools still go in $PREFIX/tools/gdal on macos and linux
They go in $PREFIX on windows
This means that:
- they work on all platforms when run from Kart or run standalone
- on windows they're cluttered in with all our other files, but this is a tradeoff I made to have them work standalone. Inevitably we'll end up telling some windows user to run them directly for debugging purposes etc.

## Checklist:

### DIFF
--- a/kart.spec
+++ b/kart.spec
@@ -84,7 +84,10 @@ binaries = [
     (f'{BINARY_DIR}/venv/lib/mod_spatialite.{lib_suffix_glob}', '.'),
     (f'{BINARY_DIR}/venv/{VENV_BIN_DIR}/git-lfs{exe_suffix}', '.'),
     (f'{BINARY_DIR}/venv/{VENV_BIN_DIR}/pdal{exe_suffix}', '.'),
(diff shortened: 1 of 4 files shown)


rcoup (approved):
LGTM 👍

PR #988, opened 2024-05-03
Basic export command

olsen232:
Uses python feature API + OGR drivers,
rather than the Kart python driver in contrib/

Still documented as experimental, marked as hidden, not yet in CHANGELOG.
Will document more and add to an upcoming release once tested a bit more.

## Checklist:


kart/tabular/export.py
```diff
@@ -0,0 +1,306 @@
...
+@click.option(
+    "--ref", default="HEAD", help="The revision at which to export the dataset."
+)
+@click.option(
+    "--layer",
+    help="Name of the layer to create inside the given destination. Defaults to the dataset path.",
+)
+@click.option(
+    "--layer-creation-option",
+    "-lco",
```
craigds, line 96: do these short options work with click? Cool if so - but I kind of expected it to condence multiple short single-leter options into one argument (`ls -lha` style)

kart/tabular/export.py
```diff
@@ -0,0 +1,306 @@
...
+        if d.GetName().upper() == shortname:
+            if not is_writable_vector_format(d):
+                raise click.UsageError(
+                    f"Driver {shortname} does not support writing vectors"
+                )
+            return d
+    raise click.UsageError(f"Destination format {shortname} not recognized")
+
+
+def get_driver_by_ext(filename):
```
craigds, line 43: with all these things it is possible that there may not be enough info to choose a format - e.g. a user might want to export to a `.tsv` file using the CSV driver. Also there are definitely multiple drivers in GDAL which can export the same extensions in different ways

anyway, it's probably fine for now since our GDAL is pretty minimal, but we might one day need a `--format` option to explicitly set the format

kart/tabular/export.py
```diff
@@ -0,0 +1,306 @@
...
+        out_feature = ogr.Feature(out_layer.GetLayerDefn())
+        if primary_key_as_fid:
+            out_feature.SetFID(feature[pk_name])
+        if geom_key:
+            geom = feature[geom_key]
+            out_feature.SetGeometry(geom.to_ogr() if geom else None)
+        for i, key in enumerate(regular_keys):
+            out_feature.SetField(i, feature[key])
+        out_layer.CreateFeature(out_feature)
+        out_feature = None
```
craigds, line 222: i know the OGR docs and various sample code does this all over the place, but reference counts are reference counts - there's not much point setting a thing to None as it is about to go out of scope anyway

kart/tabular/export.py
```diff
@@ -0,0 +1,306 @@
...
+def kart_col_to_ogr_field(col):
+    ogr_type = None
+    size = col.get("size")
+    if size is not None:
+        ogr_type = OGR_DATATYPE_MAP.get((col.data_type, size))
+    if ogr_type is None:
+        ogr_type = OGR_DATATYPE_MAP[col.data_type]
+    assert ogr_type is not None
+
+    if isinstance(ogr_type, tuple):
```
craigds, line 236: just make all the values in OGR_DATATYPE_MAP tuples with subtypes None where there's no subtype

craigds (approved):
LGTM

kart/tabular/export.py
```diff
@@ -0,0 +1,306 @@
...
+def list_export_formats(ctx):
+    """List the supported export formats as reported by gdal."""
+
+    for i in range(ogr.GetDriverCount()):
+        d = ogr.GetDriver(i)
+        if is_writable_vector_format(d):
+            click.echo(f"    {d.GetName()}: {d.GetMetadataItem(gdal.DMD_LONGNAME)}")
+
+
+def get_driver_by_shortname(shortname):
```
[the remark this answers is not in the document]
olsen232: Ah oops. I for whatever reason assumed it would be case sensitive, so I just wrote my own.
Using the built in one now.

kart/tabular/export.py
olsen232: They do work! Perhaps the other sort `-lha` don't work? Or perhaps it's really clever

kart/tabular/export.py
olsen232: I support `kart export my-dataset CSV:output.tsv` which would work fine for the usecase above

kart/tabular/export.py
olsen232: Ok, removed the superfluous `foo = None` lines

olsen232: Oops. Removing `out_ds = None` causes tests to fail. Added it back again :-/

kart/export.py
```diff
@@ -0,0 +1,41 @@
+import click
+
+from kart.completion_shared import repo_path_completer
+from kart.cli_util import KartCommand, forward_context_to_command
+
+
+# Aliases `kart export` to `kart table-export`.
+# One day we might support other types of export, and when we do, this will then
+# be a general purpose command that delegates to one or other more specific commands.
+# See: how `kart import` delegates to one of table-import, point-cloud-import, raster-import.
```
rcoup, line 10: Some formats can do vectors and rasters (eg: GeoPackage, FileGDB)... how can/should we deal with that in future?

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+    help="Include the primary key column as a regular field that is sent to the GDAL driver.",
+)
+@click.option(
+    "--primary-key-as-fid/--no-primary-key-as-fid",
+    is_flag=True,
+    default=False,
+    help="Report to the GDAL driver that the FID of each feature is the primary key value.",
+)
+@click.option(
+    "--override-geometry",
```
rcoup, line 116: `--override-geometry-type`?

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+)
+@click.option(
+    "--drop-null-geometry-features",
+    is_flag=True,
+    help="Skips export of those features where the geometry is null.",
+)
+@click.argument(
+    "args",
+    nargs=-1,
+    metavar="DATASET [EXPORT_TYPE:]DESTINATION",
```
rcoup, line 127: `DATASET... [EXPORT_TYPE]:DESTINATION`? 

How do we export all datasets in a repo?

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+        # correspond to 'precision' and 'scale' in most other systems, respectively:
+        if precision:
+            result.SetWidth(precision)
+        if scale:
+            result.SetPrecision(scale)
+
+    return result
+
+
+OGR_DATATYPE_MAP = {
```
rcoup, line 261: Don't we have this defined elsewhere for imports?

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+
+def is_writable_vector_format(driver):
+    m = driver.GetMetadata()
+    return m.get("DCAP_VECTOR") == "YES" and m.get("DCAP_CREATE") == "YES"
+
+
+def list_export_formats(ctx):
+    """List the supported export formats as reported by gdal."""
+
+    for i in range(ogr.GetDriverCount()):
```
rcoup, line 24: since GDAL 2.0 [vector and raster drivers are unified](https://gdal.org/development/rfc/rfc46_gdal_ogr_unification.html) (to support mixed formats); we should be using `gdal.*` objects/methods for drivers & datasets if we're writing new code.

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+    try:
+        dataset = repo.datasets(ref)[ds_path]
+    except KeyError:
+        raise click.UsageError(f"No such dataset: {ds_path}")
+
+    if dataset.DATASET_TYPE != "table":
+        raise click.UsageError(f"Dataset {ds_path} is not a vector or tabular dataset")
+
+    driver, destination = get_driver(destination_spec)
+    out_ds = driver.CreateDataSource(destination, options=list(dataset_creation_option))
```
rcoup, line 169: I'm unclear what this would do if the output already exists? I guess to allow multiple datasets to end up in the same FGDB/GPKG/etc then we should generally allow this.

For some formats the output is a directory (shapefiles); for others it's a file (GPKG), which complicates things a bit.

We don't want the working copy overwritten, that would be bad.

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+                f"Sorry, multiple CRS definitions at {ds_path!r} are not yet supported for export"
+            )
+        if crs_defs:
+            ogr_crs = make_crs(crs_defs[0], context=ds_path)
+
+        out_layer = out_ds.CreateLayer(
+            layer_name, ogr_crs, ogr_geom_type, options=list(layer_creation_option)
+        )
+    else:
+        out_layer = out_ds.CreateLayer(layer_name, options=list(layer_creation_option))
```
rcoup, line 202: Similarly here, do we always error if the layer exists already?

tests/test_export.py
rcoup, line 1: Should add some more test cases:

* using dataset creation options
* using layer creation options
* gdal config options (via env)
* existing directories, existing files, etc
* maybe a set of common formats? (SHP, FGDB, GeoJSON, CSV)
* truncated/fixed/clashing field names (SHP in particular)
* error handling / error messages

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+    if match:
+        return get_driver_by_shortname(match.group(1)), destination_spec
+    if ":" in destination_spec:
+        shortname, destination = destination_spec.split(":", maxsplit=1)
+        return get_driver_by_shortname(shortname), destination
+    return get_driver_by_ext(destination_spec), destination_spec
+
+
+@click.command(
+    "table-export",
```
rcoup, line 73: Missing option - CRS? Exporting into a non-native CRS seems like a fairly common use case.

kart/tabular/export.py
```diff
@@ -0,0 +1,304 @@
...
+    drop_null_geometry_features,
+    args,
+    **kwargs,
+):
+    """
+    Experimental export command - exports a tabular kart dataset at a particular commit.
+
+    Uses GDAL's OGR drivers to do so - consult https://gdal.org/drivers/vector/index.html
+    to find the options specific to a particular driver, which can then be set using
+    -dsco and -lco just as when using `ogr2ogr`.
```
rcoup, line 148: Should add a comment about setting GDAL configuration options via the environment.

rcoup (commented):
We're likely going to want to set some format-specific defaults for config options eventually. Though I guess we can punt that by saying "we follow GDALs defaults, here's how to change them"

It'd be handy to set `OGR_CURRENT_DATE` (GPKG), `DBF_DATE_LAST_UPDATE` (SHP), to commit time.

And we should add whatever metadata/descriptions/titles/etc we have avaialble.

How are field-names laundered (eg: >8 chars for shapefiles)? At a minimum it'd be nice to show it's _happening_ (eg: `'MyReallyLongField' has been truncated to 'MYREALLYLO'`) — I'm not sure whether ogr2ogr does that or can expose it.

PR #990, opened 2024-05-08
Address some `kart export` review comments - part 1

olsen232:
## Description

Address some review comments from previous PR

## Related links:

Previous PR
https://github.com/koordinates/kart/pull/988

## Checklist:

### DIFF
--- a/kart/tabular/export.py
+++ b/kart/tabular/export.py
@@ -11,6 +11,10 @@ from kart.cli_util import (
 from kart.completion_shared import repo_path_completer
 from kart.crs_util import make_crs
 from kart.exceptions import CrsError
+from kart.tabular.ogr_adapter import (
+    kart_schema_col_to_ogr_field_definition,
+    KART_GEOM_TYPE_TO_OGR_GEOM_TYPE,
+)
 
 
 def is_writable_vector_format(driver):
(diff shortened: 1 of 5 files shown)


kart/tabular/export.py
```diff
@@ -145,9 +158,13 @@ def table_export(
 
     Uses GDAL's OGR drivers to do so - consult https://gdal.org/drivers/vector/index.html
     to find the options specific to a particular driver, which can then be set using
-    -dsco and -lco just as when using `ogr2ogr`.
+    -dsco and -lco just as when using `ogr2ogr`. If these are not set, GDAL's defaults
+    are used, which can also be found in the documentation.
+
+    GDAL's general configuration options can be controlled by setting environment variables -
```
olsen232, line 164: > Should add a comment about setting GDAL configuration options via the environment.

Done.

kart/tabular/export.py
```diff
@@ -208,7 +229,7 @@ def table_export(
         if col.data_type == "geometry":
             continue  # Handled separately
         regular_keys.append(col.name)
-        out_layer.CreateField(kart_col_to_ogr_field(col))
+        out_layer.CreateField(kart_schema_col_to_ogr_field_definition(col))
```
olsen232, line 232: > [Regarding OGR_DATATYPE_MAP] Don't we have this defined elsewhere for imports?

Yes, sort of.
In practice the forward dict and the reverse dict and/or the forward function and reverse function always end up being slightly different - types which exist in one format but not in the other cannot be perfectly roundtripped and so the mapping is not bijective.

However, I think there is value in having the forward and reverse mapping in one file right next to each other. That file is now `ogr_adapter.py` and I added some basic roundtrip tests in `ogr_adapter_test.py`

tests/test_export.py
```diff
@@ -57,3 +61,155 @@ def test_export_datasets(archive, layer, geometry, data_archive, cli_runner):
                 continue
 
             assert field_name[0:10] in output
+
+
+def test_dataset_creation_options(data_archive, cli_runner):
```
olsen232, line 66: > Should add some more test cases:
using dataset creation options
using layer creation options
...

That's all the suggested tests I've got to so far - there are more suggested tests in the comments on the last PR, and I'll get to them in the next PR.
But, I also added tests for the --primary-key-as-field and --primary-key-as-fid options, which I hope helps document them.

olsen232 (commented):
The comments in this PR that appear to be authored by me are actually copied from #988

rcoup (approved):
Nice 👍

PR #991, opened 2024-05-10
Address some `kart export` review comments - part 2

olsen232:
Mostly:
- CRS transforms
- Making sure the user can't overwrite layers without trying to
- Making sure multiple layer datasets / directories work where expected
- Various tests

## Related links:

https://github.com/koordinates/kart/pull/988 contains comments recommending changes along these lines.

## Checklist:

### DIFF
--- a/kart/tabular/export.py
+++ b/kart/tabular/export.py
@@ -318,37 +318,6 @@ def _should_export_geometry(driver, layer_creation_options):
     return True
 
 
-SINGLE_LAYER_FILE_DRIVERS = set(
-    [
-        'CSV',
-        'DGN',
-        'DXF',
-        'ESRI Shapefile',
-        'FlatGeobuf',
-        'GML',
-        'GPSBabel',
-        'GPX',
-        'GeoJSON',
-        'GeoJSONSeq',
-        'Geoconcept',
-        'JML',
-        'JSONFG',
-        'KML',
-        'MBTiles',
-        'MVT',
-        'MapInfo File',
-        'ODS',
-        'OGR_GMT',
-        'PGDUMP',
-        'PMTiles',
-        'S57',
-        'VDV',
-        'WAsP',
-        'XLSX',
-    ]
-)
-
-
 def open_dataset_for_export(
     driver,
     destination,
@@ -363,7 +332,7 @@ def open_dataset_for_export(
     If overwrite is False, makes sure that the given layer doesn't already exist at the target dataset.
     If overwrite is True and the output is a file, removes that file since not all drivers support OVERWRITE=YES.
     """
-    if driver.GetName() in SINGLE_LAYER_FILE_DRIVERS and Path(destination).is_file():
+    if _is_single_layer_file_driver(driver) and Path(destination).is_file():
         if not overwrite:
             raise InvalidOperation(
                 f"Driver {driver.GetName()} cannot add a new layer to an existing file, and file {destination} already exists.\n"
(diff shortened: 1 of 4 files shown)


kart/tabular/export.py
```diff
@@ -182,71 +203,88 @@ def table_export(
...
-        for i, key in enumerate(regular_keys):
-            out_feature.SetField(i, feature[key])
-        out_layer.CreateFeature(out_feature)
-
-    out_ds = None
+        ogr_field_defns[col.name] = kart_schema_col_to_ogr_field_definition(col)
+
+    # Finally, open the dataset for export.
+    try:
+        out_ds = open_dataset_for_export(
```
olsen232, line 246: > I'm unclear what this would do if the output already exists? I guess to allow multiple datasets to end up in the same FGDB/GPKG/etc then we should generally allow this.
For some formats the output is a directory (shapefiles); for others it's a file (GPKG), which complicates things a bit.
We don't want the working copy overwritten, that would be bad.

Logic around overwriting / updating datasets is now much improved = much more cautious.
See tests that make sure you can't overwrite an existing layer without trying to, and that you can add a layer to existing datasets where sensible (at least for the drivers tested. Drivers vary somewhat but hopefully we're testing a big enough range that the logic holds up okay).

As yet, no logic added to check if user is exporting to their working copy.
This is generally not what the user should be doing...
... but at least we prevent them overwriting the contents of their WC, unless they force it.
(Relatively tricky to detect in general if the destination is the working copy - would have to have special logic to check if two postgres:// strings resolve to the same place, for instance.)

tests/test_export.py
```diff
@@ -63,6 +64,38 @@ def test_export_datasets(archive, layer, geometry, data_archive, cli_runner):
             assert field_name[0:10] in output
 
 
+@pytest.mark.parametrize("crs", ["EPSG:4326", "EPSG:27200"])
+def test_export_with_transformed_crs(crs, data_archive, cli_runner):
```
olsen232, line 68: Here's the test checklist and my feeling about how covered it is:
- using dataset creation options ✅︎#990
- using layer creation options ✅︎#990
- gdal config options (via env) ✅#991
- existing directories, existing files, etc ✅︎#991
- maybe a set of common formats? (SHP, FGDB, GeoJSON, CSV) 🤔somewhat done #991
(tests now do at least one export to all of these formats, but they don't dig into the result.)
- error handling / error messages 🤔somewhat done #991
(the more we study individual drivers, the more we can detect their error conditions and raise more helpful errors. But so far error handling is very generic.)
- truncated/fixed/clashing field names (SHP in particular) ❌not done

olsen232 (commented):
Comments that appear to have been authored by me are actually from:
https://github.com/koordinates/kart/pull/988

craigds (approved):

kart/tabular/export.py
```diff
@@ -257,3 +295,134 @@ def _validate_primary_key_as_fid(primary_key_as_fid, schema):
...
+    if driver.GetName() == "CSV":
+        # The CSV driver doesn't export geometry unless this flag is supplied -
+        # so it crashes if we try and create a geometry column when this flag is not supplied.
+        return any(
+            lco.upper().startswith("GEOMETRY=") for lco in layer_creation_options
+        )
+    return True
+
+
+SINGLE_LAYER_FILE_DRIVERS = set(
```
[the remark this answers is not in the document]
olsen232: Tested the drivers myself to find out which ones could write two layers to the same file and then read them again. But yes, your way is better. Done.

PR #993, opened 2024-05-16
git: upgrade to a v2.45.1 base

rcoup:
## Description

Incorporate a bunch of bug & security fixes.

## Related links:

Sourced from https://github.com/koordinates/git/releases/tag/kart-v0.15.2 which builds on the git `v2.45.1` tag.
The associated Windows mingit build is https://github.com/koordinates/git/releases/tag/kart-v0.15.2-windows-mingit-build

## Checklist:

- [x] Have you reviewed your own change?
- [x] Have you included test(s)?
- [ ] Have you updated the [changelog](https://github.com/koordinates/kart/blob/master/CHANGELOG.md)?

PR #995, opened 2024-06-04
git 2.45.1 and vcpkg 2024.04.26

rcoup:
## Description

Update Git to 2.45.1 (#993) and vcpkg to 2024.04.26.

Spatialite caused a segfault updating to the latest on macOS+Linux, so I've left it at the previous release. Nothing obvious in the changelog post-5.1.0, but I haven't bisected or tried trunk from upstream.

A couple of minor tweaks for build/test robustness.

## Related links:

Builds on & replaces #993 

## Checklist:


.github/workflows/build.yml
```diff
@@ -12,12 +12,37 @@ env:
...
+          role-to-assume: 'arn:aws:iam::276514628126:role/ci-kart'
+          aws-region: us-east-1
+          mask-aws-account-id: 'false'
+          role-session-name: kart-ci-gha
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+        with:
+          registry-type: public
+          mask-password: 'false'
```
craigds, line 33: this doesn't seem necessary / a good idea? even if we're not using debug-level logging atm (are we?)

craigds (changes requested):
otherwise LGTM

.github/workflows/build.yml
rcoup: It's how they document it: https://github.com/aws-actions/amazon-ecr-login?tab=readme-ov-file#run-an-image-as-a-service

craigds (approved):

PR #1011, opened 2024-11-07
`import`: Numeric table selection choices

### LINKED ISSUE #881: enumerate the layers when using kart import
kart import, with no layer specified, gives a list of layers to import with a prompt to type the layer name that you would like to import. It would be useful if the list were enumerated  and then the number could be used to indicate which layer is to be imported. This would be easier than typing the layer name (which may be long or possibly contain characters such as '🐘' that are not trivial to enter with a keyboard).

craigds:
## Description

When importing from a database containing multiple tables, this gives you the option of choosing a table to import numerically, so that you don't have to type out the name of the table.

This is backwards compatible; table names are still accepted

⚠️ I made this change using AI (while trying out http://codebuff.com/ ). (But I have reviewed it myself and it seems correct)

## Related links:

Fixes #881

## Checklist:

### DIFF
--- a/kart/tabular/import_source.py
+++ b/kart/tabular/import_source.py
@@ -224,6 +224,10 @@ class TableImportSource:
         # Subclasses should override this if a more useful aggregate description can be generated.
         return "\n".join(s.import_source_desc() for s in import_sources)
 
+    def print_table_list(self, *, do_json=False):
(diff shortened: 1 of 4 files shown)


craigds:
Forgot to pull latest master so I didn't get the macOS 13 change

olsen232 (approved):

PR #1014, opened 2024-11-07
Set GDAL_AUTOLOAD_PLUGINS=OFF

olsen232:
## Description

This fixed an issue where unwanted plugins were being loaded during Kart commands run from within QGIS.
However, it breaks the example Kart usage at `contrib/gdal/in_process/`
Either we remove this example, or we find another fix.

## Checklist:


rcoup:
> Either we remove this example

Yeah, I think this is fine.

From a quick look at the GDAL source there [doesn't seem to be](https://github.com/OSGeo/gdal/blob/181b6b99912783e1e5635d9a6d71629eda443b49/gcore/gdaldrivermanager.cpp#L855-L858) a "no autoload but load this specific path" API.

PR #1025, opened 2024-11-29
msgspec - faster tabular dataset access

### LINKED ISSUE #1018: Speed up kart diff

craigds:
## Description

* Fixes #1018 
* supersedes #1019
* relies on #1027

On a large repo this change causes a 75% increase in throughput (in addition to previous gains from #1019) when generating a large JSONL diff via:
```
time kart diff '[EMPTY]...HEAD' -o json-lines | pv > /dev/null
```

* orjson, msgpack: 16.7 MiB/s
* orjson, msgspec: 28.3 MiB/s
* msgspec only (for both msgpack+json): 29.4 MiB/s <-- **this PR**

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,13 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 
 ## Unreleased
 
+- Much faster access to tabular/vector datasets (about 75% more features processed per second) by switching to [msgspec](https://jcristharif.com/msgspec/) - [#1025](https://github.com/koordinates/kart/pull/1025)
+- diff: Faster JSON-Lines output (also using msgspec)
 - Linux builds now require glibc 2.28+ [#1027](https://github.com/koordinates/kart/pull/1027) - This means minimum distro versions are:
     - Debian 10+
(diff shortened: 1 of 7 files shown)


olsen232 (approved):

craigds:
build fails:

* linux: probably need to upgrade from manylinux2014 to manylinux_2_28 – it's about time anyway
* macOS: ?

PR #1027, opened 2024-12-11
CI upgrades

craigds:
### Upgrade to manylinux 2.28
manylinux2014 is ancient enough that a lot of github actions are starting to not run on it any more; the amount of effort required to deal with this image seems to be rapidly escalating.

This PR upgrades to manylinux_2_28 images.

I think this means Kart is now expected to run on OSes with glibc 2.28 or newer:

>    Debian 10+
>    Ubuntu 18.10+
>    Fedora 29+
>    CentOS/RHEL 8+

In other words Kart on Linux now requires an OS newer than 5-6 years old.

### 

## Related links:

https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux_2_28-almalinux-8-based

## Checklist:


vcpkg-vendor/fix_vendor_libs.py
```diff
@@ -152,6 +152,7 @@
         "/usr/lib/libSystem.B.dylib",
         "/usr/lib/libc++.1.dylib",
         "/usr/lib/libcharset.1.dylib",
+        "/usr/lib/libffi.dylib",
```
rcoup, line 155: we _build_ libffi via vcpkg, we definitely shouldn't be using the system libffi

https://github.com/koordinates/kart/blob/ea74d9ea0cf86cce3e2adf75ee298e157fcf57af/vcpkg-vendor/vcpkg.json#L73

rcoup (changes requested):
LGTM except the libffi thing.

olsen232 (approved):

rcoup:
(I presume `/usr/local/` here was meant to be `/usr/lib/`?)

> Allow deps from /usr/local/ to be found in vendor
>
> Previous logic was that deps with prefix /usr/local/ are "system" deps and either whitelisted or disallowed, but are not searched for in vcpkg-vendor.
> Removed this concept - now if any dep is not whitelisted as a system dep, we search for it in vcpkg-vendor, regardless of its prefix.

Doesn't this mean we'll unilaterally update a library to depend on the vcpkg version instead of the system one?

eg:
1. `/usr/lib/libfoo` is a random system library
2. we build `libfoo` via vcpkg
3. we build a python wheel which is meant to link to our `libfoo`. But it doesn't find it, and links to the system one instead.
4. we find the issue during fix-vendor-libs: previously we would error, but now we just update the library dependencies to point to the vcpkg one?

I mean, in practise it's probably fine? Since ABIs are versioned (`libfoo.99.dylib`), and the vcpkg one will be newer than the system one in basically all cases...

PR #1032, opened 2025-01-08
diff: Faster startup & much-reduced memory consumption

craigds:
## Description

Lazily populate diff items, when `--no-sort-keys` is used (added in #1031; this makes it much more effective)

This mostly eliminates diff bootstrapping time (first feature appears almost immediately) and also avoids linear memory consumption during large diffs

Output seems to be ~30% faster overall, probably mostly due to reduced allocation and garbage collection.
On a 2.3GB JSONL diff locally I see memory consumption drop from ~8GB to ~600MB.

## Checklist:


rcoup:
Can you add some more details on **_why_** it saves so much time/memory? Do the diff items get populated _eventually_ during the process?

kart/diff_structs.py
```diff
@@ -430,6 +433,10 @@ def __json__(self):
         return {k: v for k, v in self.items()}
 
 
+class InvalidatedDeltaDiff(Exception):
+    pass
+
+
 class DeltaDiff(Diff):
     """
```
olsen232, line 441: You've made some big changes to DeltaDiff - add something to the docstring

Need to briefly document what items, iter_items, data, _lazy_initial_contents, and _evaluate_lazy_initial_contents is / how they fit together.

kart/diff_structs.py
```diff
@@ -439,16 +446,83 @@ class DeltaDiff(Diff):
...
     def __setitem__(self, key, delta):
         if key != delta.key:
             raise ValueError("Delta must be added at the appropriate key")
         super().__setitem__(key, delta)
 
+    def _evaluate_lazy_initial_contents(self):
+        if self._lazy_initial_contents is None:
+            return
+        for k, v in self._lazy_initial_contents:
+            if k not in self:
```
olsen232, line 475: Can it happen that k is already in self? Or, can you explain this line?

olsen232 (approved):
This worked out a smaller change than I feared

craigds:
> Can you add some more details on why it saves so much time/memory?

[RichTableDataset.diff](https://github.com/koordinates/kart/blob/e1110622c7d6b1eb1409e625a50168dfe940c0fb/kart/tabular/rich_table_dataset.py/#L137-L139) calls `self.diff_feature()`. That returns a generator that yields `kart.diff_structs.Delta` objects - one for each `pygit2.Delta` we've fetched from the repo.

However, `RichTableDataset.diff` then wraps that generator in a `DeltaDiff` object (which is essentially a dict), so the generator is fully consumed before continuing. This means for large repos we instantiate millions of Delta objects and use a lot of memory before producing any output.

By default, the `kart diff` command then proceeds to sort the diff by PK value. This is a nice-to-have but not really required. Doing this means we cannot _avoid_ consuming the entire generator, hence why this PR is only effective when user passes in the `--no-sort-keys` option. We should probably have a discussion about whether that option could/should be the default.

> Do the diff items get populated eventually during the process?

My initial goal in writing this change was to reduce TTFB for the diff, without being overly concerned about overall performance. i.e. I would have been happy if the TTFB was quicker but the overall speed remained the same. So my initial implementation attempted to store the Delta values in the DeltaDiff as they were consumed, so that then the DeltaDiff could be consumed twice.

I found that initial approach actually made things about 20% _slower_ overall, while succeeding in reducing the TTFB. I didn't really look into why it was slower, since I soon realised it would be better to not store the iterated deltas at all after dumping them to the output (they're only consumed once unless something has gone wrong.) I have added a sentinel value to invalidate the DeltaDiff once the generator has been consumed, so we can't accidentally consume it twice.

Making the change to not store the Delta objects after consumption meant that the memory usage dropped substantially, and as a byproduct of this, the whole process because a lot faster. I think this makes sense since we previously saw flame graphs showing up to 40% of runtime caused by garbage collection.

kart/diff_structs.py
craigds: I was thinking about diff concatenation - deltas that are explicitly added to this diff after the fact should probably override the lazy deltas. But in fact that's not correct because those deltas may in fact need to be _combined_ with the lazy deltas.
It looks like this isn't actually required, since diff concatenation calls `.keys()` anyway and thus consumes the generator fully. Which will work correctly.

kart/diff_structs.py
```diff
@@ -430,25 +433,117 @@ def __json__(self):
...
+    It is possible to pass in an iterator of Deltas (e.g. a generator) to the DeltaDiff constructor,
+    in which case the Deltas are NOT immediately stored in the DeltaDiff.
+    This is useful because there may be a lot of Deltas, and we don't want to store them in memory.
+
+    The only correct way to consume a DeltaDiff populated by a generator is to call `iter_items()`,
+    which will consume the iterator as it yields Deltas.
+    Calling that method will invalidate the DeltaDiff, so it cannot be used again (doing so will throw an exception)
+
+    Calling any other dict-like method (keys(), items(), len() etc) will consume the iterator and store its
+    contents in memory, which may be expensive.
```
rcoup, line 455: Do we _need_ this as a "dict-like" object considering all the `self._evaluate_lazy_initial_contents()` calls in the methods you're adding?

Or would it be better to treat it a bit like Django's QuerySet/ValuesQuerySet or QueryDict/MutableQueryDict classes where you explicitly get a related-but-different class for the occasions when you need to do special things, and only that class implements the full interface?

craigds: I had the same thought, but I didn't see a way to change it without being quite invasive. The problem is that sometimes we concatenate these diffs together, usually when you do a diff between a commit and a working copy. In that case we get a commit<>commit diff, and then we get a WC diff and we call `Diff.concatenate(diff1, diff2)` to merge them together.

In that situation we need to be able to access the diffs in a dicty way so we can merge the contents. This can't be done without consuming the generator function into memory.

So we could have a completely separate set of diff structs for commit<>WC diffs, but it would probably cause a lot of headaches IMO. vs this relatively isolated hack

rcoup: I think the reasoning/approach is good, I'm more worried about a footgun for accidentally starting to always consume the iterator when we make a seemingly innocuous change in the future.

> So we could have a completely separate set of diff structs for commit<>WC diffs

I didn't mean a completely separate set of structs, more like the existing lazy class having with a `.consume()`/`.resolve()`/etc method which consumes the iterator and returns/recasts to some sort of ResolvedDataDiff subclass, which has the `__len__`/`__getitem__`/`keys`/`values`/etc methods defined on it.

craigds: See #1034 for a partial implementation of this idea

PR #1034, opened 2025-01-09
LazyDeltaDiff

craigds:
## Description

Add a LazyDeltaDiff class

This is a refactoring of #1032; the aim is to make the lazy-ness of the
DeltaDiff class more explicit.

A LazyDeltaDiff doesn't behave like a dict, unlike other diff structs.
It must be consumed either via the `iter_items()` method or by calling
`resolve()` to convert it to a `DeltaDiff` first.

## Related links:

#1032 
## Checklist:

### DIFF
--- a/kart/diff_structs.py
+++ b/kart/diff_structs.py
@@ -451,65 +451,6 @@ class Diff(RichDict):
         return self
 
 
-class LazyDeltaDiff:
-    """
-    A LazyDeltaDiff is like a DeltaDiff containing an iterator of Deltas, which is lazily evaluated.
-    This is useful because there may be a lot of Deltas, and we don't want to store them in memory.
-
-    The only correct way to consume a LazyDeltaDiff populated by a generator is to call `items()`,
-    which will consume the iterator as it yields Deltas.
-    Calling that method will invalidate the LazyDeltaDiff, so it cannot be used again (doing so will throw an exception)
-
-    To consume the iterator into memory and turn the LazyDeltaDiff into a DeltaDiff, call `resolve()`
-    """
-
-    _wrapped_iter: Iterator[Delta]
-
-    def __init__(self, initial_contents: Iterable[Delta] = ()):
-        wrapped_iter = iter(initial_contents)
-        try:
-            first_item = next(wrapped_iter)
-        except StopIteration:
-            self._wrapped_iter = iter(())
-            self._bool = False
-        else:
-            self._wrapped_iter = chain((first_item,), wrapped_iter)
-            self._bool = True
-        self._consumed = False
-
-    def __bool__(self):
-        return self._bool
-
-    def __add__(self, other):
-        resolved = self.resolve()
-        resolved += other
-        return resolved
-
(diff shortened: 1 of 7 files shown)


rcoup:
From #1035, should address here.

> I also noticed via pytest that `DeltaDiff.__repr__()` calls `.__str__()` (c/- RichDict) which then calls `.keys()` and consumes the lazy iterator too, we should fix that for lazy diffs.

craigds:
> I also noticed via pytest that DeltaDiff.__repr__() calls .__str__() (c/- RichDict) which then calls .keys() and consumes the lazy iterator too, we should fix that for lazy diffs.

I think this PR solves that by default (I haven't included any `__repr__` on LazyDeltaDiff)

kart/diff_structs.py
```diff
@@ -432,117 +438,96 @@ def type_counts(self):
...
-        return result
+        return self._bool
 
-    def __len__(self):
-        self._evaluate_lazy_initial_contents()
-        return super().__len__()
+    def __add__(self, other):
+        resolved = self.resolve()
+        resolved += other
+        return resolved
```
rcoup, line 486: Do we need to consume it here?

```suggestion
    def __add__(self, other):
      self._check_not_consumed()
      self._consumed = True
      return self.__class__(itertools.chain(self._wrapped_iter, other.items())

    def __iadd__(self, other):
      self._check_not_consumed()
      self._wrapped_iter = itertools.chain(self._wrapped_iter, other.items())
```

Maybe I'm missing what it's used for?

kart/diff_structs.py
```diff
@@ -432,117 +438,96 @@ def type_counts(self):
...
+        return DeltaDiff(self._wrapped_iter)
 
-    def values(self):
-        """
-        Overrides the dict.values() method to ensure we consume any lazy initial contents first
-        """
-        self._evaluate_lazy_initial_contents()
-        return super().values()
+
+class DeltaDiff(Diff):
```
rcoup, line 513: Move this above LazyDeltaDiff?

kart/diff_structs.py
```diff
@@ -432,117 +438,96 @@ def type_counts(self):
...
+        resolved = self.resolve()
+        resolved += other
+        return resolved
 
-    def items(self):
-        self._evaluate_lazy_initial_contents()
-        return super().items()
+    def _check_not_consumed(self):
+        if self._consumed:
+            raise ValueError("LazyDeltaDiff has already been consumed")
```
rcoup, line 490: Match the error if you muck with dicts while iterating?

```suggestion
            raise RuntimeError("LazyDeltaDiff has already been consumed")
```

rcoup (approved):
Thanks! LGTM, couple of small comments 👍

kart/diff_structs.py
craigds: The deltas are _merged_ when you add two diffs. i.e. if you add the `a..b` and `b..c` diffs, the result is `a..c`. This isn't the same as just chaining the list of deltas together.

PR #1036, opened 2025-01-16
mssql: Support custom CRSs by maintaining OGR convention tables

olsen232:
## Description

For custom CRSs in mssql, follow the OGR convention of storing them in a table called `spatial_ref_sys` (which we create if necessary) using the same table definition that OGR uses.
Also stores info about the columns which reference these CRSs in `geometry_columns`.

Reads CRSs from the same tables when diffing, committing and importing from mssql server.

## Related links:

OGR code for creating these tables:
https://github.com/OSGeo/gdal/blob/196f8550298a9e05b4c9739a6022f81349473d13/ogr/ogrsf_frmts/mssqlspatial/ogrmssqlspatialdatasource.cpp#L1349-L1352

## Checklist:

### DIFF
--- a/kart/sqlalchemy/adapter/sqlserver.py
+++ b/kart/sqlalchemy/adapter/sqlserver.py
@@ -233,25 +233,34 @@ class KartAdapter_SqlServer(BaseKartAdapter, Db_SqlServer):
             )
         )
 
-        srs_table_joins = {
-            "MSSRS.*": (
-                "sys.spatial_reference_systems MSSRS",
-                "MSSRS.spatial_reference_id",
-            )
-        }
+        srs_table_selects = [
+            "MSSRS.authority_name AS ms_auth_name",
+            "MSSRS.authorized_spatial_reference_id AS ms_auth_srid",
+            "MSSRS.well_known_text AS ms_wkt",
+        ]
+        srs_table_joins = [
+            ("sys.spatial_reference_systems MSSRS", "MSSRS.spatial_reference_id")
+        ]
+
         if has_ogr_spatial_ref_sys_identifier:
-            srs_table_joins["OGRSRS.*"] = (
-                f"{ogr_spatial_ref_sys_identifier} OGRSRS",
-                "OGRSRS.srid",
-            )
+            srs_table_selects += [
+                "OGRSRS.auth_name AS ogr_auth_name",
+                "OGRSRS.auth_srid AS ogr_auth_srid",
+                "OGRSRS.srtext AS ogr_wkt",
+            ]
+            srs_table_joins += [
+                (f"{ogr_spatial_ref_sys_identifier} OGRSRS", "OGRSRS.srid")
+            ]
+
+        srs_table_selects = ", ".join(srs_table_selects)
 
         ms_crs_info = list(
             cls._parse_crs_rows(
                 sess.execute(
                     f"""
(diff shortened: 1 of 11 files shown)


kart/sqlalchemy/adapter/sqlserver.py
```diff
@@ -219,38 +219,72 @@ def all_v2_meta_items_including_empty(
...
+            if not crs_row:
+                continue
+            column_name = crs_row["column_name"]
+            column_srid = crs_row["column_srid"]
+            if not column_srid:
+                continue
+
+            ms_cols = ["authority_name", "authorized_spatial_reference_id", "well_known_text"]
+            ogr_cols = ["auth_name", "auth_srid", "srtext"]
+            for cols in (ms_cols, ogr_cols):
```
craigds, line 277: why not select particular columns rather than `*` and use aliases to normalise the schema of the returned rows? Then this gets less crazy

tests/test_working_copy_sqlserver.py
```diff
@@ -585,52 +593,137 @@ def test_checkout_custom_crs(
                 '-     UNIT["degree", 0.0174532925199433,',
                 '-         AUTHORITY["EPSG", "9122"]],',
                 '-     AUTHORITY["EPSG", "4326"]]',
-                "- ",
-                "--- nz_pa_points_topo_150k:meta:schema.json",
-                "+++ nz_pa_points_topo_150k:meta:schema.json",
-                "  [",
-                "    {",
+                '- ',
```
craigds, line 596: pre-commit hook not running? Ruff should be converting these to double-quotes

`pre-commit install`

craigds (approved):

tests/test_working_copy_sqlserver.py
olsen232: Re-committed everything from my mac where this is working, haven't got around to fixing it on my VM

kart/sqlalchemy/adapter/sqlserver.py
```diff
@@ -219,38 +219,72 @@ def all_v2_meta_items_including_empty(
...
+
+        ms_crs_info = list(cls._parse_crs_rows(
             sess.execute(
                 f"""
-                SELECT TOP 1 :column_name AS column_name, {cls.quote(g)}.STSrid AS srid, SRS.*
+                SELECT TOP 1 :column_name AS column_name, {cls.quote(g)}.STSrid AS column_srid, {', '.join(srs_table_joins.keys())}
                 FROM {table_identifier}
-                LEFT OUTER JOIN sys.spatial_reference_systems SRS
-                ON SRS.spatial_reference_id = {cls.quote(g)}.STSrid
+                {chr(10).join(f"LEFT OUTER JOIN {table} ON {column} = {cls.quote(g)}.STSrid" for table, column in srs_table_joins.values())}
```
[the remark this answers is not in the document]
olsen232: Not done - wouldn't work as a generator expression, which I mostly like how it is.
Switched to " ".join instead - the newlines in this SQL are mostly only helpful in making the source code more readable, so there's no point in having ugly looking ones that are added programmatically - the db engine doesn't care.

kart/sqlalchemy/adapter/sqlserver.py
olsen232: Done.

PR #1038, opened 2025-02-04
Faster diff (for some diffs) by skipping pygit2

craigds:
## Description
Background:
===========

pygit2 Tree.diff_to_tree is very slow for some large trees. This seems particularly bad on a repo containing a large table dataset (many rows, but few columns / small objects)

e.g. with this 43M-feature repo:

```
$ git ls-tree -r c62a4290d7d1d4039ff06afcbe364bac6abf0e69 | wc -l
43482704
```

A small diff (a couple thousand features changed) takes 25s for `kart diff` to bootstrap. cProfile shows the time all spent in:
```
25.534 {method 'diff_to_tree' of '_pygit2.Tree' objects}
```

I was unable to find anything obvious in pygit2 or libgit2 source code pertaining to this. e.g. the flags we're passing appear to be the right ones.

Since `git diff-tree` can produce this diff within ~0.1s I just switched to using that as a subprocess. The results are identical but much faster (at least in the above case)

Performance of larger diffs / smaller repos
===========================================

The difference is less stark with other repos.

If I switch to a full diff (`[EMPTY]...HEAD`) I find no noticeable difference in performance between the pygit2 and git implementations.

If I switch to a repo with less features but larger objects, I find a much smaller difference than the above.

So to reiterate the win appears to be largest when dealing with repos with:

* large number of features
* each feature is small
* final diff covers a small fraction of the repo (e.g. 1% or less)

e.g. on [NZ Property
Titles](https://data.linz.govt.nz/layer/50804-nz-property-titles/repository/) I found a ~25% improvement when asking for a 14477-feature diff (0.6% of total features)

before:
```
kart diff 'HEAD^^^...HEAD' -o json-lines --no-sort-keys --output
/dev/null
1.89s user 0.21s system 99% cpu 2.111 total
```

After:
```
kart diff 'HEAD^^^...HEAD' -o json-lines --no-sort-keys --output
/dev/null
1.38s user 0.13s system 98% cpu 1.525 total
```
## Checklist:

### DIFF
--- a/kart/base_dataset.py
+++ b/kart/base_dataset.py
@@ -2,6 +2,7 @@ import functools
 import logging
 import sys
 from collections.abc import Iterable
+from itertools import chain
 from typing import Callable, Optional, Protocol, TYPE_CHECKING, ClassVar, Iterator
 
 import click
@@ -15,7 +16,7 @@ from kart.exceptions import InvalidOperation, UNSUPPORTED_VERSION, PATCH_DOES_NO
 from kart.key_filters import DatasetKeyFilter, MetaKeyFilter, UserStringKeyFilter
 from kart.meta_items import MetaItemFileType, MetaItemVisibility
 from kart.raw_diff_delta import RawDiffDelta
-from kart.utils import chunk
+from kart.utils import chunk, iter_records_from_file
 
 
 class BaseDatasetMetaClass(type):
@@ -205,8 +206,7 @@ class BaseDataset(metaclass=BaseDatasetMetaClass):
         if reverse:
             self_subtree, other_subtree = other_subtree, self_subtree
 
-        # TODO: iterate over 0-terminated 'lines' of output rather than using check_output. Efficiently. How?
-        output = subprocess_util.check_output(
+        p = subprocess_util.Popen(
             [
                 "git",
                 "diff-tree",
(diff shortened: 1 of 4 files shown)


kart/base_dataset.py
```diff
@@ -189,12 +191,55 @@ def diff_to_working_copy(
...
+            ]
+        )
+        if not output:
+            # empty diff
+            return
+
+        pieces = output.split(b"\0")
+        # diff-tree adds an extra \0 on the end of the output; trim it off
+        assert len(pieces) % 2 == 1
+        pieces.pop()
```
olsen232, line 227: assert that pieces.pop() == b'' or whatever you expect it to be

kart/base_dataset.py
```diff
@@ -189,12 +191,55 @@ def diff_to_working_copy(
...
+        reverse: bool = False,
+        path_transform: Callable[[str], str] | None = None,
+    ) -> Iterator[RawDiffDelta]:
+        self_subtree = self.get_subtree(subtree_name)
+        other_subtree = other.get_subtree(subtree_name) if other else self._empty_tree
+
+        if reverse:
+            self_subtree, other_subtree = other_subtree, self_subtree
+
+        # TODO: iterate over 0-terminated 'lines' of output rather than using check_output. Efficiently. How?
```
olsen232, line 208: Found this PEP from 2005 that is.... languishing...
It does have a little snippet inside it that shows how you can do this kind of thing though:
https://bugs.python.org/issue1152248

olsen232 (approved):

PR #1061, opened 2025-08-25
Override crs in imported tiles

### LINKED ISSUE #1060: `--override-crs` should rewrite the CRS in imported LAZ files
In #1057 I added the `--override-crs` flag to pointcloud imports. It avoids errors when tiles have different CRSes, and sets the CRS of the dataset in the `meta/` item.

It doesn't actually change the CRS in the imported LAZ files though. It needs to, otherwise those LAZ files are not going to be useable with most software (or will require CRS override at usage time)

We should:

* use the reader's `override_srs` option in the existing PDAL conversion pipelines: https://pdal.io/en/stable/stages/readers.las.html#readers-las
* force a PDAL conversion step even if not otherwise required, when this option is used.

craigds:
## Description

When --override-crs is given for raster or pointcloud import, the tiles need to be rewritten (using PDAL or GDAL) to override the CRS (as well as setting the CRS in the dataset meta). This change does that.

## Related links:

Fixes #1060

## Checklist:


kart/point_cloud/import_.py
```diff
@@ -250,12 +247,30 @@ def get_actual_merged_metadata(self, all_metadata):
...
+                )
+        elif self.convert_to_cloud_optimized and not is_copc(tile_source.metadata):
+            # Convert to COPC without CRS override
             return convert_tile_to_copc
         return None
 
     def existing_tile_matches_source(self, source_oid, existing_summary):
         """Check if the existing tile can be reused instead of reimporting."""
+        # If override_crs is specified, we need to rewrite the CRS in the files,
+        # so we can't reuse existing tiles
```
craigds, line 270: there's actually at least a minor issue here - if you import the same files multiple times, each time with the same `--override-crs`, this will import all tiles again each time, which is unnecessary.

I'm wondering if we should be checking the CRS of the previous tile itself here (or we could check the CRS of the previous dataset and assume it matches the previous tile, which is *much* cheaper esp if the tiles are on LFS). If the previous dataset/tile matches the override, we can possibly return True here (if the sourceOID matches)

craigds: fixed this in the last commit.

PR #1064, opened 2025-10-07
Some build fixes

craigds:
## Description

A collection of build fixes to improve the state of CI - see individual commits.

This drops MacOS (arm64) temporarily since it never finds a runner, gets linux (arm64 & dev) builds green, and makes MacOS (amd64) get a lot further than it did (and makes it much faster).

More fixes in a later PR to get things even healthier.

## Checklist:

### DIFF
--- a/kart/tabular/export.py
+++ b/kart/tabular/export.py
@@ -361,9 +361,15 @@ def open_dataset_for_export(
         if "OVERWRITE=YES" in layer_creation_options:
             layer_creation_options.remove("OVERWRITE=YES")
 
-    result = driver.Open(destination, update=True)
-    if result is not None:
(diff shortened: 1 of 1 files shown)


olsen232 (approved):
We've never done it before but we should consider putting an OVERLAY_README.md in each overlay that lists why we need an overlay for this particular package.

When an overlay changes, like right now, it's impossible for the reviewer to know what's changing because it changed upstream and what changes have been made locally. It would also help us realise if we ever have no or few good reasons to have an overlay at all, which would be nice if we ever got to delete some overlays.

craigds:
yep good point, it's hard to know why we have the ones we have. I'll add something

PR #1073, opened 2025-11-10
Cleanup ci

craigds:
## Checklist:


craigds:
this doesn't actually work because the macos/windows builds happen on the push event and dont have access to the PR information. so they can't tell if the ci-build-all label has been applied. frustrating but okay

PR #1100, opened 2026-04-28
Support some kart functions when running from a git clone

### LINKED ISSUE #1099: Support running some kart commands from git clone
**Is your feature request related to a problem? Please describe.**

I like to store other files with my kart repositories, such as README's and JSON Schemas, it would be great if I could have both a kart checkout and a git checkout of the same repository so I don't need to have two copies of the same git repo.

**Describe the solution you'd like**

I would like this flow to succeed

```
git clone --no-checkout kart@data.koordinates.com:linz/nz-canal-centrelines-topo-150k
cd nz-canal-centrelines-topo-150k
kart export ...
```

**Describe alternatives you've considered**
using git worktrees off the kart repo, kind of work but are limited by which branches are associated 

```
kart clone kart@data.koordinates.com:linz/nz-canal-centrelines-topo-150k
cd nz-canal-centrelines-topo-150k/.kart
git worktree add --no-checkout ../another-branch
``` 

**Additional context**

blacha:
## Description

It can be useful to run some kart commands from a sparse git checkout, eg `kart export` currently this fails with "Error: Current directory is not an existing Kart repository" 

Kart can partially work with these folders, I have added a warning log to show support may not work.

## Checklist:

- [x] Have you reviewed your own change? - Works locally
- [No] Have you updated the [changelog](https://github.com/koordinates/kart/blob/master/CHANGELOG.md)? - I am Wondering if this should be a "undocumented feature"?

Fixes #1099

### DIFF
--- a/kart/repo.py
+++ b/kart/repo.py
@@ -553,6 +553,9 @@ class KartRepo(pygit2.Repository):
 
     def validate_gitdir_name(self):
         if not self.is_bare and self.gitdir_path.stem not in self.KART_DIR_NAMES:
+            if self.gitdir_path.stem == ".git":
+                L.warning("Selected repo does't follow Kart conventions, but proceeding anyway.")
+                return
             raise NotFound(
                 "Selected repo doesn't follow Kart convention of keeping internals in a '.kart' folder. Perhaps a git repo?",
                 exit_code=NO_REPOSITORY,
--- a/kart/repo.py
+++ b/kart/repo.py
@@ -1,6 +1,7 @@
 import contextlib
 import logging
 import re
+import os
 import struct
 import sys
 from enum import Enum
@@ -553,7 +554,7 @@ class KartRepo(pygit2.Repository):
 
     def validate_gitdir_name(self):
         if not self.is_bare and self.gitdir_path.stem not in self.KART_DIR_NAMES:
-            if self.gitdir_path.stem == ".git":
+            if self.gitdir_path.stem == ".git" and os.environ.get("KART_ALLOW_FROM_GIT") == "1":
                 L.warning("Selected repo does't follow Kart conventions, but proceeding anyway.")
                 return
             raise NotFound(
--- a/kart/repo.py
+++ b/kart/repo.py
@@ -559,7 +559,7 @@ class KartRepo(pygit2.Repository):
                 and os.environ.get("KART_ALLOW_FROM_GIT") == "1"
             ):
                 L.warning(
-                    "Selected repo does't follow Kart conventions, but proceeding anyway."
+                    "Selected repo doesn't follow Kart conventions, but proceeding anyway."
                 )
                 return
             raise NotFound(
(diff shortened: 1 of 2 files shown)


craigds (changes requested):
Thanks for submitting this, sorry it took a week to respond.

TBH it's not quite what I had in mind - specific commands like `kart export` can work but I suspect trying to run some commands (eg `import`) will error pretty confusingly.

I also think running kart from a git repo is likely to usually be an accident - a developer with muscle memory perhaps :)

@olsen232 suggested an alternative to allow-listing specific commands:

> I think this would be good if you could put it in the gitconfig or an env var or something. Don't want kart to try and read git repos in general - who knows what it'll do

Which of those approaches would best suit your use-case? Either or both would be good I think.

blacha:
@craigds Most of the commands I tried seemed to work as expected including "import".

"import" does leave the git tree not quite in a consistent state but the commit is there and the checked out files can be accessed via `git` if required

Simple commands all work as expected

```shell
kart data ls
kart fsck
kart show
```

Some actions require a working copy, but creating that works as expected.

```shell
kart diff # fails -- requires working copy

kart create-workingcopy 
kart diff # now works
```

I have also tried a entire commit flow

```shell
kart create-workingcopy
# Edit feature in QGIS
kart commit  -m 'simple edit'
git show # shows the kart commit
```

IMO having a warning saying this may not behave as expected is appropriate, another option could be to enable it via a env flag `KART_ALLOW_FROM_GIT=1`

olsen232:
Hi, sorry for taking a while to get to this. 

I'm more concerned that people who are new to Kart - which is most people, right at the moment - will do something they didn't intend to with Kart*, than I am with fresh possibilities of hybrid git/Kart repositories. I'm happy for there to be a way to make this work but would prefer if it didn't affect the current default behaviour where git repos and Kart repos are not directly interoperable and you get an error if you type git instead of kart or vice versa. So, I think an env flag like KART_ALLOW_FROM_GIT would be a good compromise.

*Possible ways to shoot yourself in the foot by having a repo where you can run both git and Kart include:
- creating a non-bare git workdir where all your your geospatial features are checked out as individual files. These will just be a bunch of binary files that no program in particular can do anything with - much better they just stay hidden in the backend like they normally do in a Kart repo.
- switching branch using git instead of kart will not cause your GPKG (or other database type) working copy to be updated with the contents of that branch, so next time you run Kart it will be confused by the mismatch between the git state and the working copy state.

kart/repo.py
```diff
@@ -553,6 +554,14 @@ def is_kart_branded(self):
 
     def validate_gitdir_name(self):
         if not self.is_bare and self.gitdir_path.stem not in self.KART_DIR_NAMES:
+            if (
+                self.gitdir_path.stem == ".git"
+                and os.environ.get("KART_ALLOW_FROM_GIT") == "1"
+            ):
+                L.warning(
+                    "Selected repo does't follow Kart conventions, but proceeding anyway."
```
olsen232: does't -> doesn't

olsen232 (approved):

kart/repo.py
blacha:
```suggestion
                    "Selected repo doesn't follow Kart conventions, but proceeding anyway."
```

PR #1101, opened 2026-05-06
feat: track attachment files across kart status / restore / checkout / commit

TeoGoddet:
Makes attachment files (plain blobs tracked alongside datasets — `LICENSE.txt`, `README.md`, project config, etc.) first-class in the working-copy workflow. Until now they were invisible to `status`, couldn't be restored or committed, and were left stale when HEAD moved.

Part of #583 (steps 2–6).

### What changes

- `kart status` reports modified/untracked/deleted attachment files. JSON output gains a `files` key under `workingCopy`, only emitted when there are attachment changes (shape unchanged for repos without attachments).
- `kart restore [PATH]` accepts attachment paths alongside dataset filters.
- `kart checkout` / `switch` / `reset` sync attachment files when HEAD moves: extract files in the new tree, delete files that are gone from it.
- `kart clone` and `kart create-workingcopy` extract attachment files to disk. Previously a fresh clone of a repo with a `LICENSE.txt` left the file missing from the workdir.
- `kart commit` picks up attachment changes. WC diffs store blob OIDs rather than raw bytes, so `apply_files_diff` resolves them via pygit2 before writing.

### Notes

- Change detection reuses the existing workdir-index machinery (the same `git diff`/mtime approach used for tile datasets) rather than re-hashing every attachment on each diff. The plumbing is shared between `FileSystemWorkingCopy` (tile datasets) and a lightweight `AttachmentWorkdirIndex` (tabular-only repos) via a common base. When no index exists yet, status/diff fall back to a full scan rather than creating an index as a side effect of a read.
- Tile dataset files (e.g. `elevation/EK.tif`) are excluded from attachment tracking so a raster tile isn't committed as both an LFS tile and a plain blob.

### Not changed

- `kart commit-files` (scripted file commits) is untouched.
- The dataset working-copy pipeline (GeoPackage sync, `reset_to_head`) is unchanged.
- No new CLI flags. Repos without attachment files behave as before.

### Tests

`test_status` / `test_restore` / `test_checkout` cover the status, restore, checkout, delete-on-switch and commit round-trips. `test_workdir_cache` covers the index/cache layer. `test_attachment_edge_cases` covers the `repo[oid]` error path and that `create-workingcopy` puts files on disk.

The QGIS plumbing commit referenced by #1103 is also extracted there for standalone review; review this one first.

### DIFF
--- a/kart/checkout.py
+++ b/kart/checkout.py
@@ -225,6 +225,9 @@ def checkout(
         repo.configure_do_checkout_datasets(non_checkout_spec, False)
 
     TableWorkingCopy.ensure_config_exists(repo)
+    # Capture old_tree before set_head so _remove_deleted_attachment_files can compare
+    # the tree before the switch with the tree after.
+    old_tree = repo.head_tree if not repo.head_is_unborn else None
     repo.set_head(head_ref)
 
     repo_key_filter = (
@@ -238,7 +241,6 @@ def checkout(
         else ()
     )
 
-    old_tree = repo.head_tree if not repo.head_is_unborn else None
     if do_switch_commit or do_switch_spatial_filter or discard_changes:
         # Changing commit, changing spatial filter, or discarding changes mean we need to update every dataset:
         repo.working_copy.reset_to_head(
@@ -538,18 +540,26 @@ def _restore_attachment_files(repo, source_tree, rel_paths):
     Paths that do not exist as files in source_tree are silently skipped, so callers can safely
     pass the same filter list that was used for dataset restore (the reviewer's preferred approach:
     each filter is tried on both sides; the side that matches wins).
+
+    Successfully restored paths are also added to the workdir-index so that future calls to
+    WorkdirDiffCache.dirty_attachment_paths() can detect changes efficiently via git's mtime
+    optimisation rather than hashing all tracked files on every diff.
     """
     workdir = str(repo.workdir_path)
+    restored = []
     for rel_path in rel_paths:
         try:
             subprocess.check_call(
                 ["git", "-C", workdir, "checkout", source_tree.id.hex, "--", rel_path],
                 stderr=subprocess.DEVNULL,
             )
+            restored.append(rel_path)
         except subprocess.CalledProcessError:
             # Path doesn't exist as a file in source_tree (e.g. it's a dataset name or
             # feature key) — silently skip so the caller can pass unfiltered user args.
             pass
+    if restored:
+        _update_attachment_workdir_index(repo, restored)
 
 
 def _restore_all_attachment_files(repo, source_tree):
@@ -561,6 +571,30 @@ def _restore_all_attachment_files(repo, source_tree):
     if not tracked:
         return
     _restore_attachment_files(repo, source_tree, sorted(tracked))
+
+
+def _update_attachment_workdir_index(repo, rel_paths):
+    """
+    Ensures the file-system working copy exists and records the given attachment paths in its
+    workdir-index, so WorkdirDiffCache can detect subsequent changes via git's mtime
+    optimisation (avoiding a full re-hash of every tracked attachment on every diff).
+
+    Tile-based datasets (point-cloud / raster) already maintain this workdir-index. For
+    tabular-only repos that have attachment files but no tile datasets, the workdir would not
+    otherwise be initialised - so we create it here (index + state DB) and track attachments
+    through the same mechanism rather than a parallel one.
+    """
+    from kart.workdir import FileSystemWorkingCopy, FileSystemWorkingCopyStatus
+
+    wc = FileSystemWorkingCopy(repo)
+    if wc.status() == FileSystemWorkingCopyStatus.UNCREATED:
+        wc.create_and_initialise()
+    # Keep the workdir's tracked tree in sync with HEAD. For tile repos the dataset reset does
+    # this; an attachment-only workdir has no tile reset, so set it here (and on every HEAD
+    # change) to avoid a spurious BAD_WORKING_COPY_STATE from kart status / diff.
+    if not repo.head_is_unborn:
+        wc.update_state_table_tree(repo.head_tree)
+    wc.add_paths_to_index(rel_paths)
 
 
 def _remove_deleted_attachment_files(repo, old_tree, new_tree):
@@ -598,8 +632,6 @@ def _restore_attachments_to_head(repo, old_tree=None):
     _restore_all_attachment_files(repo, new_tree)
     if old_tree is not None:
         _remove_deleted_attachment_files(repo, old_tree, new_tree)
-
-
 
 
 @click.command(cls=KartCommand)
--- a/kart/checkout.py
+++ b/kart/checkout.py
@@ -238,6 +238,7 @@ def checkout(
         else ()
     )
 
+    old_tree = repo.head_tree if not repo.head_is_unborn else None
     if do_switch_commit or do_switch_spatial_filter or discard_changes:
         # Changing commit, changing spatial filter, or discarding changes mean we need to update every dataset:
         repo.working_copy.reset_to_head(
@@ -246,7 +247,7 @@ def checkout(
             non_checkout_datasets=non_checkout_datasets,
         )
         if do_switch_commit or discard_changes:
-            _restore_attachments_to_head(repo)
+            _restore_attachments_to_head(repo, old_tree=old_tree if do_switch_commit else None)
     elif do_switch_checkout_datasets:
         # Not doing any of the above - just need to change those datasets newly added / removed from the non_checkout_list.
         repo.working_copy.reset_to_head(
@@ -448,11 +449,12 @@ def switch(ctx, create, force_create, discard_changes, do_guess, refish):
 
         head_ref = branch.name
 
+    old_tree = repo.head_tree if not repo.head_is_unborn else None
     repo.set_head(head_ref)
 
     if do_switch_commit or discard_changes:
         repo.working_copy.reset_to_head()
-        _restore_attachments_to_head(repo)
+        _restore_attachments_to_head(repo, old_tree=old_tree if do_switch_commit else None)
 
 
 def _find_remote_branch_by_name(repo, name):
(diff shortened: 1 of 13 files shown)


kart/checkout.py
```diff
@@ -500,18 +509,114 @@ def restore(ctx, source, filters):
...
+    # source tree (matching the all-datasets restore above). Otherwise we restore only the explicitly
+    # named files. Untracked files in the working directory are left alone (matching git's `git
+    # restore` semantics).
+    if not filters:
+        _restore_all_attachment_files(repo, source_tree)
+    elif file_filters:
+        _restore_attachment_files(repo, source_tree, file_filters)
+
+
+def _split_restore_filters(repo, source_tree, filters):
```
olsen232: I think this will have an undesirable bug as it is - a dataset could exist on one branch but not on another branch. Say the dataset is called "mydataset" and exists only at tree A. Right now going from tree A to tree B will treat "mydataset" as a dataset filter, and going from tree B to tree A will treat "mydataset" as a file filter.

My preferred solution would be: don't bother to split the filters into dataset filters and file filters. Leave the parsing code as is, and pass the same filters in both places, but be willing to treat them as dataset filters *and* file filters - ie, if the user passes the following filters:
`aaa`, `bbb:1`, `bbb:2`, `README.txt`
Then that matches all of the following:
- all features at a possible dataset called `aaa`
- a possible file called `aaa`
- features `bbb:1` and `bbb:2`
- all features at a possible dataset called `README.txt`
- a possible file called `README.txt`

Just by the way:
The other possible solution is to look in both trees for datasets before splitting into dataset filters and file filters. But, that's a bit more code for - in my opinion - no real gain, since reusing the same filters in both places does what we want anyway, and, also in that case you would need to invent the concept of an empty- filter. Eg if the user provides no filters, that would be converted into code as no dataset and no file filters: (None, None). But if the user provides one file filter eg README.txt, that would have to be converted to two filters, one of which matches nothing eg: (EMPTY, ["README.txt"]). So, I prefer the first solution.

olsen232:
Hi @TeoGoddet 

Thanks a lot for the contribution!

I've just started reviewing this. I'll do my best to help you get it merged - it's a big change so there might be some changes required. Hopefully your other change https://github.com/koordinates/kart-qgis-plugin/pull/133 doesn't depend too closely on the specifics. I'd recommend holding off on building too much more on top of this until we can get this merged.

kart/diff_util.py
```diff
@@ -175,14 +189,18 @@ def get_file_diff(
...
+    directory (target_rs is then assumed to be the HEAD-equivalent tracked by the working
+    directory). Otherwise it is generated between base_rs.tree and target_rs.tree.
     """
+    repo = target_rs.repo
+    old_tree = base_rs.tree
 
-    # We don't yet support attachment diffs in the workdir
-    assert not include_wc_diff
+    if include_wc_diff:
+        return _get_workdir_file_diff(repo, old_tree, repo_key_filter)
```
olsen232: Ok I feel a bit bad about this one, but...

Kart already has code for finding the difference between the current workdir and what's checked in.
It's found in kart/workdir.py. It's used for tile-based datasets. Oddly, even though Kart doesn't support attachments, or maybe just barely supports attachments, it has for a long time supported datasets that are made of files: specifically point-cloud datasets made of LAZ files and raster datasets made of TIFF files.

You may have found it and thought it looked needlessly complex or perhaps too dataset specific. There's some truth to that, of course: it was written to support the file-based datasets - it doesn't yet support attachments.

However, it does have certain advantages. If I recall correctly - it reuses git machinery to maintain an index of files. A tile-based dataset could contain gigabytes of tiles. Using the simplest algorithm for detecting diffs - hashing each tile and comparing it to the known hash - would mean that every time the `kart diff` or even `kart status` command was run, you'd have to wait for all the tiles in your Kart repository to be hashed.
Git has this same problem for large repositories but it works around it by storing not just the content-hash but also the OS's last-modified-time in the index. If the file on disk's last-modified-time matches the last-modified-time in the index, we can infer that it's hash must also match whatever is in the index, so we don't need to re-hash the file.

Possibly, this is less important for maybe-smaller "attachment" files than for possibly-gigabytes-datasets-files. But, I still want to reuse the mechanisms we already have in workdir, for the following reasons:

- no need to check in a second mechanism for discovering which files are dirty when we already have one.
- I believe the existing mechanism already actually finds dirty attachment files - it searches for all dirty files and then sorts them into which dataset they belong with. The files that are left over after this sort are the dirty attachments - we may as well use this data
- if people do start checking in large attachments, the git-index based mechanism will run faster since it won't have to hash all the large attachments every time.

Look at how WorkdirDiffCache is passed around and used. This is created at most once when a diff or status command is run, and tile-based datasets use it to build up a list of tiles that have changed.

Possible pitfall - I think a Kart working copy is considered to be made out of two optional parts - the tabular part (often this is a GPKG) and the file part that we call the workdir. If you have only tabular data in your Kart repository, the workdir may not have been initialized. To add attachment support, we would have to make sure the workdir is always initialized and looking for file-based diffs.

olsen232 (changes requested):
Just a couple of high level comments so far that unfortunately could require quite a lot of code changes to address.

My humblest apologies for not being more proactive on the issue - had the issue been more specific about what I am envisioning, I could have saved you some work.

kart/diff_util.py
TeoGoddet: I will propose a new unified version

kart/checkout.py
TeoGoddet: I implemented your solution

TeoGoddet:
Hello @olsen232, I've pushed a version that is reusing the existing mechanism.

tests/test_attachment_edge_cases.py
```diff
@@ -0,0 +1,67 @@
...
+
+H = pytest.helpers.helpers()
+
+
+# ---------------------------------------------------------------------------
+# repo[oid] error semantics in RepoStructure.apply_files_diff
+# ---------------------------------------------------------------------------
+
+
+def test_repo_subscript_raises_keyerror_for_missing_oid(data_archive):
```
olsen232, line 22: This file feels like it was Claude's idea, I think most or even all of this file should be deleted:

- it seems to be a memorial to something that went wrong during development. Perhaps an earlier attempt at building this failed due to not understanding what repo[oid] or repo.get(oid) would do, and so someone, quite possibly Claude, has decided that as well as fixing the implementation, we should also make a record of what went wrong and how the repo behaves in this test. However it's not necessary to check in a record of any mistakes that were made during development of a feature, in fact it's best if we just check in a final clean working version.

- this doesn't actually target any Kart code or any particular possible Kart regression - tests for this behaviour probably already exist in pygit2.

olsen232 (changes requested):
Hi Teo,

I appreciate you are trying to add some work to Kart which has a large scope - large enough that none of the maintainers have found the time to embark on it.

It's difficult to review such a large change made externally, and I'm not sure I'm doing a great job. I still have issues with overall structure of this change.

I reread workdir.py and found there is a flag called X_KART_ATTACHMENTS (I haven't read the code in some time so apologies if I am only gradually telling you about the functionality you need to know about that Kart already has).

Setting this experimental flag causes Kart to do some - not all - of the things that your change also does. It doesn't cause Kart to look for file changes when you run `kart diff` or `kart status` but, when enabled, it does cause Kart to keep the attachment files on the repo in sync with the current commit - writing, updating, and deleting them where required as you check out different branches.
This means that Kart already contains code for doing that particular task, and also the related sub-tasks:
- keeping the workdir index in sync with the attachment files it is writing to the workdir
- updating the workdir state table with metadata about which commit we are currently synced to

This PR has taken some steps in the right direction by beginning to lean more on existing functionality that is within workdir.py, like dirty_paths.
But, it's still not using this other existing functionality that is inside workdir.py (see write_attached_files_to_workdir and surrounding code).
A better approach than the current one would be to use these existing methods wherever possible - enhancing them whereever it is needed.

Otherwise we'll end up with two conflicting implementations that handle or partially handle attachments from within Kart. As this PR stands right now, when you check out a branch, I think all of the following can happen:

1. the working copy is instructed to do a reset from commit A to commit B
  a) it delegates to the tabular working copy part, which does a reset from commit A to commit B
    i. it updates the data as needed in all of the tables
    ii. then records that the tabular working copy part is now synced with commit B
  b) then it delegates to the file-system working copy part, which does a reset from commit A to commit B
    i. it updates the tiles as needed in all of the tile-based datasets, and updates the workdir-index in sync with these
    ii. it skips updating any attachment files, since this is incomplete and X_KART_ATTACHMENTS flag is not set
    iii. then it records that the file-system working copy part is now synced with commit B
2. Finally, checkout.py does some of the same things as the file-system working copy:
  i. it updates any attachment files and updates the workdir-index in sync with these
  ii. then it records that the file-system working copy part is now synced with commit B


From a long-term maintenance standpoint, I have to balance my excitement that we're finally adding this feature to Kart with the knowledge that, at some point in the future we'll probably have to fix this by removing the redundant implementation and folding it all back into workdir.py
There's also the risk that checkout.py will conflict with workdir.py in some way - currently they both take responsibility for writing only a subset of the files that need to be changed, but both update workdir-state.db to record which commit is now checked out as if they had synced all the necessary files.
And, there's the risk that we fail to add a call to the new attachment-specific-reset code in checkout.py in every place where we make the call to reset the working copy.

olsen232:
Ok you might find this a bit frustrating but hopefully on balance you'll be happy to know that I got claude to do this - me telling you what I wanted you to tell Claude was maybe not fair on any of us.
It's here right now: https://github.com/koordinates/kart/tree/kart-attachments - 

You can if you want test the plugin changes against it, etc.
I haven't tested it enough myself yet to PR it yet.

TeoGoddet:
Hello, 
To be honest, my initial goal was to be able to support QGIS project in the QGIS project. 
It turned out to be working (with a lot of call to the git exe) but I was feeling that the code should not be in the plugin but rather in the main kart exe. 
I should have identified this was a too big for a first contribution on kart.

About using Claude, I agree the situation is a problem and I must apologize for it. 
I'm sorry that I could not identify the code duplication and I must admit I did not reviewed the test cases enough.

I will test the plugin on your branch

olsen232:
Sure - we're all learning about claude still. It's a strange new world.

My branch might not do everything yours does, I remember yours had extra commands that you added for the plugin to use, feel free to start from my branch and re-introduce the commands.

PR #1102, opened 2026-05-07
ESRI Rest Serivce Import support

hamishcampbell:
## Description

Esri Rest services were already importable with the (slightly secret) OGR:ESRIJSON [driver](https://gdal.org/en/stable/drivers/vector/esrijson.html) but it was painful to get the correct parameters. E.g. you needed to know that the `where`, `outFields` and `returnGeometry` fields should be set, and you couldn't import multiple layers from a single source easily.

This PR adds a dedicated `esri:` handler that accepts an Esri ArcGIS Map/Feature Server URL and simplifies the import process. For example:

```
> kart import esri:https://HOST/PATH/MapServer/ -a
Importing 19 features from https://HOST/PATH/MapServer/0:ESRIJSON) to First Layer/ ...
Added 19 Features to index in 0.0s
Overall rate: 2959 features/s)
Closed in 0s
Importing 3 features from PkGeneratingTableImportSource(https://HOST/PATH/MapServer/1:ESRIJSON) to Second Layer/ ...
Added 3 Features to index in 0.0s
Overall rate: 4675 features/s)
Closed in 0s
Importing 303 features from PkGeneratingTableImportSource(https://HOST/PATH/MapServer/2:ESRIJSON) to Third Layer/ ...
Added 303 Features to index in 0.0s
Overall rate: 10269 features/s)
Closed in 0s
```

Any supplied URL parameters are preserved. E.g. a narrower set of features & fields can be imported from a specific source:

```
# Only import the attributes `field1` and `field2` of features with `status=current`
kart import esri:https://HOST/PATH/MapServer/1?where=status%3Dcurrent&outFields=field1,field2
```

This PR includes other improvement to the importer framework to make it simple to add new handlers. See d823ec752cc07d04d02148eecfb86925687a1ba6 for more detail. The subsequent commit 17b7dc4f6ad414fcb8aef2f0d95576f7e68fa319 has the esri-specific changes.

## Future Improvements

This doesn't recurse into folders, so the multi-layer importer will only work from a single MapServer/FeatureService endpoint. It would be straight-forward to extent this to iterate into folders in future.

## Related links:

N/A

## Checklist:

### DIFF
--- a/kart/import_sources.py
+++ b/kart/import_sources.py
@@ -13,12 +13,13 @@ class ImportType(Enum):
 
     SQLALCHEMY_TABLE = auto()
     OGR_TABLE = auto()
+    ESRI_REST_TABLE = auto()
     POINT_CLOUD = auto()
     RASTER = auto()
 
     @property
     def import_cmd(self):
-        if self in (self.SQLALCHEMY_TABLE, self.OGR_TABLE):
+        if self in (self.SQLALCHEMY_TABLE, self.OGR_TABLE, self.ESRI_REST_TABLE):
             from kart.tabular.import_ import table_import
 
             return table_import
@@ -37,6 +38,11 @@ class ImportType(Enum):
             from kart.tabular import SqlAlchemyTableImportSource
 
             return SqlAlchemyTableImportSource
+
+        elif self is self.ESRI_REST_TABLE:
+            from kart.tabular import ESRIRestImportSource
+
+            return ESRIRestImportSource
         elif self is self.OGR_TABLE:
             from kart.tabular import OgrTableImportSource
 
@@ -113,6 +119,12 @@ ALL_IMPORT_SOURCE_TYPES = [
         "PATH.shp",
         ImportType.OGR_TABLE,
         file_ext=(".shp", ".shx", ".dbf"),
+    ),
+    ImportSourceType(
+        "ESRI Rest Service",
+        "esri:https://HOST/PATH/FeatureServer/[LAYER_ID]",
+        ImportType.ESRI_REST_TABLE,
+        uri_scheme="esri",
     ),
     ImportSourceType(
         "OGR", "OGR:...", ImportType.OGR_TABLE, uri_scheme="OGR", hidden=True
--- a/kart/import_sources.py
+++ b/kart/import_sources.py
@@ -1,9 +1,7 @@
 from enum import Enum, auto
 import os
 
-import click
-
-from kart.exceptions import NotYetImplemented
+from kart.exceptions import ImportSourceError, NotYetImplemented
 
 
 class ImportType(Enum):
@@ -59,7 +57,6 @@ class ImportSourceType:
         *,
         uri_scheme=None,
         file_ext=None,
-        optional_prefix=None,
         hidden=False,
     ):
         self.name = name
@@ -73,7 +70,6 @@ class ImportSourceType:
             self.file_ext = (file_ext,)
         elif isinstance(file_ext, tuple):
             self.file_ext = file_ext
-        self.optional_prefix = optional_prefix
         self.hidden = hidden
 
     @property
@@ -92,7 +88,6 @@ ALL_IMPORT_SOURCE_TYPES = [
         "PATH.gpkg",
         ImportType.SQLALCHEMY_TABLE,
         file_ext=".gpkg",
-        optional_prefix="GPKG:",
     ),
     ImportSourceType(
         "PostgreSQL",
@@ -200,7 +195,7 @@ def suggest_specs(suggestions=None, import_types=None, indent="    "):
 
 
 def bad_spec_error(spec, suggestions=None):
-    """If the user provided a bad format specification, returns a UsageError that tries to help them fix it."""
+    """If the user provided a bad format specification, returns a ImportSourceError that tries to help them fix it."""
     if suggestions is None:
         suggestion = from_spec(spec, allow_unrecognised=True)
         suggestions = (
@@ -210,9 +205,9 @@ def bad_spec_error(spec, suggestions=None):
     try_the_following = (
         "Try one of the following" if len(suggestions) >= 2 else "Try the following"
     )
-    return click.UsageError(
+    return ImportSourceError(
         f"Unrecognised import-source specification: {spec}\n"
-        f"{try_the_following}:\n{suggest_specs(suggestions)}"
+        f"{try_the_following}:\n{suggest_specs(suggestions)}",
     )
 
 
--- a/kart/tabular/esri_rest_import_source.py
+++ b/kart/tabular/esri_rest_import_source.py
@@ -3,7 +3,8 @@ Import source for ESRI Rest services (MapServer/FeatureServer).
 
 ESRI Rest services can be imported using the "ESRIJSON" OGR driver which does the
 heavy lifting of fetching data from the service and converting it to a format we can
-import. This module provides two classes:
+import. This module provides ESRIRestImportSource which routes to either of the following
+depending on supplied source:
 
 ESRIRestServerSource:
     Discovers and lists all available layers from a MapServer/FeatureServer endpoint.
@@ -16,7 +17,7 @@ ESRIRestServerSource:
         tables = source.get_tables()
 
         # Import all layers from a service
-        kart import "https://example.com/.../MapServer"
+        kart import "esri:https://example.com/.../MapServer"
 
 ESRIJSONImportSource:
     Imports a specific layer from a MapServer/FeatureServer by layer ID.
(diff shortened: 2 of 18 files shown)


kart/import_sources.py
```diff
@@ -39,6 +38,11 @@ def import_source_class(self):
             from kart.tabular import SqlAlchemyTableImportSource
 
             return SqlAlchemyTableImportSource
+
+        elif self is self.ESRI_REST_TABLE:
+            from kart.tabular import ESRIRestImportSource
+
+            return ESRIRestImportSource
```
hamishcampbell, line 45: This is awkward IMO, but necessary without further refactoring of `ImportType`. A future improvement here would be to simplify the types to `TABULAR, POINT_CLOUD, RASTER` and define the import source class in `ImportSourceType` rather than here. That would simplify some other pieces too - e.g. the warning if you try to `--link` a tabular source.

kart/tabular/ogr_import_source.py
```diff
@@ -83,17 +86,7 @@ def adapt_source_for_ogr(cls, source):
...
             if prefix in LOCAL_PATH_FORMATS:
                 if not os.path.exists(ogr_source):
-                    raise NotFound(
-                        f"Couldn't find {ogr_source!r}", exit_code=NO_IMPORT_SOURCE
-                    )
-        else:
-            # see if any subclasses have a handler for this.
-            for subclass in cls._all_subclasses():
-                if "handle_source_string" in subclass.__dict__:
-                    retval = subclass.handle_source_string(ogr_source)
```
hamishcampbell, line 93: I tried this approach first but ultimately added extra steps vs adding the custom `esri:` prefix. I'm not 100% sure what the intention with this code is, but it's not used by anything so I've removed it on the basis that import source selection via `from_spec` is more reliable/universal.

hamishcampbell:
Hm seems I might've broken the old `OGR:[DRIVER]:...` approach here, which was unintended.

tests/test_esri_rest_import_source.py
```diff
@@ -0,0 +1,446 @@
...
+        assert isinstance(source, ESRIRestServerSource)
+
+    def test_routes_to_layer_source(self):
+        """Test that layer URLs route to ESRIJSONImportSource."""
+        # Note: This will fail at OGR open time, but we can verify the routing
+        try:
+            source = ESRIRestImportSource.open(
+                "esri:https://example.com/arcgis/rest/services/MyService/MapServer/0"
+            )
+            assert isinstance(source, ESRIJSONImportSource)
```
olsen232: This test doesn't look quite right:

Firstly, `with pytest.raises` is preferable to `try ... except ... pass`

Having an assert within the block from whence we except an exception to be raised doesn't look right either - the assert could easily be skipped when we just straight to the except, or, the assert could raise an exception that we catch and ignore, in which case, why assert it.

kart/tabular/esri_rest_import_source.py
```diff
@@ -0,0 +1,413 @@
...
+        source,
+        ogr_source,
+        dest_path=None,
+        primary_key=None,
+        meta_overrides=None,
+    ):
+        # For ESRI JSON sources, the datasource typically has only one layer
+        # If no table is specified or the specified table doesn't exist,
+        # use the first (and usually only) layer in the datasource
+        if table is None or ogr_ds.GetLayerByName(table) is None:
```
olsen232, line 329: I feel like this could be rewritten for clarity. I think there's effectively 2 sets of 2 branches that needn't actually interfere with each other, which should make it easier to follow:

1a) if there is one layer in the source import that layer
1b) otherwise, import the specified one (argument "table")

2a) if argument table is set, use that as dest_path
2b) otherwise, use the name of the layer

olsen232: Actually I just did some testing and I don't think the case where the table doesn't match is reachable currently - at least when I tried it, this was prevented by a check elsewhere:

```
kart/tabular/import_.py(280)table_import()
    279             meta_overrides["metadata.xml"] = meta_overrides.pop("xmlMetadata")
--> 280         import_source = base_import_source.clone_for_table(
    281             table,

kart/tabular/ogr_import_source.py(211)clone_for_table()
    210         meta_overrides = {**self.meta_overrides, **meta_overrides}
--> 211         self.check_table(table)
    212
    
kart/tabular/ogr_import_source.py(262)check_table()
    261         if table_name not in self.get_tables():
--> 262             raise NotFound(
    263                 f"Table '{table_name}' not found",
```

What should happen here might depend on if or how you address my other comment about finding the names of these layers.

kart/import_sources.py
olsen232: Feel free to address if you're feeling keen 😃

kart/tabular/esri_rest_import_source.py
```diff
@@ -0,0 +1,413 @@
...
+import sys
+from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
+from urllib.request import urlopen
+
+from .import_source import TableImportSource
+from .ogr_import_source import OgrTableImportSource
+from kart.exceptions import ImportSourceError
+
+
+class ESRIRestImportSource(TableImportSource):
```
olsen232, line 44: OK this is great in that it supports everything - you can import all tables from FeatureServer, or just one table from FeatureServer by name, or you can specify FeatureServer/0 and it'll import specifically that. Users probably don't want to read the help files so it's good if they can throw a URL at it and it will do something sensible.

What I don't like so much is that if you import FeatureServer/0, then the name is lost and replaced with ESRIJSON. This is probably Jack Dangermond's fault - a better URL would maybe be FeatureServer/name_of_dataset. but we can't fix that here. However, I wonder can we do something a bit cleverer where when someone asks to import FeatureServer/0, we also hit FeatureServer to find the name of the layer, and use that name if no other name is specified by the user.

kart/tabular/esri_rest_import_source.py
```diff
@@ -0,0 +1,413 @@
...
+ESRIJSONImportSource:
+    Imports a specific layer from a MapServer/FeatureServer by layer ID.
+    Automatically constructs proper query URLs with necessary parameters.
+
+    Example:
+        # Import a specific layer
+        kart import "esri:https://example.com/.../MapServer/0" --dataset my_layer
+
+        # Or without the esri: prefix
+        kart import "https://example.com/.../MapServer/0"
```
olsen232: kart import "https://example.com/.../MapServer/0 doesn't seem to work.
We could make it work by checking for the infix `/FeatureServer/` etc but searching for infixes gets a bit weirder than checking for prefixes - things can have more than one infix.
Feel free to either make it work or edit this comment.

kart/tabular/esri_rest_import_source.py
```diff
@@ -0,0 +1,413 @@
...
+    Use this when you want to import multiple layers from a service or need to see
+    what layers are available.
+
+    Example:
+        # List all layers in a service
+        source = ESRIRestServerSource.open("https://example.com/.../MapServer")
+        tables = source.get_tables()
+
+        # Import all layers from a service
+        kart import "https://example.com/.../MapServer"
```
olsen232: This currently doesn't work without the esri: prefix

kart/tabular/esri_rest_import_source.py
```diff
@@ -0,0 +1,413 @@
...
+import. This module provides two classes:
+
+ESRIRestServerSource:
+    Discovers and lists all available layers from a MapServer/FeatureServer endpoint.
+    Use this when you want to import multiple layers from a service or need to see
+    what layers are available.
+
+    Example:
+        # List all layers in a service
+        source = ESRIRestServerSource.open("https://example.com/.../MapServer")
```
olsen232, line 16: Bit of a mix of docs here, some aimed at callers and some aimed at end-users.

For end-users who want to know all the layers, there's also:
`kart import "esri:https://example.com/.../MapServer" --list`

And for importing a particular layer:
`kart import "esri:https://example.com/.../MapServer" LAYER`

olsen232 (changes requested):

kart/tabular/esri_rest_import_source.py
hamishcampbell: Thanks, I'll update to fetch the metadata from the feature service URL (`.../0?f=json`)

kart/tabular/esri_rest_import_source.py
hamishcampbell: Comment updated

tests/test_esri_rest_import_source.py
hamishcampbell: Ah correct, I've fixed the test and made it clearer what it is trying to achieve

PR #1106, opened 2026-05-20
support git bundle-uri

blacha:
## Description

git bundles are a way of storing a full copy of a repository as a single file, they can be used to speed up initial clones of repositories, especially in networks where access to the source kart repository is slow but access to a internal bundle store is very fast.

```shell
$ kart clone --no-checkout kart@data.koordinates.com:linz/nz-building-outlines
```
approx 4-5 minutes

```shell
$ kart clone --no-checkout --bundle-uri=https://d1jzh93b1t1cv.cloudfront.net/source/nz-building-outlines.bundle kart@data.koordinates.com:linz/nz-building-outlines
```
< 1 minute (I saw as fast as ~30 seconds when using 10gbps connection to the store)

Bundles can be created with 

```shell
kart git bundle create ./nz-building-outlines.bundle --all
```

## Checklist:


hamishcampbell:
BTW you might be interested to know that Kart works quite well with git-remote-s3 (https://github.com/awslabs/git-remote-s3) which essentially supports git bundles in S3 buckets.

PR #1124, opened 2026-08-01
Speed up CI test runs: disable coverage, exclude build dirs from Defender

craigds:
## Description

Two zero-cost CI speedups, aimed mostly at the ~60-minute Windows unit-test step:

- Pass `--no-cov` to pytest in CI (via the `ci-base` preset): every CI test run pays coverage instrumentation overhead, but no workflow uploads or reads the coverage data. Local runs are unchanged.
- Exclude the workspace and temp dirs from Windows Defender real-time scanning: the runner is ephemeral, and scanning taxes every one of the thousands of compiler/git/kart subprocess spawns and temp files.

## Checklist:

- [ ] Have you included test(s)? (n/a - CI config)
- [ ] Have you updated the [changelog](https://github.com/koordinates/kart/blob/master/CHANGELOG.md)? (n/a - no user-facing change)

### Risks involved

- [x] No notable risks

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -85,6 +85,12 @@ jobs:
 
       # Setup
 
+      # Defender real-time scanning taxes every compiler/test subprocess spawn
+      # and temp file; the runner is ephemeral so scanning the build is pointless.
+      - name: "setup: defender exclusions"
+        shell: pwsh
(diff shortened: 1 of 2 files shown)


craigds:
actually it's not clear that this is a noticeable speedup after all 🤷 

but it's probably fine anyway, it can't _hurt_. and no one's been looking at that cov report so really we don't need it

craigds:
but it's also pretty low value, never mind

PR #1125, opened 2026-08-03
Helper mode: give subprocesses the caller's stdin/stdout/stderr

craigds:
## Description

In helper mode the pager produced no output, or `cat: standard output: Bad file descriptor` with `PAGER=cat` - and it alternated, working every second invocation.

The helper child only rebound `sys.stdin`/`stdout`/`stderr` to the fds passed over the socket. The process's real fds 0/1/2 still pointed at whatever the helper daemon had there: `cli_helper/kart.c` marks 0/1/2 `FD_CLOEXEC` before exec'ing the helper, so they get reused for its own sockets. Subprocesses inherit those, so the pager wrote into a socket or a closed descriptor. The alternation came from the accept loop holding each connection open until the next `accept()`, which shifted the received fd numbers by one each time - on every second run the caller's stdin happened to land on fd 1 and the pager wrote to the terminal by accident.

Fixed by `dup2()`ing the caller's fds onto 0/1/2 in the helper child. Second commit closes the sockets each process no longer needs after forking.

Also fixes the same problem for subprocess stderr, which was going to the caller's *stdin* fd.

## Related links:

Introduced as a user-visible bug by #1080 (pager support), though the underlying fd handling has always been wrong.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 - Piped (non-tty) output is now always UTF-8 on all platforms. Previously text output used the locale codepage on Windows (e.g. cp1252), inconsistent with JSON output (which is always UTF-8). [#1115](https://github.com/koordinates/kart/issues/1115)
 - `export`: Fixed a crash (`KeyError: 'GEOMETRY Z'`) when exporting datasets whose geometry type is the generic `GEOMETRY` with a `Z`, `M`, or `ZM` suffix. [#1063](https://github.com/koordinates/kart/issues/1063)
 - `log`: Fixed `<dataset>:tile` and raw-path filters silently matching nothing for point-cloud and raster datasets. [#1119](https://github.com/koordinates/kart/pull/1119)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,8 +11,8 @@ _When adding new entries to the changelog, please include issue/PR numbers where
 - Piped (non-tty) output is now always UTF-8 on all platforms. Previously text output used the locale codepage on Windows (e.g. cp1252), inconsistent with JSON output (which is always UTF-8). [#1115](https://github.com/koordinates/kart/issues/1115)
 - `export`: Fixed a crash (`KeyError: 'GEOMETRY Z'`) when exporting datasets whose geometry type is the generic `GEOMETRY` with a `Z`, `M`, or `ZM` suffix. [#1063](https://github.com/koordinates/kart/issues/1063)
 - `log`: Fixed `<dataset>:tile` and raw-path filters silently matching nothing for point-cloud and raster datasets. [#1119](https://github.com/koordinates/kart/pull/1119)
-- Fixed commands that write text output (eg `log`) hanging forever when stdin is a pipe that is never written to or closed, as happens under some CI runners. [#1121](https://github.com/koordinates/kart/pull/1121)
```
craigds, line 14: this didn't really seem worth mentioning, also it was probably introduced by the recent vcpkg update anyway so not really relevant release-package users

olsen232 (approved):

